PR #45, opened 2017-02-01
Adds Finatra for skeleton api

kenoir:
Addresses https://github.com/wellcometrust/platform-api/issues/16


app/src/main/scala/uk/ac/wellcome/platformapi/modules/ElasticClientModule.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  @Singleton
+  @Provides
+  def providesXPackConfig(): Option[XPackConfig] =
+    Option(xpackEnabled()).collect { case true =>
+      XPackConfig(user(), sslEnabled())
+    }
+}
+
+object ElasticClientModule extends TwitterModule {
```
kenoir: @jamesgorrie take a look at how Finatra does dependency injection ... it's beautiful.

app/build.sbt
```diff
@@ -0,0 +1,81 @@
...
+
+libraryDependencies ++= Seq(
+  "com.twitter" %% "finatra-http" % versions.finatra,
+  "com.twitter" %% "finatra-httpclient" % versions.finatra,
+  "ch.qos.logback" % "logback-classic" % versions.logback,
+
+  "com.typesafe" % "config" % "1.3.1",
+  "org.elasticsearch" % "elasticsearch" % "5.1.2",
+  "com.sksamuel.elastic4s" %% "elastic4s-core" % "5.1.5",
+  "com.sksamuel.elastic4s" %% "elastic4s-xpack-security" % "5.1.5",
```
alexwlchan: Any reason we don’t incorporate these versions into the `versions` field above?

app/src/main/scala/uk/ac/wellcome/platformapi/controllers/MainController.scala
```diff
@@ -0,0 +1,63 @@
...
+  entries: Record,
+  parent: Option[Collection]
+)
+
+@Singleton
+class MainController @Inject()(
+  calmService: CalmService
+) extends Controller {
+
+  val apiBaseUrl = "/api/v0"
```
alexwlchan: Not that it’s hugely important right now, but I think the consensus was we drop the `/api` from URLs, as it’s all being served from `api.wc.org`.

app/src/main/scala/uk/ac/wellcome/platformapi/controllers/MainController.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  get(s"${apiBaseUrl}/record") { request: CalmRequest =>
+    val recordCollectionPair = for {
+      recordOption <- calmService.findRecordByAltRefNo(request.altRefNo)
+      collectionOption <- calmService.findParentCollectionByAltRefNo(request.altRefNo)
+    } yield RecordCollectionPair(recordOption, collectionOption)
```
alexwlchan: Am I right in thinking that `recordOption` and `collectionOption` are futures here, so we don’t hit the `yield` until both have resolved? Or does `RecordCollectionPair` become its own future?

app/src/main/scala/uk/ac/wellcome/platformapi/models/Record.scala
```diff
@@ -0,0 +1,26 @@
...
+  def apply(hit: RichSearchHit): Record = {
+    val data: Map[String, AnyRef] =
+      hit.sourceAsMap.filter(o => o._2 != null)
+
+    Record(
+      altRefNo=data("AltRefNo").toString,
+      title=data("Title").toString,
+      materialType=data("Material").toString,
+      date=data.get("Date").map(_.toString),
+      acquisition=data.get("Acquisition").map(_.toString)
```
alexwlchan: What’s the difference between `foo.toString` and `foo.map(_.toString)`? Is it something to do with futures (and if so, why are only some of the fields futures?). Or because some of the fields are absent/null?

app/src/test/scala/uk/ac/wellcome/platformapi/controllers/MainControllerFeatureTest.scala
```diff
@@ -0,0 +1,20 @@
...
+class MainControllerFeatureTest extends FeatureTest {
+
+  override val server = new EmbeddedHttpServer(new Server)
+
+  "Server" should {
+    "respond" in {
+      server.httpGet(
+        path = "/",
+        andExpect = Ok,
+        withBody = "{\"message\":\"success\"}")
```
alexwlchan: tests!

:D

app/src/main/scala/uk/ac/wellcome/platformapi/exceptions/ElasticsearchExceptionMapper.scala
```diff
@@ -0,0 +1,22 @@
...
+)
+
+@Singleton
+class ElasticsearchExceptionMapper @Inject()(response: ResponseBuilder)
+  extends ExceptionMapper[ElasticsearchException] {
+
+  override def toResponse(request: Request, exception: ElasticsearchException): Response = {
+    val errorResponse = ErrorResponse(exception.getMessage)
+
+    response.internalServerError.json(errorResponse)
```
alexwlchan: This exposes error messages from the Elasticsearch library to end-users. It would be better if this logged the error message, and then we just returned a generic 500 error.

app/src/main/scala/uk/ac/wellcome/platformapi/controllers/ManagementController.scala
```diff
@@ -0,0 +1,27 @@
...
+
+  get("/management/healthcheck") { request: Request =>
+    response.ok.json(Map("message" -> "ok"))
+  }
+
+  get("/management/clusterhealth") { request: Request =>
+    elasticsearchService.client
+      .execute { clusterHealth() }
+      .map { response.ok.json }
+  }
```
alexwlchan: Open Q: do we want to expose these management endpoints? To end-users: definitely not. To internal staff: we should talk to James et al. about what they’re doing with healthcheck/monitoring and so on.

app/src/main/scala/uk/ac/wellcome/platformapi/controllers/MainController.scala
```diff
@@ -0,0 +1,63 @@
...
+case class CalmRequest(
+  @NotEmpty @QueryParam altRefNo: String
+)
+
+case class RecordCollectionPair(
+  record: Option[Record],
+  collection: Option[Collection]
+)
+
+case class RecordResponse(
```
alexwlchan: Consider defining these elsewhere?

app/src/main/scala/uk/ac/wellcome/platformapi/controllers/MainController.scala
```diff
@@ -0,0 +1,63 @@
...
+  altRefNo: String,
+  entries: Record,
+  parent: Option[Collection]
+)
+
+@Singleton
+class MainController @Inject()(
+  calmService: CalmService
```
alexwlchan: Open Q: how does an instance of `CalmService` actually appear here?

Issue #21, opened 2017-01-27
Create CALM data ingestor


jtweed:

alexwlchan:
Superseded by #52, #53, #54 and #55.

PR #46, opened 2017-02-07
Create Calm data ingestor, part 1

alexwlchan:
Related to #21.

This isn’t done, but it exposes the inconsistencies and muckiness of the Calm data.


alexwlchan:
Also related: https://github.com/wellcometrust/platform-infra/pull/1

This currently fails because we have at least one record with an empty `RefNo` field.

calm_to_dynamodb.py
```diff
@@ -0,0 +1,199 @@
...
+    import boto3
+
+    dynamodb = boto3.resource('dynamodb')
+    table = dynamodb.Table('CalmData')
+
+    # We do some bookkeeping to avoid sending the same record to DynamoDB
+    # multiple times, which is both slow and unnecessary cost.  The full
+    # Calm export has ~331k records, so the more we can skip, the better.
+    try:
+        existing = json.load(open('existing.json', 'r'))
```
alexwlchan: This should probably be kept in S3, perhaps along with a record of when the record was last pushed into DynamoDB (for comparison with the `Modified` key in Calm).

alfraser86:
@alexwlchan love the architecture image.

calm-ingest/prototype.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python
```
alexwlchan: This script is a bit rough around the edges (and in particular, is pulling from local cache rather than DynamoDB, but proves the idea).

alexwlchan:
@kenoir Could you review this in its current form please? I’m aware this is starting to get quite big, and I’d rather merge it now, then refine it, than continue growing this patch.

calm-ingest/calm_to_dynamodb.py
```diff
@@ -0,0 +1,212 @@
...
+    args = parse_args()
+
+    dynamodb = boto3.resource('dynamodb')
+    table = dynamodb.Table('CalmData')
+
+    # We do some bookkeeping to avoid sending the same record to DynamoDB
+    # multiple times, which is both slow and unnecessary cost.  The full
+    # Calm export has ~331k records, so the more we can skip, the better.
+    try:
+        existing = json.load(open('existing.json', 'r'))
```
kenoir: I guess this is ok for now, but loading from a file like this makes parallelisation impossible.

calm-ingest/mapping_collection.py
```diff
@@ -0,0 +1,60 @@
...
+"""
+Elasticsearch mapping for Collection records in Calm.
+"""
+
+import json
+
+import requests
+
+
+MAPPING = {
```
kenoir: Is there any mileage in just making these `.json` files and loading them? Might allow abstraction of the `push_mapping` below.

calm-ingest/prototype.py
kenoir: Do we need to keep track of where we are in the ingest from Dynamo to ES? This is ok to merge for now, but can we chat about how this step is going to work when we start getting updates rather than batch processing old stuff.

kenoir:
👍 Nice. Merge away, we can talk about my comments when you're in.

PR #61, opened 2017-02-27
Move to multi-project layout

kenoir:
In order to accomodate microservice approach with common-lib


project/Common.scala
```diff
@@ -0,0 +1,32 @@
...
+object Common {
+  val settings: Seq[Def.Setting[_]] = Seq(
+
+    scalaVersion := "2.11.8",
+    organization := "uk.ac.wellcome",
+    git.baseVersion := "0.0.1",
+
+    resolvers += Resolver.sonatypeRepo("releases"),
+    resolvers += "maven.twttr.com" at "https://maven.twttr.com",
+    resolvers += "elasticsearch-releases" at "https://artifacts.elastic.co/maven",
```
alexwlchan: Should this dependency be moved down?

build.sbt
```diff
@@ -0,0 +1,25 @@
+import sbt.Keys._
+import com.amazonaws.regions.{Region, Regions}
+
+scalaVersion := "2.11.8"
+
+fork in run := true
+
+enablePlugins(JavaAppPackaging)
+enablePlugins(GitVersioning)
```
alexwlchan: Can/should these lines be moved down?

PR #64, opened 2017-02-28
Tidy up the Circle build/test scripts

### LINKED ISSUE #63: Tidy up the scripts for multi-project builds in Circle
We merged some fairly simple scripts as part of #62. They work, but they need tidying up:

* [x] Make it easier to add new projects
* [x] Comments
* [x] Only build parts of the project that have changed

alexwlchan:
Changes:

* Consolidate the Circle scripts into a single file that takes an argument “compile”, “test” or “deploy”.
* Only build projects based on whether their files have changed in the current branch.

I’ll want to squash the Git bits before we merge, because I think there’s a non-trivial chance we’ll want to revert to the “just build everything” behaviour later.

Resolves #63.

### DIFF
--- a/run_circle.sh
+++ b/run_circle.sh
@@ -17,9 +17,9 @@ set -o errexit
 set -o nounset
 
 
-
 # Read command-line arguments
-if (( $# == 1 )); then
+if (( $# == 1 ))
+then
     TASK="$1"
 else
     echo "Usage: run_circle.sh <TASK>"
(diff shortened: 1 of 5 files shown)


run_circle.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
```
kenoir: Maybe `#!/usr/bin/env bash`

alexwlchan: Sounds sensible.

PR #71, opened 2017-03-02
[WIP] Add transformer for Dynamo -> Transform -> SQS

alexwlchan:
Transforms "raw" records from dirty source systems (e.g. CALM) and transforms records to our "unified" format.

This app does not write to ES but publishes updates to SNS from where they are picked up by a downstream ingestor (see https://github.com/wellcometrust/platform-api/pull/75).

```
DynamoDB Stream -> **Transformer** -> SNS
```

Supercedes #67: rebased against current master and with unused imports removed.


kenoir:
Resolves: https://github.com/wellcometrust/platform-api/issues/54

PR #73, opened 2017-03-06
Add Swagger Support

kenoir:
Adds `GET: /api-docs/ui`

Branched from: https://github.com/wellcometrust/platform-api/pull/71

untitled

This needs fleshing out with some real examples.


kenoir:
Resolves: https://github.com/wellcometrust/platform-api/issues/44

jtweed:
I'm going to add our own custom SwaggerController so that we're not loading the webjars stuff and the path for the swagger.json matches the convention in the spec.

jtweed:
Ok, I'm done. @kenoir if you can review and merge this, I will then use the same branch to start adding some API structure and early docs.

jtweed:
Have done that, to a very basic level. This is ready to merge once it's been reviewed.

kenoir:
@jtweed this looks great 👍 

It needs the correct CORS headers set as currently without the UI a Swagger (browser) client cannot consume the JSON. That's a systemic problem with our setup rather than an issue with this changeset so please merge.

PR #76, opened 2017-03-08
Add ingestor for SQS -> ES

kenoir:
Adds a worker app that sits between SQS and Elasticsearch reading from SQS and parsing JSON to our "unified" ES model, records are then POSTed to Elasticsearch.

```
Transformer -> SNS -> SQS -> **Ingestor** -> Elasticsearch
```

Supercedes: https://github.com/wellcometrust/platform-api/pull/75


kenoir:
Resolves https://github.com/wellcometrust/platform-api/issues/68

build.sbt
```diff
@@ -22,6 +22,17 @@ lazy val api = project.
   enablePlugins(GitVersioning).
   settings(libraryDependencies ++= Dependencies.apiDependencies)
 
+lazy val ingestor = project.
```
kenoir: This is rubbish!

kenoir:
This branch fixes https://github.com/wellcometrust/platform-api/issues/70

build.sbt
```diff
@@ -22,6 +22,17 @@ lazy val api = project.
...
+  dependsOn(common).
+  settings(Common.settings: _*).
+  settings(Finatra.settings: _*).
+  settings(Revolver.settings: _*).
+  settings(EcrPlugin.projectSettings: _*).
+  settings(Packager.settings: _*).
```
alexwlchan: It would be nice if we could reduce the repetition here.

common/src/main/scala/uk/ac/wellcome/models/UnifiedItem.scala
```diff
@@ -0,0 +1,15 @@
+package uk.ac.wellcome.models
+
+import com.sksamuel.elastic4s._
+import uk.ac.wellcome.utils.JsonUtil
+
+
+case class UnifiedItem(
```
alexwlchan: Is this the "CleanedRecord" class we were working with last week?

What’s the convention for documenting/commenting classes in Scala?

ingestor/src/main/resources/logback.xml
```diff
@@ -0,0 +1,14 @@
...
+         ch.qos.logback.classic.encoder.PatternLayoutEncoder by default -->
+    <encoder>
+      <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+    </encoder>
+  </appender>
+
+  <root level="debug">
+    <appender-ref ref="STDOUT" />
+  </root>
+</configuration>
```
alexwlchan: Ditto comment above about repetition – isn’t this the same as we use in several other projects?

ingestor/src/main/scala/uk/ac/wellcome/ingestor/controllers/ManagementController.scala
```diff
@@ -0,0 +1,16 @@
...
+import scala.concurrent.ExecutionContext.Implicits.global
+
+import com.twitter.finagle.http.Request
+import com.twitter.finatra.http.Controller
+
+
+@Singleton
+class ManagementController @Inject()() extends Controller {
+  get("/management/healthcheck") { request: Request =>
+    response.ok.json(Map("message" -> "ok"))
```
alexwlchan: Maybe we should include the app name here? We now have several healthcheck endpoints that just return

```json
{"message": "ok"}
```

It might be nice to be able to distinguish by output alone.

ingestor/src/main/scala/uk/ac/wellcome/ingestor/modules/SQSWorker.scala
```diff
@@ -0,0 +1,147 @@
...
+  //TODO: Extract processors out, everything else into a base trait
+  def indexDocument(
+    document: String,
+    elasticsearchService: ElasticsearchService
+  ): Future[Unit] = Future {
+    implicit val jsonMapper = UnifiedItem
+
+    JsonUtil.fromJson[UnifiedItem](document).map(item => {
+        elasticsearchService.client.execute {
+	  //TODO: Push index and type to config?
```
alexwlchan: yes, let’s do that

ingestor/src/main/scala/uk/ac/wellcome/ingestor/modules/SQSWorker.scala
```diff
@@ -0,0 +1,147 @@
...
+        message <- Future(extractSNSMessage(msg).get)
+
+        processorOption = message.getSubject().flatMap(chooseProcessor)
+
+        processor <- Future(processorOption.getOrElse({
+	  error(s"Unrecognised message subject ${message.getSubject()}")
+          throw new RuntimeException("Failed to process message")
+        }))
+
+        _ = processor.apply(message.getBody(), elasticsearchService)
```
alexwlchan: Why the assignment here?

ingestor/src/main/resources/logback.xml
kenoir: It is possible to move the log config into code and then share that between projects as here: 

https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/LogConfig.scala

Then include:

https://github.com/guardian/grid/blob/master/thrall/app/Global.scala#L9

^ That example will need to be reworked for Finatra but the idea is sound I think.

ingestor/src/main/scala/uk/ac/wellcome/ingestor/controllers/ManagementController.scala
kenoir: I think there should be a `/management/manifest` endpoint for each application that carries:

- Name
- Version

Like this:

https://github.com/guardian/grid/blob/master/common-lib/src/main/scala/com/gu/mediaservice/lib/management/Management.scala#L18

Also extractable to a common library.

ingestor/src/main/scala/uk/ac/wellcome/ingestor/modules/SQSWorker.scala
kenoir: As this is a `for` comprehension there is limited syntax to support running "side effecting" functions as here. Essentially the `_` here is saying "throw away the value because we don't care".

common/src/main/scala/uk/ac/wellcome/models/UnifiedItem.scala
kenoir: http://docs.scala-lang.org/style/scaladoc.html

common/src/main/scala/uk/ac/wellcome/models/DynamoConfig.scala
```diff
@@ -0,0 +1,6 @@
+package uk.ac.wellcome.models
```
kenoir: Move these under AWS namespace.

ingestor/src/main/scala/uk/ac/wellcome/ingestor/modules/SQSWorker.scala
```diff
@@ -0,0 +1,95 @@
+package uk.ac.wellcome.platform.ingestor.modules
+
+import java.util.concurrent.TimeUnit
```
kenoir: Check if we can remove some of these imports.

ingestor/src/main/scala/uk/ac/wellcome/ingestor/modules/SQSWorker.scala
```diff
@@ -0,0 +1,95 @@
...
+import com.sksamuel.elastic4s.ElasticClient
+import com.sksamuel.elastic4s.ElasticDsl._
+import uk.ac.wellcome.models.SQSMessage
+
+import uk.ac.wellcome.utils.JsonUtil
+
+object SQSWorker extends TwitterModule {
+  override val modules = Seq(SQSConfigModule, SQSClientModule, AkkaModule)
+
+  val waitTime = flag("sqs.waitTime", 20, "SQS wait time")
```
kenoir: Do we need these flags in both places?

kenoir:
@alexwlchan I've addressed the review comments.

common/src/main/scala/uk/ac/wellcome/utils/TryBackoff.scala
```diff
@@ -0,0 +1,50 @@
...
+  val maxAttempts  = 75
+
+  def run(f: (() => Unit), system: ActorSystem, attempt: Int = 0): Unit = {
+
+    val nextAttemptTry: Try[Int] = Try {
+      f(); 0
+    } recoverWith {
+      case e: Throwable => {
+        error(s"Failed to get new messages (attempt: ${attempt})")
+	Try(attempt + 1)
```
alexwlchan: nit: indentation

Issue #77, opened 2017-03-08
Create basic works ontology


ChristopherHilton:
Determine basic properties that apply to works (considered broadly) and items, in order to inform API design.

Issue #31, opened 2017-01-27
Agree approach to identifiers


jtweed:

alexwlchan:
Copying some notes from a conversation with @kenoir:

Our identifier service needs to solve a couple of problems:

* **Uniqueness.** Because we have multiple data sources allocating their own identifiers, there might be records from two sources which have the same internal ID, but which refer to different items.
* **Merging.** Records in different data sources with different IDs might refer to the same item – we need to be able to tie these together under the same ID in the new system.
* **Deterministic.** Given a record, we need to be able to check if there’s already an allocated ID (so we can update it rather than create a new record).

Our current idea looks something like this:

<img width="599" alt="identifiers" src="https://cloud.githubusercontent.com/assets/301220/23792614/9862a95c-057f-11e7-92c2-88951168f7d7.png">

Records within each data source are as follows:

```
record:
    - primary_id
    - identifiers
        - alt_id_1
        - alt_id_2
        - ...
```

where we might want to correlate `identifiers` across multiple data sources.

So when the transformer processes a record, it asks an ID service _“I have this primary ID X and identifiers Y from data source Z, do you have a canonical ID?”_ The ID service checks our Elasticsearch index for matching records:

* If one exists, you return it
* If it doesn’t exist, you create one from scratch _and create a stub record for it in Elasticsearch_. This is to prevent race conditions – e.g. if a record for an item is created in Sierra and Calm on the same day, we don’t want the ingestors to create two distinct IDs. This stub record should be unsearchable through the public API.

How those identifiers are allocated remains an open problem.

jtweed:
Are we sure it's a good idea to rely on Elasticsearch here? Is it not safer to use DynamoDB for the ID service?

We may have multiple ES indexes, they may be out of date, it's more likely to need reverted to a snapshot etc...

jtweed:
I think we may also need to create an identifier for every entity we ingest, so that we can consistently refer to them internally. We can't create valid JSON[-LD] without consistent identifiers. We'd then need the ID service to nominate one as canonical for de-duped entities.

alexwlchan:
> I think we may also need to create an identifier for every entity we ingest, so that we can consistently refer to them internally

Each data source comes with its own identifiers (and some refer to each other) – my idea was that the alt identifiers would be lifted directly from the individual data sources.

The internally consistent identifier is then the one we mint ourselves, but don’t expose publicly.

alexwlchan:
[...]

PR #78, opened 2017-03-08
Adds works ontology

ChristopherHilton:
This is the works ontology, as far as it's got

Relates to #77


kenoir:
Oh god, why is it all on one line?

jtweed:
Windows line endings?

@ChristopherHilton Can we set your editor up to use Unix line endings?

alexwlchan:
> why is it all on one line?

Actually, I suspect encodings are at fault – inspecting the file, I see a bunch of `\xa0` bytes. That’s a non-breaking space in Latin 1, not sure what in UTF-8. I bet the encoding doesn’t match what GitHub expects.

alexwlchan:
Put another way, running `dos2unix works.ttl` has zero effect on legibility.

kenoir:
@ChristopherHilton I think the long term solution to this is to either a) Get a mac with admin privileges so you can install what you like, b) get IT to install a better editor. @jtweed what do we need to do to get Chris a mac?

ChristopherHilton:
Agreed, I think simplest is likely to be getting a Mac so there’s consistency in kit across the whole team.  I’d still need access to CALM and Sierra via a Windows machine but that’s not insuperable obviously – either we do it via a VM or I just keep the old laptop also.  A second docking station and a toggle for the screen would presumably be fairly simple for IT to set up?

ChristopherHilton:
Alex has cracked it I think - it's the browser.  Using Chrome we actually have line breaks!  Progress...

ontologies/works.ttl
```diff
@@ -0,0 +1,344 @@
...
+    dcterms:title "Works"@en ;
+    dcterms:created "13/02/2017"^^xsd:date ;
+    dcterms:description "Ontology describing the properties of works"@en .
+	
+
+#######  Classes ##### 	
+
+ww:Work rdf:type owl:Class ;
+	rdfs:label "Work"@en ;
+	rdfs:comment "An individual work such as a text, archive item or picture; or a grouping of individual works (so, for instance, an archive collection counts as a work, as do all the series and individual files within it).  Each work may exist in multiple instances (e.g. copies of the same book).  N.B. this is not synonymous with "work" as that is understood in FRBR but represents something lower down the FRBR hierarchy, namely manifestation. Groups of related items are also included as works because they have similar properties to the individual ones."@en ;
```
alexwlchan: I’m assuming “FRBR” is https://en.wikipedia.org/wiki/Functional_Requirements_for_Bibliographic_Records?

Might be worth including the full version the first time, for readers who aren’t familiar.

ontologies/works.ttl
```diff
@@ -0,0 +1,344 @@
...
+	
+ww:Collection rdf:type owl:Class ;
+	rdfs:label "Collection"@en ;
+	rdfs:comment "A grouping of material comprising more than one work, the grouping being either organic (i.e., a Fonds) or artificial."@en ;
+	rdfs:subClassOf ww:Work
+	rdf:isDefinedBy <TBC/works> .
+	
+ww:Item rdf:type owl:Class ;
+	rdfs:label "Item"@en ;
+	rdfs:comment "A specific instance of a work: for instance, an individual physical copy with its own location.  This corresponds to the bottom, Instance, layer of FRBR.  Works that are unique - that is, have only one copy - are described in terms of a Work for their intellectual content, linked to an Item describing their phyiscal characteristics."@en ;
```
alexwlchan: /s/phyiscal/physical

ontologies/works.ttl
```diff
@@ -0,0 +1,344 @@
...
+	rdf:isDefinedBy <TBC/works> .
+	
+ww:Location rdf:type owl:Class ;
+	rdfs:label "Location"@en ;
+	rdfs:comment "The place at which a work was published or otherwise created."@en ;
+	rdf:isDefinedBy <TBC/works> .
+	
+ww:Genre rdf:type owl:Class ;
+	rdfs:label "Genre"@en ;
+	rdfs:comment "The genre to which a work belongs: Genre being taken to mean a particular form carrying certain assumptions about content, authorial stance, etc.  Examples might include biography, article, obituary, satire, etc."@en ;
```
alexwlchan: Nit: either say “Examples might include” or “etc”, but I don’t think you need both.

ontologies/works.ttl
```diff
@@ -0,0 +1,344 @@
...
+ww:identifier rdf:type owl:DatatypeProperty ;
+	rdfs:label "identifier"@en ;
+	rdfs:comment "Relates a work to an identifier such as a manuscript number"@en ;
+	rdfs:domain ww:Work ;
+	rdfs:range rdf:langString ; 
+	rdf:isDefinedBy <TBC/works> .
+	
+ww:level rdf:type owl:DatatypeProperty ;
+	rdfs:label "level"@en ;
+	rdfs:comment "Indicates the position of a work within a hierarchy of works: for example, Collection, Series, Item etc. Item in CALM descriptions is taken to mean an individual work, which is then manifested in a physical or digital item."@en ;
```
alexwlchan: My (admittedly limited) understanding is that our ontologies are data source agnostic – should we be mentioning CALM here?

ontologies/works.ttl
```diff
@@ -0,0 +1,344 @@
...
+ww:systemIdentifier rdf:type owl:DatatypeProperty ;
+	rdfs:label "systemIdentifier"@en ;
+	rdfs:comment "Relates the work to an identifier that is not made visible to the public but used to generate hierarchy."@en ;
+	rdfs:domain ww:Work ;
+	rdfs:range rdf:langString ; 
+	rdf:isDefinedBy <TBC/works> .
+	
+ww:systemNumber rdf:type owl:DatatypeProperty ;
+	rdfs:label "systemNumber"@en ;
+	rdfs:comment "Relates the work's metadata to an system-generated GUID."@en ;
```
alexwlchan: Is this the canonical identifier discussed in #31?

ontologies/works.ttl
kenoir: 👍

ontologies/works.ttl
```diff
@@ -0,0 +1,344 @@
+
+@prefix ww: <TBC/works> . 
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> . 
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix dcterms: <http://purl.org/dc/terms/> .
+@prefix owl:     <http://www.w3.org/2002/07/owl#> .
+@prefix rdfs:    <http://www.w3.org/2000/01/rdf-schema#> .
+
+TBC
```
kenoir: This file currently is not currently valid (i.e. it's not machine readable). I think you can safely remove the TBC here.

kenoir:
It'd be nice to have some docs and tooling in the repo around viewing/editing/understanding the ontology files. At the very least a `README.md` would be good here too.

ontologies/works.ttl
jtweed: Agree with @alexwlchan - I don't think we should mention specific systems in the ontologies. We should describe the semantics, not the source.

Also, I think there's an argument not to add this predicate just yet. I think we will need it, but we should consider it as part of requirements for tree shortcuts when those arise, e.g. root, leaf, depth etc.

ontologies/works.ttl
jtweed: I'm not sure the intent here, but just the naming makes me uncomfortable. The identifier discussed in #31 will be in the URI. We may also have a separate property for us to use internally for ease of access, but the reality is that we will probably use the aliased "id" property for that in the JSON.

In other words, internally we'll be treating the JSON the same as other non-RDF clients, blissfully unaware that if we parse the @context and add the @base, that id is actually @id and is a URI. If that makes any sense!

I'd drop this predicate for now, or rename if @ChristopherHilton can tell us what it is...

ChristopherHilton: This is the systemNumber?  Basically just there to describe the RecordID field in CALM, and if necessary the similar field in Preservica, which gives a unique system reference to each record (illustration from CALM attached).  It's not something we've ever surfaced deliberately in the front end and have wanted to suppress when it does appear, but for completeness's sake I included it.... however happy to drop it pro tem if it creates more bother than it's worth.
calm_recordid

ChristopherHilton: This is the systemNumber?  Basically just there to describe the RecordID field in CALM, and if necessary the similar field in Preservica, which gives a unique system reference to each record (illustration from CALM attached).  It's not something we've ever surfaced deliberately in the front end and have wanted to suppress when it does appear, but for completeness's sake I included it.... however happy to drop it pro tem if it creates more bother than it's worth.
calm_recordid

jtweed: I'd drop for now until we have a use case.

ontologies/works.ttl
ChristopherHilton: Fair point, I'll remove the reference to the System That Shall Not Be Named...

ontologies/works.ttl
ChristopherHilton: OK

ontologies/works.ttl
ChristopherHilton: Will do!

ChristopherHilton:
Just added in the stuff raised in comments on the previous version - removed mention of CALM, expanded on FRBR, etc.

ontologies/works.ttl
```diff
@@ -0,0 +1,344 @@
+
+@prefix ww: <TBC/works> .
```
kenoir: http://wellcomecollection.org/ontologies/workss

ontologies/works.ttl
```diff
@@ -0,0 +1,344 @@
...
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> . 
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix dcterms: <http://purl.org/dc/terms/> .
+@prefix owl:     <http://www.w3.org/2002/07/owl#> .
+@prefix rdfs:    <http://www.w3.org/2000/01/rdf-schema#> .
+
+TBC
+    rdf:type owl:Ontology ;
+    dcterms:title "Works"@en ;
+    dcterms:created "13/02/2017"^^xsd:date ;
```
kenoir: Not a valid xsd:date

ontologies/works.ttl
```diff
@@ -0,0 +1,344 @@
...
+	
+
+#######  Classes ##### 	
+
+ww:Work rdf:type owl:Class ;
+	rdfs:label "Work"@en ;
+	rdfs:comment "An individual work such as a text, archive item or picture; or a grouping of individual works (so, for instance, an archive collection counts as a work, as do all the series and individual files within it).  Each work may exist in multiple instances (e.g. copies of the same book).  N.B. this is not synonymous with "work" as that is understood in the International Federation of Library Associations and Institutions' Functional Requirements for Bibliographic Records model (FRBR) but represents something lower down the FRBR hierarchy, namely manifestation. Groups of related items are also included as works because they have similar properties to the individual ones."@en ;
+	rdf:isDefinedBy <TBC/works> .
+	
+ww:Collection rdf:type owl:Class ;
```
kenoir: Remove for now, pending further thinking over next couple of months. Think we might favour a more graph-like model where we just have Work with many relationship predicates.

ontologies/works.ttl
```diff
@@ -0,0 +1,344 @@
...
+	rdfs:comment "A grouping of material comprising more than one work, the grouping being either organic (i.e., a Fonds) or artificial."@en ;
+	rdfs:subClassOf ww:Work
+	rdf:isDefinedBy <TBC/works> .
+	
+ww:Item rdf:type owl:Class ;
+	rdfs:label "Item"@en ;
+	rdfs:comment "A specific instance of a work: for instance, an individual physical copy with its own location.  This corresponds to the bottom, Instance, layer of FRBR.  Works that are unique - that is, have only one copy - are described in terms of a Work for their intellectual content, linked to an Item describing their physical characteristics."@en ;
+	rdf:isDefinedBy <TBC/works> .	
+	
+ww:Agent rdf:type owl:Class ;
```
kenoir: Move to separate agents ontology

ontologies/works.ttl
```diff
@@ -0,0 +1,344 @@
...
+	rdfs:label "Agent"@en ;
+	rdfs:comment "An individual or organisation that stands in some relationship to a work."@en ;
+	rdf:isDefinedBy <TBC/works> .
+
+ww:Format rdf:type owl:Class ;
+	rdfs:label "Format"@en ;
+	rdfs:comment "The format of a work: namely, whether it is a printed book, archive, image, moving image, etc."@en ;
+	rdf:isDefinedBy <TBC/works> .
+	
+ww:Location rdf:type owl:Class ;
```
kenoir: Move to separate ontology, call Place

ontologies/works.ttl
```diff
@@ -0,0 +1,344 @@
...
+	rdfs:label "Item"@en ;
+	rdfs:comment "A specific instance of a work: for instance, an individual physical copy with its own location.  This corresponds to the bottom, Instance, layer of FRBR.  Works that are unique - that is, have only one copy - are described in terms of a Work for their intellectual content, linked to an Item describing their physical characteristics."@en ;
+	rdf:isDefinedBy <TBC/works> .	
+	
+ww:Agent rdf:type owl:Class ;
+	rdfs:label "Agent"@en ;
+	rdfs:comment "An individual or organisation that stands in some relationship to a work."@en ;
+	rdf:isDefinedBy <TBC/works> .
+
+ww:Format rdf:type owl:Class ;
```
kenoir: Relationship to controlled work type list?

ontologies/works.ttl
```diff
@@ -0,0 +1,344 @@
...
+	rdfs:label "Location"@en ;
+	rdfs:comment "The place at which a work was published or otherwise created."@en ;
+	rdf:isDefinedBy <TBC/works> .
+	
+ww:Genre rdf:type owl:Class ;
+	rdfs:label "Genre"@en ;
+	rdfs:comment "The genre to which a work belongs: Genre being taken to mean a particular form carrying certain assumptions about content, authorial stance, etc.  Examples might include biography, article, obituary, or satire."@en ;
+	rdf:isDefinedBy <TBC/works> .	
+
+ww:IPRightsStatus rdf:type owl:Class ;
```
kenoir: Licence? Separate ontology?

ontologies/works.ttl
```diff
@@ -0,0 +1,344 @@
...
+	rdfs:label "IPRightsStatus"@en ;
+	rdfs:comment "The specific licence under which the work in question is released to the public - for example, one of the forms of Creative Commons - if it is a precise licence to which a link can be made."@en ;
+	rdf:isDefinedBy <TBC/works> .	
+	
+ww:Subject rdf:type owl:Class ;
```
kenoir: Separate ontology

kenoir:
@ChristopherHilton This has been open for a bit. What do we need to do to get it merged?

kenoir:
Resolves https://github.com/wellcometrust/platform-api/issues/77

Issue #53, opened 2017-02-15
Create library to do updates from Calm to DynamoDB


alexwlchan:
Our current script takes an XML export from Calm and shoves the whole thing into a new DynamoDB table.

We should use the Calm API to process updates (possibly using the “Modified” field) to update an existing DynamoDB table with new data.

This task only covers the “update existing data” part. Related: #52.

kenoir:
We should probably be able to do both.

alexwlchan:
Closed by #80.

PR #80, opened 2017-03-09
Add Calm ~> DynamoDB adapter

### LINKED ISSUE #52: Create library to transform data from Calm into DynamoDB
Fully-documented, tested script for taking an XML export for a complete reindex of the Calm data.

This task only covers the “insert data into a fresh table” part. Related: #53.

alexwlchan:
This adds the interesting logic for a Calm ~> DynamoDB investor. Current status:

- [x] Fetch first page of results from the OAI
- [x] Fetch additional pages based on the `resumptionToken`
- [x] Parse records/resumption token from the OAI XML
- [x] Turn records into objects that can be pushed into DynamoDB
- [x] Push those objects to DynamoDB
- [x] Rename this to "calm_adapter"
- [x] Rewrite this as a series of actors
- [ ] Add the ability to ingest the Calm XML dump

Because the OAI is incredibly slow, I think we eventually want to break this up into a series of actors – and then we don’t block waiting for the OAI. Initial thoughts on what those actors might look like:

calm_ingestor1

We can take a complete XML dump _or_ fetch incremental updates from OAI – but doing an initial ingest from the OAI will be much slower.

Resolves #52, #53.


calm_ingestor/src/main/scala/uk/ac/wellcome/transformer/controllers/ManagementController.scala
```diff
@@ -4,12 +4,82 @@ import javax.inject.{Inject, Singleton}
...
+import scala.util.matching.Regex
 
 @Singleton
 class ManagementController @Inject()() extends Controller {
   get("/management/healthcheck") { request: Request =>
     response.ok.json(Map("message" -> "ok"))
   }
+
+  get("/runfetch") { request: Request =>
+    val data = Http("http://localhost:5000")
```
alexwlchan: Because the OAI is so slow, I’m using a local server that just serves hard-coded data for now, I haven’t got to multiple pages yet.

alexwlchan:
I’ve just pushed a version that rewrites this as a series of actors, with broadly the same functionality as the old one. Still to do:

* [ ] Actually push records to DynamoDB
* [ ] Don’t ingest every record at once – we should ask Dynamo “what’s the latest record you already have”, then look at the modified timestamp (probably)
* [ ] Error handling. What if a record doesn’t ingest correctly?
* [ ] Docs

Having tried it now, it’s only the first call to the OAI that’s slow. Subsequent calls return extremely quickly. If we make the Dynamo table configurable, we could do a complete reingest by just pointing at an empty Dynamo table.

Plus a whole bunch of TODOs in comments.

calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/actors/OaiParserActor.scala
```diff
@@ -0,0 +1,135 @@
...
+  //     </resumptionToken>
+  //
+  val resumptionTokenPattern = "<resumptionToken[^>]*>([^>]+)</resumptionToken>".r("token")
+
+  // An OAI response may be an incomplete list with a resumptionToken
+  // value that allows us to make requests for the rest of the list.
+  // Look for such a value, and if present, start a new request.
+  //
+  // https://www.openarchives.org/OAI/openarchivesprotocol.html#FlowControl
+  def startNextRequest(response: String): Unit = {
```
alexwlchan: This should be handled in the `OaiHarvestActor` instead.

calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/actors/OaiParserActor.scala
```diff
@@ -0,0 +1,135 @@
...
+  //
+  //      <fieldName urlencoded="URL%20encoded%20field%20value">
+  //
+  // where the latter may have a closing slash if the value is empty.
+  // In the latter case, we want to capture the field name and value.
+  val streamParserPattern = "<(?:/record|([A-Za-z0-9]+) urlencoded=\"([^\"]*)\"/?)>".r("name", "value")
+
+  def parseRecords(data: String): Unit = {
+    // Stores the name/value pairs found so far in each record
+    var currentRecord: MutableMap[String, String] = MutableMap()
```
alexwlchan: Boo MutableMap, look at Akka agents instead.

PR #82, opened 2017-03-16
Add json ld context

jtweed:
Adds JSON-LD context and basic /works implementation that returns JSON-LD


api/src/main/resources/context.json
```diff
@@ -0,0 +1,8 @@
+{
+  "@context": {
+    "@base": "http://id.wellcomecollection.org/",
+    "@vocab": "http://wellcomecollection.org/ontologies/works/",
```
alexwlchan: By default any new stuff will be HTTPS – is it a hard requirement that these are HTTP?

api/src/main/scala/uk/ac/wellcome/platform/api/Server.scala
```diff
@@ -17,27 +20,33 @@ import io.swagger.models.Swagger
...
+  override def jacksonModule = ApiJacksonModule
 
   override def configureHttp(router: HttpRouter) {
     router
       .filter[CommonFilters]
       .filter[LoggingMDCFilter[Request, Response]]
       .filter[TraceIdMDCFilter[Request, Response]]
       .filter(CorsFilter())
-      .add[CollectionController]
+      .add[ContextController]
```
alexwlchan: nit: let’s put the `ContextController` towards the end – so our other Finatra apps looks semi-consistent: ManagementController, MainController, custom controllers.

api/src/main/scala/uk/ac/wellcome/platform/api/controllers/ContextController.scala
```diff
@@ -0,0 +1,21 @@
...
+import com.twitter.inject.annotations.Flag
+import com.twitter.finagle.http.Request
+import com.twitter.finatra.http.Controller
+
+import javax.inject.{Inject, Singleton}
+
+
+@Singleton
+class ContextController @Inject()(
+  @Flag("api.prefix") apiPrefix: String) extends Controller {
```
alexwlchan: nit: style elsewhere in the project would be to write

```scala
class ContextController @Inject()(
  @Flag("api.prefix") apiPrefix: String
) extends Controller {
```

which I also think is more readable.

Also, I’m slightly surprised we need the `@Flag("api.prefix")` here – maybe we should refactor all this into an `ApiConfig` module that we DI here? It’s a bit ugly. (See the `DynamoConfig` module in the transformer project for an example of what I’m talking about.)

api/src/main/scala/uk/ac/wellcome/platform/api/controllers/MainController.scala
```diff
@@ -2,6 +2,7 @@ package uk.ac.wellcome.platform.api.controllers
 
 import com.twitter.finagle.http.Request
 import com.twitter.finatra.http.Controller
+import uk.ac.wellcome.platform.api.models._
```
alexwlchan: Why move this import? Is this an IntelliJ thing?

(And are there rules for import order in Scala/Java? I’ve been vaguely adapting the Python rules.)

api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
```diff
@@ -8,12 +8,20 @@ import com.twitter.finatra.http.Controller
...
+import uk.ac.wellcome.platform.api.utils.ApiRequestUtils
 
 
 @Singleton
-class WorkController @Inject()(
-  @Flag("api.prefix") apiPrefix: String) extends Controller with SwaggerSupport {
+class WorksController @Inject()(
+  @Flag("api.prefix") apiPrefix: String,
+  @Flag("api.context") apiContext: String,
+  calmService: CalmService) extends Controller with SwaggerSupport {
```
alexwlchan: nit: ditto above

api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
```diff
@@ -37,7 +47,11 @@ class WorkController @Inject()(
         .routeParam[String]("id", "The work to return",  required = true)
         .responseWith[Object](200, "Work")
     } { request: Request =>
-      response.notImplemented
+      calmService.findRecordByAltRefNo(request.params("id")).map(resultOption =>
+        resultOption
+          .map(result => response.ok.json(
```
alexwlchan: Slightly cleaner to write

```scala
calmService.findRecordByAltRefNo(request.params("id")).map(
  _.map(result => response.ok.json(
```

api/src/main/resources/context.json
jtweed: There is no clear and widely adopted approach to this within the semantic web community. This is the best I can find from the W3C:

https://www.w3.org/blog/2016/05/https-and-the-semantic-weblinked-data/

And TimBL wants the https scheme deprecated:

https://www.w3.org/DesignIssues/Security-NotTheS.html

I have asked Data Language for the advice on the current usual practice.

api/src/main/scala/uk/ac/wellcome/platform/api/Server.scala
jtweed: Happy to. I'd just put it all in alphabetical order, but will change to this instead.

api/src/main/scala/uk/ac/wellcome/platform/api/controllers/ContextController.scala
jtweed: Will change the line layout, I'm style getting my head around Scala style...

@Flag is the Finatra way to do this, I introduced it as part of the Swagger support. So I'd vote to keep for now, then revisit when we come back to the API properly and decide if we want to use @Flag or not across the board.

api/src/main/scala/uk/ac/wellcome/platform/api/controllers/MainController.scala
jtweed: Unintentional. I accidentally removed it and put it back in a different place. I've been following a Python-esque order too, but I'm not too concerned about this class, as it needs removed (along with CalmService) when we start building this out next quarter.

api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
jtweed: Will reformat, but keep @Flag, as above.

api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
jtweed: Will change, despite flashbacks to overuse of $_ in dodgy Perl 😄

PR #84, opened 2017-03-20
Add a shutdown procedure for the Calm adapter

alexwlchan:
* When the OAI harvest actor gets its final response, it tears down itself and sends a teardown message to the OAI parser actor.
* When the OAI parser actor gets the teardown message, it tears down itself and sends a teardown message to the Dynamo actor.
* When the Dynamo actor gets the teardown message, it waits 60 seconds for any outstanding tasks to finish, then tears down the entire application.

I’ve tested with locally, with a small subset of OAI records – this does cause the warmup actor to clamp down the capacity.


alexwlchan:
Okay, now we send an appropriate SNS message in the shutdown() method of the CalmAdapterWorker.

PR #93, opened 2017-03-22
Add incremental updates to the Calm ~> DynamoDB adapter

### LINKED ISSUE #92: Add incremental updates to the Calm ~> DynamoDB adapter
When you run the Calm adapter, it will pull in _every_ record in Calm. This generates ~250k Kinesis stream events on the other side of our DynamoDB table, which will get expensive quickly.

We should modify the *OaiParserActor* so that it only pulls in records with an "updated" field that’s <N&nbsp;days old, where N is a number the user should specify.

alexwlchan:
This patch adds a new config flag `oaiDaysToFetch` (default&nbsp;3). Any records that occur more than $DAYS ago are discarded by the parser as uninteresting. To disable this behaviour and retrieve the entire set, set the parameter to –1.

This isn’t entirely ideal – we still grab the entire set from the OAI – but when fiddling with the `from` URL query parameters, I wasn’t able to get the data set I wanted, so doing the logic on our side is probably better for understanding and maintainability.

Additionally, the Dynamo warm-up limit now defaults to a safe number, so you can’t inadvertently create an expensive table while testing.

Resolves #92.

### DIFF
--- a/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/modules/OaiHarvestConfigModule.scala
+++ b/calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/modules/OaiHarvestConfigModule.scala
@@ -5,7 +5,7 @@ import javax.inject.Singleton
 import com.google.inject.Provides
 import com.twitter.inject.TwitterModule
 
-case class OaiHarvestConfig(oaiUrl: String)
+case class OaiHarvestConfig(oaiUrl: String, oaiDaysToFetch: Long)
 
 object OaiHarvestConfigModule extends TwitterModule {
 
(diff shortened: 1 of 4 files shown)


alexwlchan:
Anecdotal testing: running this over the entire data set, this makes just 481 Dynamo writes vs 256558 records in the complete table.

kenoir:
Tests fail as `DynamoDB table name must not be empty`

calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/modules/OaiHarvestConfigModule.scala
```diff
@@ -5,7 +5,7 @@ import javax.inject.Singleton
 import com.google.inject.Provides
 import com.twitter.inject.TwitterModule
 
-case class OaiHarvestConfig(oaiUrl: String)
+case class OaiHarvestConfig(oaiUrl: String, oaiDaysToFetch: Long)
```
kenoir: Move to models.

alexwlchan: The entire `OaiHarvestConfig` class?

kenoir: Just the `OaiHarvestConfig` model. In this case it doesn't need to be in common though:

Probably:

`calm_adapter/src/main/scala/uk/ac/wellcome/calm_adapter/models/`

alexwlchan:
Review markup made.

PR #97, opened 2017-03-23
Adds an Identifiable trait to generate IDs

kenoir:
Note: This does not generate unique IDs! The result of generating an ID
will need to be checked against existing IDs. This change simply
implements the ID generation rules:

- 8 characters
- Composed of digits and the Roman alphabet
- Case-insensitive
- No check digit
- No ambiguous characters

Some example IDs generated by this algorithm:

```
eppptsp9
xekcz4rb
svub7yhy
ybyj9p54
mthth6c2
9zz293n3
```

This gives us an ID space of around 850 billion (31^8) items ... do we think that's enough?

Resolves: https://github.com/wellcometrust/platform-api/issues/31

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/utils/IdentifierGenerator.scala
+++ b/common/src/main/scala/uk/ac/wellcome/utils/IdentifierGenerator.scala
@@ -1,0 +1,15 @@
+package uk.ac.wellcome.utils
+
+import scala.util.Random
+
+trait Identifiable {
+  val identifierLength = 8
+  val forbiddenLetters = List('o', 'i', 'l', '1')
+  val letterRange = ('1' to '9') ++ ('a' to 'z')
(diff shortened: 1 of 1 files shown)


kenoir:
Paging @jennpb to this PR.

jennpb:
@kenoir This seems like a sensible approach to me, given the removal of ambiguous numbers/letters and keeping it relatively short at 8 characters. This covers the case where someone might be reading the number aloud, as when reading it out over the phone to a member of staff (i.e. I'm looking at wellcomecollection.org/item/a23vyxz8).

alexwlchan:
Seems like we have an agreed approach for what the identifiers should look like. I’m going to merge this, then open a separate issue to track building the service that actually hands them out – because I don’t think we’re getting to that any time soon.

alexwlchan:
See issue #101 for follow-up.

PR #107, opened 2017-03-30
Added list of identifiers to UnifiedItem

alicefuzier:
Created first integration test for CalmService with embedded ElasticSearch. Removed source from UnifiedItem and added a list of Identifiers to it. Modified Record accordingly

### DIFF
--- a/api/src/test/scala/uk/ac/wellcome/platform/api/services/CalmServiceTest.scala
+++ b/api/src/test/scala/uk/ac/wellcome/platform/api/services/CalmServiceTest.scala
@@ -1,0 +1,31 @@
+package uk.ac.wellcome.platform.api.services
+
+import com.sksamuel.elastic4s.ElasticDsl.indexInto
+import com.sksamuel.elastic4s.testkit.ElasticSugar
+import org.scalatest.{AsyncFunSpec, Matchers}
+import uk.ac.wellcome.finatra.services.ElasticsearchService
+import uk.ac.wellcome.models.{Identifier, UnifiedItem}
+import uk.ac.wellcome.platform.api.models.Record
+
+class CalmServiceTest extends AsyncFunSpec with ElasticSugar with Matchers{
+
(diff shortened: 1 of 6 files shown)


api/src/test/scala/uk/ac/wellcome/platform/api/services/CalmServiceTest.scala
```diff
@@ -0,0 +1,31 @@
...
+import uk.ac.wellcome.finatra.services.ElasticsearchService
+import uk.ac.wellcome.models.{Identifier, UnifiedItem}
+import uk.ac.wellcome.platform.api.models.Record
+
+class CalmServiceTest extends AsyncFunSpec with ElasticSugar with Matchers{
+
+  val elasticSearchService = new ElasticsearchService(client)
+  val calmService = new CalmService(elasticSearchService)
+
+  it("should find records") {
```
kenoir: ✨ ✨ tests!✨ ✨

transformer/src/main/scala/uk/ac/wellcome/transformer/actors/PublishableMessageRecordActor.scala
```diff
@@ -34,7 +34,7 @@ class PublishableMessageRecordActor @Inject()(
       JsonUtil.toJson(unifiedItem) match {
         case Success(stringifiedJson) => {
           val message = SNSMessage(
-            Some(unifiedItem.source),
+            Some("Foo"),
```
kenoir: Maybe `Some("calm")` to reduce levels of "Foo".

kenoir:
@alicefuzier Merge and deploy is up to you :)

Pinging @jtweed to notify him that this will break the API (if we deploy that service) as the current data in the index does not match this model. 

This model is however more correct and the pipeline is _almost_ ready to run data through in the correct format!

jtweed:
Fantastic thanks!

PR #110, opened 2017-03-31
WIP Use Futures instead of actors

alicefuzier:
Actors add extra complexity that we don't need since all the functionality we need is already provided by Futures. So we're refactoring the actors from the bottom up to be replaced by futures


transformer/src/main/scala/uk/ac/wellcome/transformer/modules/StreamRecordProcessorFactoryModule.scala
```diff
@@ -37,21 +34,20 @@ class StreamsRecordProcessor(
     records: JList[Record],
     checkpointer: IRecordProcessorCheckpointer
   ): Unit = {
-    records.asScala.map { record =>
-      reciever.map(_ ! record)
+    records.asScala.foreach { record =>
+      receiver.receiveRecord(record)
```
alexwlchan:
```scala
records.asScala.map(receiver.receiveRecord)
```

Would that work? I think that’s equivalent and looks a little cleaner, but there may well be a subtlety I’m missing.

transformer/src/main/scala/uk/ac/wellcome/transformer/receive/RecordReceiver.scala
```diff
@@ -0,0 +1,93 @@
...
+
+    info(s"Received record ${keys}")
+    RecordMap(keys)
+  }
+
+
+  def receiveRecord(record: RecordAdapter): Future[Unit] = {
+    for {
+      recordMap   <- recordToRecordMap(record)
+      o           <- extractDynamoCaseClass(recordMap)
```
alexwlchan: Can we come up with a better variable name than `o`?

PR #148, opened 2017-04-03
Initial commit of the Miro adapter

### LINKED ISSUE #120: Create adapter for Miro data in S3

alexwlchan:
Resolves #120.

### DIFF
--- a/adapter_miro/utils.py
+++ b/adapter_miro/utils.py
@@ -1,0 +1,47 @@
+# -*- encoding: utf-8 -*-
+
+
+def _render_value(value):
+    """Renders a value."""
+    if value is None or value == 'None':
+        return None
+    else:
+        try:
+            return value.strip()
+        except AttributeError:
+            return value
+
+
+def _elem_to_list(elem):
+    """Convert an lxml list-like Element to a Python list.
+
+    Some of the Miro data stores lists in XML by using ``<_>`` keys, e.g.:
+
+        <fruit>
+            <_>apple</_>
+            <_>banana</_>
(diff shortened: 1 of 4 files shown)


adapter_miro/utils.py
```diff
@@ -0,0 +1,47 @@
...
+        # Miro stores lists with <_> keys, so if we spot one, this element
+        # should actually be treated as a list.
+        if name == '_':
+            return _elem_to_list(elem)
+
+        if child.getchildren():
+            res[name] = elem_to_dict(child)
+        else:
+            res[name] = _render_value(child.text)
+    return res
```
alexwlchan: What about

```xml
<apple>
  <key>value</key>
  <key>value2</key>
</apple>
```

```json
{"key": "value2"}
```

adapter_miro/utils.py
```diff
@@ -0,0 +1,47 @@
+# -*- encoding: utf-8 -*-
+
+
+def _render_value(value):
+    """Renders a value."""
+    if value is None or value == 'None':
+        return None
+    else:
+        try:
+            return value.strip()
```
alexwlchan: `or None`

alexwlchan:
Markups made: we spot lists that don’t use an `_`, or dicts with non-unique keys. Either will just cause the entire script to throw an `AssertionError`, because whatever it is will most likely require human intervention, and doing something “clever” is liable to go wrong.

@alicefuzier Now you’ve reviewed this, please could I trouble you for a green tick so I could get it merged? =)

PR #158, opened 2017-04-06
Info logs by default

alexwlchan:
Debug logs in Cloudwatch are a pain to wade through and arguably add minimal to no value.

On the rare occasions they might be useful, we can turn them back on, but I think we should have a default policy of INFO logs and above.

### DIFF
--- a/api/src/main/resources/logback.xml
+++ b/api/src/main/resources/logback.xml
@@ -8,7 +8,7 @@
     </encoder>
   </appender>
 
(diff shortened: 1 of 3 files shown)


alexwlchan:
The one thing it would be useful to log at INFO level would be the config flags passed to an application when it starts, which are currently DEBUG only.

kenoir:
Worth thinking about what happens to logs if we log out config with secrets in.

PR #149, opened 2017-04-03
Creating idminter service

alicefuzier:
Creating a service that generates IDs for resources and checks for overlaps in DynamoDB. 
Addresses issue #123

PR #163, opened 2017-04-10
Remove id from UnifiedItem

alicefuzier:
To be merged after #149 
As the canonicalId has become part of the IdentifiedUnifiedItem - which wraps the UnifiedItem - the UnifiedItem should not have the "id" field any longer


alexwlchan:
🎉

PR #164, opened 2017-04-10
Add everything in the .idea folder in .gitignore

alicefuzier:
IntelliJ creates lots of files in the .idea folder and listing all the useless ones is tedious. All the information needed to import a project in IntelliJ should be provided by sbt anyway

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+.idea
+*.iml
+
 # Created by https://www.gitignore.io/api/java,scala,intellij,linux,vim,osx,python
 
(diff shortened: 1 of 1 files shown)


pollecuttn:
@kenoir You're meant to be not working! :o)

alexwlchan:
I am 👍 on the overall idea, but 👎 on mixing changes with the autogenerated gitignore file from gitignore.io.

Stick `.idea` at the top of the file and leave the original intact, and you’d get a green tick from me.

PR #165, opened 2017-04-11
Make demo item media URL https

jamesgorrie:
Making the URL for the demo item media HTTPS so we can use it over HTTPs on the client.

## Who is this change for?
@alexwlchan @alicefuzier


alexwlchan:
Having a few issues with builds right now, but once they’re fixed I’ll get this deployed.

PR #179, opened 2017-04-12
transformer with extra 🐍

alexwlchan:
I was getting a bit frustrated with the Scala transformer, so I knocked up a Python-based transformer on my lunch break. We probably shouldn’t abandon Scala at this stage, but putting it out there to see what people think.

### DIFF
--- a/transformer.py
+++ b/transformer.py
@@ -1,0 +1,98 @@
+#!/usr/bin/env python
+# -*- encoding: utf-8 -*-
+
+import json
+import operator
+import os
+import time
+
+import boto3
+
+
+def get_new_records(TableName):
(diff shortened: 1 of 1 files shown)


alexwlchan:
Things that are missing, among others:

* Packaging
* Error handling – any exception will tear down the entire process
* Tests

kenoir:
I think as this is transforming to a type shared by other apps this should definitely be Scala.

alicefuzier:
@kenoir I agree, but shouldn't you be on holiday??

kenoir:
I'm just contributing to an interesting open source project! Very relaxing.

PR #180, opened 2017-04-12
Future refactoring

alicefuzier:
Futures should be used to wrap IO bound operations, not CPU only operations. Also, blocking operations should be wrapped in a blocking construct so that the ExecutionContext can deal with them.
## Who is this change for?
Hopefully make transformer and any other app using futures more performant and more stable

### DIFF
--- a/transformer/src/main/scala/uk/ac/wellcome/transformer/receive/RecordReceiver.scala
+++ b/transformer/src/main/scala/uk/ac/wellcome/transformer/receive/RecordReceiver.scala
@@ -11,31 +11,38 @@ import uk.ac.wellcome.utils.GlobalExecutionContext.context
 import uk.ac.wellcome.utils.JsonUtil
 
 import scala.concurrent.Future
-import scala.util.{Failure, Success}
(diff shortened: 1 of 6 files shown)


transformer/src/main/scala/uk/ac/wellcome/transformer/receive/RecordReceiver.scala
```diff
@@ -11,47 +11,58 @@ import uk.ac.wellcome.utils.GlobalExecutionContext.context
...
       case Right(calmDynamoRecord) =>
         info(s"Parsed DynamoDB record $calmDynamoRecord")
         calmDynamoRecord
       case Left(dynamoReadError) =>
         error(s"Unable to parse record ${record.value}")
         throw new Exception(
           s"Unable to parse record ${record.value} received $dynamoReadError")
+    }.recover {
+      case e: Throwable =>
+        error("Error extracting transforable case class", e)
```
alexwlchan: /s/transforable/transferable

PR #182, opened 2017-04-13
[WIP] Try to get faster builds

alexwlchan:

Make our builds go faster!

### Who is this change for?

Bored developers!

---

Notes:

* Trashing the `common/target` and `~/.ivy2` directories on my local machine causes a compile to go from 22s to 446s (!!).

### DIFF
--- a/cache_containers.sh
+++ b/cache_containers.sh
@@ -7,13 +7,15 @@ set -o verbose
 
 IMAGES="peopleperhour/dynamodb alicefuzier/fake-sns s12v/elasticmq"
 
+mkdir -p ~/.ivy2/docker
 for img in $IMAGES
 do
-    if [[ -f ~/.ivy2/docker/$img.tar ]]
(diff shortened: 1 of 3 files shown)


alexwlchan:
Plan is to get a couple of runs of a complete build on Circle to see if this makes a difference, and merge if it does.

alexwlchan:
Initial two runs, we save about 6 minutes from caching.

Some other ideas:

* Cache the Docker images used in the tests
* Don’t stop and restart the Docker containers between tasks
* Don’t run a stack of separate sbt tasks each time; we’re paying the cost of starting the JVM multiple times

alexwlchan:
Take advantage of Circle’s parallelism?

kenoir:
If we intend to move to Jenkins, we should try and make sure changes here are portable.

alexwlchan:
Superseded by #188.

Issue #175, opened 2017-04-12
Speed up docker image creation


alicefuzier:
Currently our builds take a long time and most of that time is being spent in creating the docker image. We should investigate if we can optimize it so that it takes less time

alexwlchan:
[Latest build](https://travis-ci.org/wellcometrust/platform-api/builds/224362899) took 9 minutes for a complete set; no more than 5½ minutes to get an individual rebuild. I reckon that’s fast enough for now.

kenoir:
Wow. Good work.

PR #188, opened 2017-04-20
Switch to using Travis for significantly faster CI builds

### LINKED ISSUE #169: Switch to a CI that allows us to build apps individually

### LINKED ISSUE #173: Get Jenkins working in ECS

alexwlchan:

This swaps out Circle CI for Travis in our builds.

Resolves #169, #175.

This doesn’t resolve #173 – we’re only doing builds in Travis, not Terraform changes.

## Who is this change for?

Developers who are bored of slow builds.

You can see [Travis in action](https://travis-ci.org/alexwlchan/platform-api) on my fork. Switching to Travis comes with several advantages:

* **Easily configurable parallel builds.** In theory Circle supports parallelisation, but I haven’t been able to work out how to use it, and even if I could we can only do four builds at a time. Travis can do five on the free plan.
* **Performance.** Parallelisation and caching mean that Travis builds are fast – once the cache is warmed, I can build a new set of images in ECR in less than 7 minutes. For comparison, the [last Circle build on master](https://circleci.com/gh/wellcometrust/platform-api/465) took 43 minutes.
* **Granularity.** Because each project is a separate build task, you can rebuild it separately and get a new image in ECR within ~4m.

There are potentially more gains to be had if we cache even more (e.g. the result of the `publishLocal` task), but a 6x speedup is already pretty nice. This also lets us throw away some of our custom configuration for Circle, yay.


alexwlchan:
This also supersedes #182. I managed to get a full test run on Circle down to 14 minutes, which is still twice the time it takes to do a full deploy on Travis.

.travis.yml
```diff
@@ -0,0 +1,60 @@
...
+    - scala: 2.12.2
+      env: PROJECT=api
+    - scala: 2.12.2
+      env: PROJECT=transformer
+    - scala: 2.12.2
+      env: PROJECT=calm_adapter
+    - scala: 2.12.2
+      env: PROJECT=ingestor
+    - scala: 2.12.2
+      env: PROJECT=id_minter
```
alicefuzier: minor thing and possibly wrong...are we not using scala 2.11.8?

.travis.yml
```diff
@@ -0,0 +1,60 @@
...
+
+script:
+  # https://graysonkoonce.com/getting-the-current-branch-name-during-a-pull-request-in-travis-ci/
+  - export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo $TRAVIS_BRANCH; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
+  - echo "TRAVIS_BRANCH=$TRAVIS_BRANCH, PR=$PR, BRANCH=$BRANCH"
+  - sbt "project $PROJECT" dockerComposeTest
```
alicefuzier: we should change this as we did in run_circle.sh to be 
```
sbt "project $PROJECT" dockerComposeUp
sbt "project $PROJECT" test
sbt "project $PROJECT" dockerComposeStop
```
dockerComposeTest does not run all the tests for some reasons

.travis.yml
alexwlchan: \>_>

It seems to be inconsistent between local and remote:

```console
$ sbt about
[info] Set current project to terraform (in build file:/Users/chana/repos/platform-infra/terraform/)
[info] This is sbt 0.13.13
[info] The current project is {file:/Users/chana/repos/platform-infra/terraform/}terraform 0.1-SNAPSHOT
[info] The current project is built against Scala 2.10.6
```

From the Circle logs:

```
Resolving org.scala-lang#scala-library;2.11.8 ...
```

2.12.2 is definitely wrong (I’m not sure where I got that from), but we should look to see if we can make set consistent.

alexwlchan:
> do you still have to run all of them even if you have a change in only one of the project?

For now, yes. None of Travis, Circle or GoCD have a way to make tasks dependent on source changes unless you write the logic yourself.

It would be possible to adapt the stuff from the Circle shell scripts to run in Travis, so tasks exit immediately if there aren’t any source changes. Even faster!

> will this help us in having independent deploys as well?

I don’t think this makes it easier or harder.

.travis.yml
alicefuzier: I think the Circle logs is correct. If you see https://github.com/wellcometrust/platform-api/blob/master/project/Common.scala#L8 we explicitly declare 2.11.8 as our scala version. Do we need to specify it in the travis.yaml as well since it's defined in our sbt build? Can we not avoid the repetition?

As for the sbt about output, that seems to be run within platform-infra which is not an sbt project, so it doesn't have any scalaVersion defined. Since it's not defined the default for sbt is 2.10 cause sbt is built with 2.10.

alexwlchan: > Do we need to specify it in the travis.yaml as well since it's defined in our sbt build? Can we not avoid the repetition?

Maybe not, I still see 2.11.8 in several places in the build output.

> that seems to be run within platform-infra which is not an sbt project, so it doesn't have any scalaVersion defined

Oops, yep.

alexwlchan:
Travis build times got clobbered because removing the Scala version means it lost all the caches, hopefully should be green soon.

alexwlchan:
I’ll rebase against master, then I’ll merge.

kenoir:
Sweet.

PR #192, opened 2017-04-24
Make sure we build prod images in Travis

alexwlchan:

Fix our builds to be prod, not dev. As spotted by @alicefuzier, we’re currently building images with the "dev" tag.

As a nice side effect, this means that (AFAICT) all of our Travis config is now defined in a checked-in file, and not squirrelled away in the GUI settings.

## Who is this change for?

Developers who want to deploy applications!


alexwlchan:
Ho hum, this doesn't seem to have worked. I'll continue digging. 😕

PR #193, opened 2017-04-24
Build Docker images without using sbt-docker

alexwlchan:

Allow us to load config from an S3 bucket directly at startup. This means we don’t need to rebuild an application for config changes, and reduces our external dependencies. In future, this means that:

* We could more easily run the same application with different config sets if necessary (e.g. the transformer)
* Because config is loaded at startup, we can trigger a restart on a `terraform apply` by having a Lambda watch an S3 object for changes

## Who is this change for?

Developers.

- [ ] Tests 
- [ ] Docs?


.travis.yml
```diff
@@ -50,18 +50,22 @@ script:
     if [[ "$BRANCH" == "master" ]];
     then
       if [[ "$PROJECT" != "common" ]]; then
-        sbt "project $PROJECT" -DconfigBucket=$CONFIG_BUCKET -Denv=$BUILD_ENV ecr:push;
+        sbt "project $PROJECT" stage;
+        export NAME="0.0.1-$(git rev-parse HEAD)_alex";
```
alexwlchan: This name is deliberate – at least initially, it will produce broken images, and I don’t want to deploy to prod until I’ve got the infra pieces in place as well.

alexwlchan:
This will require changes on the infra side as well, so when we hit "Merge" a bunch of stuff will probably break. Still to do:

* [ ] Swap out .conf for .ini files in S3
* [ ] Add S3 read permissions to our applications
* [ ] Rip out all the old ecr-plugin config

alexwlchan:
Rebased against current master.

PR #222, opened 2017-05-03
Ensure the Calm adapter tries to shut down the right service

alexwlchan:

Get the Calm adapter to shut down correctly. We renamed the service, and now it breaks the Lambda which tries to shut it down. I don't think it's worth making this configurable because the service name should rarely change.

## Who is this change for?

Internal.

- [x] Tests – not required
- [x] Docs – not required
- [x] Spoken to the right people – no external dependencies


alexwlchan:
Rebased against master.

PR #223, opened 2017-05-03
Namespace release files

kenoir:

In order that they can be downloaded more easily (as `aws s3` does not
support wildcards).

## Who is this change for?

Devloper

- [X] Tests?
- [X] Docs?


alexwlchan:
👍 for scripting the creation of a .tfvars file

Issue #225, opened 2017-05-03
IDs shouldn't start with a number


jtweed:
IDs currently can start with a number, but this makes using them as a string harder in certain contexts. Let's put in a rule that the first character has to be a letter.

alexwlchan:
Two parts to this:

- [ ] Modify the existing generator to spit out IDs that don't start with numbers
- [ ] Flush the existing identifiers table

PR #226, opened 2017-05-03
ID minter only generates IDs that start with letters

alexwlchan:

Addresses the first half of #225 – new IDs won't start with a number.

Once this is merged, we'll need to flush the existing tables, as have some existing identifiers that begin with numbers.

## Who is this change for?

End-users who may be using identifiers as strings, for whom a leading numeral is potentially problematic.

- [x] Tests? See unit test in patch
- [x] Docs? Not required.

### DIFF
--- a/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/utils/Identifiable.scala
+++ b/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/utils/Identifiable.scala
@@ -16,9 +16,12 @@ object Identifiable {
   def generate = {
     (1 to identifierLength)
       .map(x => x match {
-        // In certain contexts, IDs are used as a string.  This can cause
-        // problems if the first character is a digit, so we have a rule
-        // that the first character must be a letter.
+        // One of the serialization formats of RDF is XML, so for
+        // compatibility, our identifiers have to comply with XML rules.
+        // XML identifiers cannot start with numbers, so we apply the same
+        // rule to the identifiers we generate.
+        //
+        // See: http://stackoverflow.com/a/1077111/1558022
         case 1 => firstCharacterSet(Random.nextInt(firstCharacterSet.length))
         case _ => allowedCharacterSet(Random.nextInt(allowedCharacterSet.length))
       })
--- a/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/utils/Identifiable.scala
+++ b/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/utils/Identifiable.scala
@@ -4,12 +4,24 @@ import scala.util.Random
 
 object Identifiable {
   private val identifierLength = 8
-  private val forbiddenLetters = List('o', 'i', 'l', '1')
-  private val letterRange = ('1' to '9') ++ ('a' to 'z')
-  private val letterSet = letterRange.filterNot(forbiddenLetters.contains)
 
-  def generate =
+  private val forbiddenLetters = List('o', 'i', 'l', '1')
+  private val numberRange = ('1' to '9')
+  private val letterRange = ('a' to 'z')
+
+  private val characterSet = numberRange ++ letterRange
+  private val allowedCharacterSet = (characterSet).filterNot(forbiddenLetters.contains)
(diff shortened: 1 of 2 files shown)


id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/utils/IdentifiableTest.scala
```diff
@@ -16,4 +16,15 @@ class IdentifiableTest extends FunSpec with PropertyChecks with Matchers {
     }
   }
 
+  // We have a rule that identifiers can't start with a number, as this
+  // makes them awkward to use as a string in certain contexts.
+  it("should never generate an identifier that starts with a number") {
+    forAll(minSuccessful(100)) { (_: Int) =>
+      {
+        val id = Identifiable.generate
+        assert(!id.head.isDigit)
```
alicefuzier: I slightly prefer using matchers rather than asserts. So something like 
```[scala]
id.head.isDigit should be (false)
```

id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/utils/IdentifiableTest.scala
```diff
@@ -16,4 +16,15 @@ class IdentifiableTest extends FunSpec with PropertyChecks with Matchers {
     }
   }
 
+  // We have a rule that identifiers can't start with a number, as this
+  // makes them awkward to use as a string in certain contexts.
+  it("should never generate an identifier that starts with a number") {
```
alicefuzier: Not sure what everyone thinks, but I think we should have one comment either in the test or in the actual code explaining why we do this. Not in both places as it is harder to ensure we keep things up to date. I personally prefer it being in tests as one of the purposes of tests should be documentation

id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/utils/Identifiable.scala
```diff
@@ -4,12 +4,24 @@ import scala.util.Random
...
+  private val characterSet = numberRange ++ letterRange
+  private val allowedCharacterSet = (characterSet).filterNot(forbiddenLetters.contains)
+  private val firstCharacterSet = allowedCharacterSet.filterNot(numberRange.contains)
 
-  def generate =
+  def generate = {
     (1 to identifierLength)
-      .map(_ => letterSet(Random.nextInt(letterSet.length)))
+      .map(x => x match {
+        // In certain contexts, IDs are used as a string.  This can cause
```
alicefuzier: It doesn't really have to do with IDs as strings, but as IDs used in XML. As one of the serialization formats of RDF is XML, it has to comply to XML rules. Apparently rules for XML IDs are that they cannot start with numbers: http://stackoverflow.com/questions/1077084/what-characters-are-allowed-in-dom-ids . As an example, Jena - one of the RDF frameworks - just assumes that everything in the model follows this rule, so it ignores numbers at the beginning of ids (crappy behavior, I know)

id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/utils/IdentifiableTest.scala
alexwlchan: 👍

I’ve taken a habit from Python – you generally refer to boolean results explicitly – but you’re right, consistency with the existing code should win out here.

alicefuzier: Actually scalatest matchers support regex so you can write something like
```[scala]
id should not startWith regex "[0-9]"
```

id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/utils/IdentifiableTest.scala
alexwlchan: 👍 for comment in only one place. I prefer the comment alongside the code, so that it’s all in one place, rather than split across multiple tests – although I don’t think it makes a lot of difference.

alexwlchan:
Rebased against master, review markups made.

alexwlchan:
I’ll merge #227, rebase against updated master, then merge this once tests pass.

PR #227, opened 2017-05-03
Update the JSON-LD context

### LINKED ISSUE #132: Ensure JSON-LD context is up to date and valid
Update `context.json` to latest mapping (see below).

We also need to fix the `@context` value, as on live it's coming through as:

`https://app:8888/catalogue/v0/context.json`

instead of:

`https://api.wellcomecollection.org/catalogue/v0/context.json`

alexwlchan:

Update `context.json` to the latest mapping, per #132. Fixes the `@context` value on API responses. See https://api.wellcomecollection.org/catalogue/v0/works/2bp9dawj for an example.

Resolves #132.

## Who is this change for?

Users of the API.

- [x] Tests? Not required.
- [x] Docs? Not required.

### DIFF
--- a/api/src/main/scala/uk/ac/wellcome/platform/api/utils/ApiRequestUtils.scala
+++ b/api/src/main/scala/uk/ac/wellcome/platform/api/utils/ApiRequestUtils.scala
@@ -11,9 +11,21 @@ object ApiRequestUtils {
       case _ => "http"
     }
 
-    val hostHeader = request.host match {
-      case Some(host) => host
-      case _ => throw new BadRequestException("Host header not set")
+    // Most users will access our API through a DNS record that comes through
+    // https://api.wellcomecollection.org.  I can't find a way to detect
+    // the DNS name through the request headers, so just hard-code it if we
+    // detect the request came through ELB.
+    //
+    // Specifically, we look for a header that is added to all ELB requests:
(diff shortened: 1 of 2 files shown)


alexwlchan:
Rebased against master; will merge once tests pass.

api/src/main/scala/uk/ac/wellcome/platform/api/utils/ApiRequestUtils.scala
```diff
@@ -11,9 +11,21 @@ object ApiRequestUtils {
...
-      case Some(host) => host
-      case _ => throw new BadRequestException("Host header not set")
+    // Most users will access our API through a DNS record that comes through
+    // https://api.wellcomecollection.org.  I can't find a way to detect
+    // the DNS name through the request headers, so just hard-code it if we
+    // detect the request came through ELB.
+    //
+    // Specifically, we look for a header that is added to all ELB requests:
+    // http://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html
+    val hostHeader: String = request.headerMap.get("X-Amzn-Trace-Id") match {
```
kenoir: Might it make more sense to have the DNS name in config (and defaulted to localhost)?

kenoir: Like this -> https://github.com/wellcometrust/platform-api/pull/228

PR #231, opened 2017-05-04
Update context.json

jtweed:
Wrap array of contexts in a context object, to match the JSON-LD spec.

Make the JSON-LD context valid

## Who is this change for?

Those consuming the API as Linked Data

- [ X] Tests?
- [ X] Docs?
- [X ] Spoken to the right people?

### DIFF
--- a/api/src/main/resources/context.json
+++ b/api/src/main/resources/context.json
@@ -1,26 +1,28 @@
-[
-  {
-    "@context": {
-      "@base": "http://id.wellcomecollection.org/",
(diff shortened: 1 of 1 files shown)


jtweed:
Actually going to change this again. The spec is really unclear, but I think I can remove the `@context` within each array item.

jtweed:
Sorry, missed the question around validation. I think some wider validation tests that check responses can be processed with the context would be useful.

Issue #234, opened 2017-05-04
Naming things is a Hard Problem™


alexwlchan:
We have a class called `UnifiedItem` that represents the JSON-LD type `Work`. There will also be a (different) JSON-LD type `Item`. Letting these names stand will obviously be fine and not be the cause of any hilarious bugs.

Let's sort out the naming.

* [x] Rename the existing `UnifiedItem` type to `Work`.
* [x] Rename the Elasticsearch indices to that works live in an index called "works".
* [x] Add a Jackson annotation so that serialising `Work` to a JSON object includes `"type": "work"`.

alexwlchan:
I’ve updated our `.tfvars` file in S3 to point everything at a new "works" index. This will shortly cause the API to be pointing at an empty index, but that's okay, because it gives us a good test of #194 – do we see 18941 records after we've reingested everything?

alexwlchan:
And I’ve deleted the old index. *waits for everything to migrate over*

PR #235, opened 2017-05-04
/s/UnifiedItem/Work

alexwlchan:
This renames all our references to "UnifiedItem" to "Work". Essentially just the result of running

```console
$ sed -ie 's/UnifiedItem/Work/g' **/*.scala
```

over the codebase, and renaming a small number of files.

The change is for developers working on the API – this means that our internal class names match the types we're sending in JSON-LD.

This addresses the first part of #234 – I’ll put the Jackson and Elasticsearch changes in a separate patch.


kenoir:
@alexwlchan CI fail is because:

```
... /ingestor/services/IdentifiedWorkIndexerTest.scala:29: reassignment to val
```

PR #236, opened 2017-05-04
Add initial Elasticsearch mappings

kenoir:

In order to only ingest valid data, and to optimise our indexes for search, we should have Elasticsearch mappings defined.

In this first pass we are looking to accommodate Miro data.

## Who is this change for?

Anyone who likes structured data.

- [x] Tests?
- [x] Docs?


kenoir:
`elastic4s` seems under rapid development so finding the right way to define mapping involves ensuring you've pinned the repo at the correct version and sometimes trawling the code.

common/src/main/scala/uk/ac/wellcome/lib/elasticsearch/mappings.scala
```diff
@@ -0,0 +1,31 @@
+package uk.ac.wellcome.lib.elasticsearch
+
+import com.sksamuel.elastic4s._
+import com.sksamuel.elastic4s.ElasticDsl._
+import com.sksamuel.elastic4s.mappings.FieldType._
+import com.sksamuel.elastic4s.analyzers._
+
+// TODO: This should have a test before any more code is written.
```
kenoir: Sorry ... should have written the tests first. I wasn't sure if this was even possible for our purposes with the elastic4s library, but it seems it is.

kenoir:
We'll want to add created and updated datetime fields for the doc is ES too.

common/src/main/scala/uk/ac/wellcome/elasticsearch/mappings/RecordsIndex.scala
```diff
@@ -0,0 +1,26 @@
+package uk.ac.wellcome.elasticsearch.mappings
+
+import com.sksamuel.elastic4s.ElasticDsl._
+import com.sksamuel.elastic4s._
+import com.sksamuel.elastic4s.analyzers._
+import com.sksamuel.elastic4s.mappings.DynamicMapping
+
+class RecordsIndex(client: ElasticClient, indexName: String) {
```
kenoir: We should test at the very least that this creates an index appropriately named.

alicefuzier:
@kenoir Yeah, about tests for the RecordsIndex I wasn't quite sure what to test cause testing that after the creation it has an index with those properties seemed kind of pointless. 
All tests against elasticsearch use the ElasticSearchLocal class so I added a call to our mapping scrpit in there, so that each test in the ingestor or the api runs with an elasticsearch provisioned that way. 
This tests both that an index exist with that name and that the APi and the ingestor can use that index, so it's actually a greater coverage. 
I did add a RecordsIndexTest class in the end cause there was no other way to test if elasticsearch would reject an item in the wrong format

common/src/main/scala/uk/ac/wellcome/elasticsearch/mappings/RecordsIndex.scala
```diff
@@ -0,0 +1,26 @@
+package uk.ac.wellcome.elasticsearch.mappings
+
+import com.sksamuel.elastic4s.ElasticDsl._
+import com.sksamuel.elastic4s._
+import com.sksamuel.elastic4s.analyzers._
+import com.sksamuel.elastic4s.mappings.DynamicMapping
+
+class RecordsIndex(client: ElasticClient, indexName: String, itemType: String) {
```
alexwlchan: Maybe call this "WorksIndex"?

common/src/test/scala/uk/ac/wellcome/elasticsearch/mappings/WorksIndexTest.scala
```diff
@@ -0,0 +1,46 @@
...
+import org.elasticsearch.transport.RemoteTransportException
+import org.scalatest.FunSpec
+import org.scalatest.concurrent.ScalaFutures
+import uk.ac.wellcome.models.{IdentifiedWork, SourceIdentifier, Work}
+import uk.ac.wellcome.test.utils.ElasticSearchLocal
+import uk.ac.wellcome.utils.GlobalExecutionContext.context
+import uk.ac.wellcome.utils.JsonUtil
+
+class WorksIndexTest extends FunSpec with ElasticSearchLocal with ScalaFutures{
+  // RecordsIndex is actually used in ElasticSearchLocal to create the index.
```
alexwlchan: comment refers to the old name

kenoir:
Looks good.

PR #238, opened 2017-05-05
Add Travis build badge

kenoir:

Explicitly show project build status.

## Who is this change for?

Folk interested in whether the most recent version of `master` is passing its' build.

- [X] Tests?
- [X] Docs?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # Digital Platform
 The new Wellcome Collection digital platform, exploring approaches to building an API that searches our archive collections.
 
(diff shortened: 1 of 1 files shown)


alexwlchan:
Personally not a fan of build badges – I think they look ugly, and really master should never be failing.

But I don't care strongly enough to argue if you and @alicefuzier disagree.

kenoir:
> and really master should never be failing.

Exactly, _but it can_.

kenoir:
Pinging @jtweed as this is a public perception thing too.

jtweed:
I quite like it, I think it shows that we care that the build is always passing. That's a positive perception, provided the build is always passing of course!

PR #240, opened 2017-05-05
Add the "type" field to the Work class when serialising to JSON

alexwlchan:

When we serialise a `Work` object to JSON, it includes a "type" field that contains the value "Work". Part of #234.

## Who is this change for?

Users of the API.

- [x] Tests – done
- [x] Docs – not required.

### DIFF
--- a/api/src/test/scala/uk/ac/wellcome/platform/api/ModelsTest.scala
+++ b/api/src/test/scala/uk/ac/wellcome/platform/api/ModelsTest.scala
@@ -1,0 +1,16 @@
+package uk.ac.wellcome.platform.api
+
+import org.scalatest.{FunSpec, Matchers}
+
+import uk.ac.wellcome.models.Work
+import uk.ac.wellcome.utils.JsonUtil
+
+
+class WorkTest extends FunSpec with Matchers {
+  it("should have an LD type 'Work' when serialised to JSON") {
+    val work = Work(identifiers=List(), label="A book about a thing")
+    val jsonString = JsonUtil.toJson(work).get
+
+    jsonString.contains("""type":"Work"""") should be (true)
+  }
+}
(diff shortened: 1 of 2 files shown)


alexwlchan:
Note: this may well break things coming out of the API, but I'll likely be making changes there as I modify the Miro transformation/Elasticsearch mappings.

api/src/test/scala/uk/ac/wellcome/platform/api/ModelsTest.scala
```diff
@@ -0,0 +1,17 @@
+package uk.ac.wellcome.platform.api
+
+import org.scalatest.{FunSpec, Matchers}
+import org.scalatest.prop.PropertyChecks
+
+import uk.ac.wellcome.models.Work
+import uk.ac.wellcome.utils.JsonUtil
+
+
+class WorkTest extends FunSpec with PropertyChecks with Matchers {
```
alicefuzier: do you need PropertyChecks?

common/src/main/scala/uk/ac/wellcome/models/Work.scala
```diff
@@ -12,9 +13,13 @@ case class SourceIdentifier(source: String, sourceId: String, value: String)
...
 /** A representation of an item in our ontology, without a canonical identifier */
-case class Work(identifiers: List[SourceIdentifier],
-                       label: String,
-                       accessStatus: Option[String] = None)
+case class Work(
+  identifiers: List[SourceIdentifier],
+  label: String,
+  accessStatus: Option[String] = None
+) {
+  @JsonProperty("type") val ld_type: String = "Work"
```
alicefuzier: I generally prefer camel case instead of underscores, unless it represents some external data format. What do people think?

alexwlchan:
Markups made.

alexwlchan:
Rebased against master, will merge once CI passes.

common/src/main/scala/uk/ac/wellcome/models/Work.scala
kenoir: Scala convention is camelCase right? We should follow the convention.

_even if underscores are more readable_

PR #247, opened 2017-05-05
Don't include null or empty fields in serialized JSON

### LINKED ISSUE #241: Omit optional fields that don't have an interesting value in the final API response
The schema in #129 has a number of optional fields, e.g. "description" and "lettering" may not be set for every work. Rather than returning a bunch of "null" values, we should just omit them from the final API response.

Implementation Q for @alicefuzier and @kenoir: do we drop the field from `Work` before we send it to Elasticsearch, or after we've got it from Elasticsearch but before we present it to the user? I lean towards the former (i.e. don't send null fields to ES), but could be persuaded the other way.

alexwlchan:

Ensure our JSON responses only include fields with interesting values. Resolves #241.

## Who is this change for?

Anybody who reads our JSON.

- [x] Tests – added
- [x] Docs – n/a


common/src/main/scala/uk/ac/wellcome/utils/JsonUtil.scala
```diff
@@ -11,6 +13,7 @@ object JsonUtil {
 
   mapper.registerModule(DefaultScalaModule)
   mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+  mapper.setSerializationInclusion(JsonInclude.Include.NON_EMPTY)
```
alicefuzier: 👍

common/src/test/scala/uk/ac/wellcome/test/ModelsTest.scala
```diff
@@ -0,0 +1,31 @@
+package uk.ac.wellcome.platform.common
+
+import org.scalatest.{FunSpec, Matchers}
+
+import uk.ac.wellcome.models.Work
+import uk.ac.wellcome.utils.JsonUtil
+
+
+class WorkTest extends FunSpec with Matchers {
```
alicefuzier: Maybe rename this to JsonUtilTest? Cause JsonUtil is the class actually tested

alexwlchan: I decided to break them into distinct test classes, because I expect to be adding tests that are specific to `Work` soon.

ScalaWilliam:
[That JSON debate](https://twitter.com/ChtefiD/status/860064576494333952)... 😁 
I'll be writing an article on this at some point.

PR #255, opened 2017-05-08
Improve logging of WorksIndex

alicefuzier:
Improve logging of WorksIndex so that it shows errors when it tries to update an index. Making tests connect to the local elasticsearch instead of the localnode. And lastly, improving test around index update to have a more significant starting status

Better logs and tests that better reflect reality
## Who is this change for?
Developers who investigate bugs

- [x] Tests?
- [x] Docs?

### DIFF
--- a/common/src/test/scala/uk/ac/wellcome/elasticsearch/mappings/WorksIndexTest.scala
+++ b/common/src/test/scala/uk/ac/wellcome/elasticsearch/mappings/WorksIndexTest.scala
@@ -29,7 +29,9 @@ class WorksIndexTest
   val worksIndex = new WorksIndex(elasticClient, indexName, itemType)
 
   override def beforeEach(): Unit = {
-    elasticClient.execute(deleteIndex(indexName)).await
+    if (elasticClient.execute(indexExists(indexName)).await.isExists){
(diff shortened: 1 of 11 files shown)


.gitignore
```diff
@@ -1,5 +1,6 @@
 .idea
 *.iml
+ElasticConnectTest.scala
```
kenoir: Why are we ignoring this?

common/src/test/scala/uk/ac/wellcome/elasticsearch/mappings/WorksIndexTest.scala
```diff
@@ -57,39 +71,62 @@ class WorksIndexTest
...
 
-    val eventualIndexResponse = client.execute(
+    println("index created")
+    val eventualIndexResponse = elasticClient.execute(
       indexInto(indexName / itemType)
         .doc("""{"json":"json not matching the index structure"}"""))
 
     whenReady(eventualIndexResponse.failed) { exception =>
-      exception shouldBe a[StrictDynamicMappingException]
+      exception shouldBe a[RemoteTransportException]
```
kenoir: De we need to look inside the exception here to check the underlying cause?

PR #262, opened 2017-05-09
Add single work endpoint

kenoir:

Return the expected result for a single `Work`

## Who is this change for?

Those wishing to consume single Work API results.

- [X] Tests?
- [X] Docs?

### DIFF
--- a/api/src/main/scala/uk/ac/wellcome/platform/api/models/Record.scala
+++ b/api/src/main/scala/uk/ac/wellcome/platform/api/models/Record.scala
@@ -3,13 +3,17 @@ package uk.ac.wellcome.platform.api.models
 import com.fasterxml.jackson.annotation.JsonProperty
 import com.sksamuel.elastic4s.searches.RichSearchHit
 import org.elasticsearch.action.get.GetResponse
-import uk.ac.wellcome.models.{IdentifiedWork, Work}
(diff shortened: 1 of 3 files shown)


kenoir:
Resolves https://github.com/wellcometrust/platform-api/issues/129

api/src/main/scala/uk/ac/wellcome/platform/api/models/Record.scala
```diff
@@ -3,13 +3,17 @@ package uk.ac.wellcome.platform.api.models
...
 
 case class Record(
   @JsonProperty("type") ontologyType: String = "Work",
   id: String,
-  label: String
+  label: String,
+  description: Option[String] = None,
+  lettering: Option[String] = None,
+  hasCreatedDate: Option[Period] = None,
+  hasCreator: List[Agent] = List()
```
alexwlchan: I didn't get a chance to review this yesterday, but is it a problem that this seems to track the `Work` type so closely?

kenoir: Not sure yet. I expect they will diverge (or rather where we keep the logic will). Experience tells me it's a good idea to separate the presentation model from the domain model.

PR #263, opened 2017-05-09
Add list endpoint

kenoir:

Adds a list of results endpoint providing a list of results!

## Who is this change for?

Those wishing to consume more than one item at a time from the API.

- [X] Tests?
- [X] Docs?

Depends https://github.com/wellcometrust/platform-api/pull/262


kenoir:
Resolves https://github.com/wellcometrust/platform-api/issues/128

api/src/main/scala/uk/ac/wellcome/platform/api/responses/ResultListResponse.scala
```diff
@@ -8,5 +8,7 @@ case class ResultListResponse(
   @JsonProperty("@context") context: String,
   @JsonProperty("type") ontologyType: String = "ResultList",
   pageSize: Int = 10,
+  totalPages: Int = 10,
```
kenoir: I think it makes more sense to code these in another PR. For now we can get results out.

PR #267, opened 2017-05-10
Add support for full-text search

### LINKED ISSUE #135: Initial support for full text search
This is an initial simple as possible implementation of full text search. No changes to weighting, relevance ranking, ordering etc.

This should be a `query` query parameter, the value of which is passed through to the Elasticsearch SimpleQueryParser, which will provide us a safe DSL for full text searching:

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html

This full text querying must be combinable with the structured querying provided by other query parameters.

alexwlchan:

Add a `query` parameter to the API for full text search.

Resolves #135.

## Who is this change for?

Users who want to see results in a way other than “the ordering of our arbitrarily assigned identifiers”.

UX team who want to do user research around search.

- [x] Tests – see patch
- [x] Docs – Swagger should automatically pick up the API changes
- [x] Spoken to the right people – based on @jtweed’s design in #135

### DIFF
--- a/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticSearchServiceTest.scala
+++ b/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticSearchServiceTest.scala
@@ -1,17 +1,18 @@
 package uk.ac.wellcome.platform.api.services
 
 import com.sksamuel.elastic4s.ElasticDsl._
-import org.junit.Test
-import org.scalatest.{AsyncFunSpec, Matchers}
+import org.scalatest.{FunSpec, Matchers}
+import org.scalatest.concurrent.ScalaFutures
 import uk.ac.wellcome.models.{IdentifiedWork, SourceIdentifier, Work}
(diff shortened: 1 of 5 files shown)


jtweed:
Is this using the `simple_query_string query`?

api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticSearchServiceTest.scala
```diff
@@ -38,14 +39,37 @@ class ElasticSearchServiceTest
...
   }
 
+  it("should only find results that match a query if doing a full-text search") {
+    val work1 = identifiedWorkWith(
+      canonicalId = "1234",
+      label = "A drawing of a dodo"
+    )
+    val work2 = identifiedWorkWith(
+      canonicalId = "5678",
+      label = "A mezzotint of a mouse"
```
kenoir: 🐁

alexwlchan:
> Is this using the `simple_query_string` query?

It is now. 😄

Issue #246, opened 2017-05-05
Tests in CI are flakey


alexwlchan:
On a fairly regular basis, I’m seeing Travis fail one of the container-based tests, but the failure clears up on a restart:

https://travis-ci.org/wellcometrust/platform-api/jobs/229136374
https://travis-ci.org/wellcometrust/platform-api/jobs/229136371

Tests should fail reliably.

kenoir:
Tests must be reliable. This is high priority I reckon.

alexwlchan:
Another flakey log in build 340: https://travis-ci.org/wellcometrust/platform-api/builds/229664810

[travis_api_log.txt](https://github.com/wellcometrust/platform-api/files/982804/travis_api_log.txt)

alexwlchan:
I’ve had to restart a job four times for my latest PR. 😞 

I wonder if we're getting squeezed by Travis when the builds are busy, and our tests expect a bit too much performance?

alicefuzier:
Hopefully it's fixed. We can reopen it if we find some more flakey tests

PR #271, opened 2017-05-10
Fix transformer flakyness

alicefuzier:
Make our builds more reliable, as currently some tests fail for no reason
## Who is this change for?
Developers who like reliable builds

- [x] Tests?
- [x] Docs?

### DIFF
--- a/common/src/test/scala/uk/ac/wellcome/test/utils/DynamoDBLocal.scala
+++ b/common/src/test/scala/uk/ac/wellcome/test/utils/DynamoDBLocal.scala
@@ -3,12 +3,7 @@ package uk.ac.wellcome.test.utils
 import com.amazonaws.auth.{AWSStaticCredentialsProvider, BasicAWSCredentials}
 import com.amazonaws.client.builder.AwsClientBuilder.EndpointConfiguration
 import com.amazonaws.services.dynamodbv2.model._
-import com.amazonaws.services.dynamodbv2.{
-  AmazonDynamoDB,
-  AmazonDynamoDBClientBuilder,
-  AmazonDynamoDBStreams,
-  AmazonDynamoDBStreamsClientBuilder
-}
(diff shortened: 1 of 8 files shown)


common/src/test/scala/uk/ac/wellcome/test/utils/DynamoDBLocal.scala
```diff
@@ -49,17 +47,40 @@ trait DynamoDBLocal extends BeforeAndAfterEach { this: Suite =>
...
-      List(identifiersTableName, miroDataTableName, calmDataTableName)
-      .foreach(tableName => clearTable(tableName))
+    clearIdentifiersTable()
+    clearMiroTable()
+    clearCalmTable()
   }
 
-  private def clearTable(tableName: String): List[DeleteItemResult] =
-    Scanamo.scan[Identifier](dynamoDbClient)(tableName).map {
+  private def clearIdentifiersTable(): List[DeleteItemResult] =
```
kenoir: Possible to DRY this up a bit?

transformer/src/main/scala/uk/ac/wellcome/transformer/modules/StreamRecordProcessorFactoryModule.scala
```diff
@@ -40,21 +39,11 @@ class StreamsRecordProcessor(receiver: RecordReceiver)
   }
 }
 
-class StreamsRecordProcessorFactory(
+class StreamsRecordProcessorFactory @Inject()(
```
kenoir: So if i'm understanding correctly it's appropriate to use the Module/Provider pattern when dependent on external config? https://twitter.github.io/finatra/user-guide/getting-started/modules.html

kenoir: > Generally, modules are only required for helping the injector instantiate classes that you don’t control. Otherwise, you would simply add the JSR-330 annotations (e.g., @Inject, @Singleton) directly to the class definition.

alicefuzier:
This should partially fix #246

PR #274, opened 2017-05-10
Add more detail to the docs about our ingest pipeline

alexwlchan:
I’ve been annoyed with the architecture diagrams for a while, for a couple of reasons:

* I mixed up the icon for a service (e.g. SQS) with the icon for a resource within that service (an SQS queue).
* I feel like the left-to-right only works if you have a wide screen, I’ve been wanting to try a vertical arrangement instead.
* A bit more detail, perhaps.

This is a stab at the dark at improving them. Not sure if I like this better or it just looks _different_, going to let it sit for a while.

Best viewed in rendered form: https://github.com/wellcometrust/platform-api/blob/new-pipeline-docs/docs/architecture.md


kenoir:
Looking good.

PR #275, opened 2017-05-11
Add support for pagination

alexwlchan:

Starting to lay the foundations for pagination support.

## Who is this change for?

Users who want to be able to see more than the first page of results.

- [ ] Tests?
- [ ] Docs?


api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticSearchService.scala
```diff
@@ -24,22 +24,34 @@ class ElasticSearchService @Inject()(@Flag("es.index") index: String,
         get(id).from(s"$index/$itemType")
       }
 
-  def findResults(sortByField: String, limit: Int = 10): Future[RichSearchResponse] =
+  private def executeSearch(searchDefinition: SearchDefinition, limit: Int, from: Int): Future[RichSearchResponse] =
```
alexwlchan: This is in anticipation of the fact that we’ll likely have more methods on this service, and it would be good not to keep repeating the Elasticsearch logic – but there’s still quite a lot of repetition. Is this a good idea?

In Python, we have a concept of _function decorators_ – you can "wrap" a function, replace it with the wrapped version, and that's where you'd add extra arguments. Looks something like:

```python
def decorator(f):
    def wrapper(*args, **kwargs):
        limit = kwargs.pop('limit')
        from = kwargs.pop('from')
        search_defn = f(*args, **kwargs)
        return execute_search(search_defn, limit=limit, from=from)
    return wrapper
```

and then the syntax is:

```python
@decorator
def find_results(sort_by_field):
    ...

@decorator
def simple_string_query_search(query_string):
    ...
```

and the individual functions just spit out the search definition.

Anything like that in Scala?

alexwlchan:
This is now working, in broad strokes. Good enough for a demo and a review, anyway.

Still to do:

* [ ] Write some tests for everything on the `WorksController` side
* [ ] Pull out some of the pagination arithmetic into a PaginationUtils module. We'll keep reusing it, and there are a few fiddly off-by-one errors I'd like to avoid.
* [ ] Address any review comments from @kenoir and @alicefuzier.

api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchServiceTest.scala
```diff
@@ -83,6 +83,38 @@ class ElasticsearchServiceTest
     whenReady(searchResultFutureWithLargeLimit) { _.hits should have size 5 }
   }
 
+  it("should be able to fetch from midway through the Elasticsearch results") {
```
alicefuzier: this is good but there should be more tests around this

alexwlchan:
I’m going to split off the pieces that change the ElasticsearchService, and put them in a separate PR. I think they’re more complete, and conceptually form a nice self-contained patch.

PR #277, opened 2017-05-11
Changes to ElasticsearchService for pagination

alexwlchan:

Make the changes to the ElasticsearchService for pagination. Breaks up #275.

## Who is this change for?

Developers working on #275.

- [x] Tests?
- [x] Docs?

### DIFF
--- a/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticSearchService.scala
+++ b/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticSearchService.scala
@@ -24,36 +24,28 @@ class ElasticSearchService @Inject()(@Flag("es.index") index: String,
         get(id).from(s"$index/$itemType")
       }
 
-  private def executeSearch(searchDefinition: SearchDefinition, limit: Int, from: Int): Future[RichSearchResponse] =
+  def listResults(sortByField: String,
+                  limit: Int = 10,
+                  from: Int = 0): Future[RichSearchResponse] =
     elasticClient
       .execute {
-        searchDefinition
+        search(s"$index/$itemType")
+          .matchAllQuery()
+          .sortBy(fieldSort(sortByField))
           .limit(limit)
           .from(from)
       }
 
-  def listResults(sortByField: String,
-                  limit: Int = 10,
-                  from: Int = 0): Future[RichSearchResponse] = {
-    val searchDefinition = search(s"$index/$itemType")
-      .matchAllQuery()
-      .sortBy(fieldSort(sortByField))
-    executeSearch(
-      searchDefinition = searchDefinition,
-      limit = limit,
-      from = from
-    )
-  }
-
-  def simpleStringQueryResults(queryString: String, limit: Int = 10, from: Int = 0): Future[RichSearchResponse] = {
-    val searchDefinition = search(s"$index/$itemType")
-      .query(simpleStringQuery(queryString))
-    executeSearch(
-      searchDefinition = searchDefinition,
-      limit = limit,
-      from = from
-    )
-  }
+  def simpleStringQueryResults(queryString: String,
+                               limit: Int = 10,
+                               from: Int = 0): Future[RichSearchResponse] =
+    elasticClient
+      .execute {
+        search(s"$index/$itemType")
+          .query(simpleStringQuery(queryString))
+          .limit(limit)
+          .from(from)
+      }
 
 }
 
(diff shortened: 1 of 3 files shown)


api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticSearchService.scala
```diff
@@ -24,21 +24,27 @@ class ElasticSearchService @Inject()(@Flag("es.index") index: String,
         get(id).from(s"$index/$itemType")
       }
 
-  def findResults(sortByField: String, limit: Int = 10): Future[RichSearchResponse] =
+  def listResults(sortByField: String,
```
alexwlchan: Rename is per a comment in #275.

api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchServiceTest.scala
```diff
@@ -47,40 +47,96 @@ class ElasticsearchServiceTest
...
+      expectedWorks = displayWorks.slice(5, 10)
+    )
+
+    // Take a slice that is beyond all our results
+    assertSliceIsCorrect(
+      searchService,
+      limit = 10,
+      from = 50,
+      expectedWorks = List()
+    )
```
alexwlchan: @alicefuzier Hopefully easier to add more tests now. Is this the sort of thing you had in mind?

alicefuzier: No really, I just meant that in general the pagination functionality only had one test but I wasn't referring necessarily to this bit (sorry actually I was a bit unclear, but that is why I think it's better to go though pull requests in person)

Anyway, I like more testing, especially corner cases, but each assertion should be on its own test with a test name rather then comments (or in addition to comments if they are not redundant at that point)

alexwlchan:
> Generally each test should test only one thing

Agreed, although we pay a cost in terms of reindexing everything to Elasticsearch, don't we? Still, change made.

I don't suppose ScalaTest has a way I can remove the repetition of `val displayWorks = populateElasticsearch()`, does it? We have fixtures in Python:

```
@pytest.fixture
def display_works(limit=10):
    return populate_elasticsearch(limit)

def test_foo(display_works):
    # the display_works variable is now available for use
```

api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchServiceTest.scala
```diff
@@ -46,41 +46,86 @@ class ElasticsearchServiceTest
...
+   *  `DisplayWork` instances.
+   */
+  private def populateElasticsearch(count: Int = 10): List[DisplayWork] = {
+    val works: List[IdentifiedWork] = (0 to 9).map { x =>
+      identifiedWorkWith(canonicalId = s"ID-$x", label=s"Work number $x")
+    }.toList
+    insertIntoElasticSearch(works: _*)
+
+    works.map { work: IdentifiedWork =>
+      DisplayWork("Work", work.canonicalId, work.work.label)
```
alicefuzier: Just one quick question: why do we map to ```DisplayWork```? These are the tests for ```ElasticSearchService``` which should be agnostic about the fact that they are ```Work```s. Is there a way to make the assertions without assuming  anything about their structure? Maybe just reading them as strings?

api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchServiceTest.scala
```diff
@@ -46,41 +46,86 @@ class ElasticsearchServiceTest
...
+    val searchResultFuture = searchService.listResults(
       sortByField = "canonicalId",
-      limit = 10
+      limit = limit,
+      from = from
     )
-    whenReady(searchResultFutureWithLargeLimit) { _.hits should have size 5 }
+    whenReady(searchResultFuture) { result =>
+      result.hits should have size expectedWorks.length
+      val returnedWorks = result.hits.map { DisplayWork(_) }
```
alicefuzier: same as above

api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchServiceTest.scala
alexwlchan: Convenience, mainly. That’s what the tests already had. I agree that these should be agnostic about the type of object they’re dealing with, but we have `Work` running throughout this setup.

I’d have no objection to pulling out the dependency on Work at a later date, but I don't fancy unpicking all those assumptions right now.

alicefuzier: Makes sense 👍

alicefuzier:
BTW forgot to reply to your question about repetition at the beginning of each test: you could use ```BeforeAndAfterEach``` trait and then override ```beforeEach`` to wrap setup code for each test. It's already in use in a lot of places so if you grep for it you should find examples

PR #278, opened 2017-05-11
Fix transformer flakyness

alicefuzier:
Again, fix some of our flakey tests. We seem to have a too strict timeout sometimes on the transformer finishing and also the kinesis worker keeps working long after the server has stopped 
## Who is this change for?
Developers who like reliable builds

- [x] Tests?
- [x] Docs?


alicefuzier:
This should (hopefully) resolve #246

PR #280, opened 2017-05-11
Uploading updated version of works.ttl

ChristopherHilton:
After discussion on how image "works" will relate to other works, works.ttl has been revised to add a note saying that when the work is an image, the fields relating to the date of creation describe the content imaged and not the actual image itself.

- [ ] Tests?
- [ ] Docs?

### DIFF
--- a/works.ttl
+++ b/works.ttl
@@ -1,0 +1,354 @@
+
+@prefix ww: <http://www.wellcomecollection.org/ontologies/works> . 
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> . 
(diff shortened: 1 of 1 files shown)


alexwlchan:
If you put this in the `ontologies` folder (overwriting the old copy of `works.ttl`), we'll be able to see just the differences between the two versions.

(I'd offer to help, but walking through an app I don't know very well over a crackly Internet connection might not do either of us any favours!)

ChristopherHilton:
Oops – of course, will do.  In there now.
C

alexwlchan:
Superseded by #290.

PR #290, opened 2017-05-12
Modification to works.ttl: adding clarification of relationship of works when one is an image

ChristopherHilton:
## Who is this change for?
This relates to the MIRO data, in which much of the descriptive metadata relates not to the digital image but to the thing of which it is an image: clarifying that CreatedDate will refer to the original and not the image itself.
Supersedes previous clumsy attempt to upload this file!

- [ ] Tests?
- [ ] Docs?


alexwlchan:
Hmm. Something funky is still going on with line endings to screw up the diff – let’s investigate on Monday.

PR #295, opened 2017-05-15
Adds Reindexer app project skeleton

kenoir:

Provides a starting place to begin working on the Reindexer

## Who is this change for?

Folks who want to reindex content into the digital platform.

- [X] Tests?
- [X] Docs?


kenoir:
Related: https://github.com/wellcometrust/platform-api/issues/194

reindexer/src/main/scala/uk/ac/wellcome/platform/reindexer/modules/ReindexModule.scala
```diff
@@ -0,0 +1,27 @@
...
+
+  override def singletonStartup(injector: Injector) {
+    info("Starting Reindexer module")
+
+    val actorSystem = injector.instance[ActorSystem]
+    run(() => start(), actorSystem)
+  }
+
+  private def start() = {
+
```
kenoir: TODO: Write app.

kenoir: ... seriously though I think we can merge this and then work on it.

PR #298, opened 2017-05-16
Start to separate the Travis logic from the build steps

### LINKED ISSUE #297: Run tests for all projects locally
In order to reduce development time we would benefit from a convenient way to run the same set of tests that occur in CI locally.

One approach would be abstract the CI (Travis) specific logic from the actual running of the tests.

alexwlchan:

Break out the build pieces from the Travis logic.

Resolves #297: it is now possible to run:

```console
$ PROJECT=api ./scripts/run_tests.sh
$ ./scripts/run_tests.sh
```

to run a subset or all of the tests.

## Who is this change for?

Developers who want it to be easier to run steps locally.

- [ ] Tests?
- [ ] Docs?

### DIFF
--- a/.travis/run.sh
+++ b/.travis/run.sh
@@ -2,7 +2,7 @@
 
 set -o errexit
 set -o nounset
-set -o verbose
+set -o xtrace
 
 # https://graysonkoonce.com/getting-the-current-branch-name-during-a-pull-request-in-travis-ci/
 if [[ "$TRAVIS_PULL_REQUEST" == "false" ]]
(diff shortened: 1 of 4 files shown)


.travis/run.sh
```diff
@@ -2,7 +2,7 @@
 
 set -o errexit
 set -o nounset
-set -o verbose
+set -o xtrace
```
kenoir: What's the difference between verbose and xtrace?

alexwlchan: Makes it slightly easier to pick out the debug content in `xtrace` (prefixed with a `+`) compared to `verbose`.

PR #299, opened 2017-05-16
Make pull request builds go faster

alexwlchan:

If we're on a pull request build, we can do a diff against master to check whether there are any file changes.  If there aren't any relevant changes, we can skip running the tests. A nice side-effect is that this will probably make non-PR builds go faster as well, because we have less contention for our share of Travis resources.

Doing this on master requires a little more care, so I haven't done that yet – this is just a proof-of-concept. Related to #203.

## Who is this change for?

Developers who want faster builds.

- [x] Tests?
- [x] Docs?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,23 @@ branches:
   only:
     - master
 
+before_install:
+  # If this is a pull request branch, we only do anything if there are
+  # interesting changes between this branch and master.
+  #
+  # Using 'exit' in a build step will terminate the rest of the build:
+  # https://docs.travis-ci.com/user/customizing-the-build/#How-does-this-work%3F-(Or%2C-why-you-should-not-use-exit-in-build-steps)
+  - |
+    if [[ "$TRAVIS_EVENT_TYPE" == "pull_request" ]]
+    then
+      git fetch origin
+      if ! python ./scripts/should_rerun_tests.py
+      then
+        echo "No changes that require us to re-run tests, skipping to the end"
+        exit 0
+      fi
+    fi
+
 install:
   - ./.travis/install.sh
 
--- a/scripts/should_rerun_tests.py
+++ b/scripts/should_rerun_tests.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python
+# -*- encoding: utf-8 -*-
+"""
+Decide whether we should re-run tests for a project on a pull request.
+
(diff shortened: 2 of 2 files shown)


alexwlchan:
This shaves off about three minutes of the total runtime, I’d like to see if skipping doing the install pieces will speed it up further – really this should be “check out branch, finish instantly”.

alexwlchan:
32:03 / 11:55 elapsed in [build #474](https://travis-ci.org/wellcometrust/platform-api/builds/232400378)

5:05 / 2:21 elapsed in [this PR’s build](https://travis-ci.org/wellcometrust/platform-api/builds/232899666)

scripts/should_rerun_tests.py
```diff
@@ -0,0 +1,63 @@
...
+    >>> modified_files('master', 'feature-branch')
+
+    """
+    files = set()
+    command = ['git', 'diff', '--name-only', a, b]
+    diff_output = subprocess.check_output(command).decode('ascii')
+    for line in diff_output.splitlines():
+        filepath = line.strip()
+        if filepath:
+            assert os.path.exists(filepath)
```
alexwlchan: Dear past!Alex: what if you've deleted a file?

Whoops.

.travis/run.sh
```diff
@@ -15,7 +15,6 @@ fi
 echo "TRAVIS_BRANCH=$TRAVIS_BRANCH, PR=$TRAVIS_PULL_REQUEST, BRANCH=$BRANCH"
 
 # Run the tests themselves.
-
```
alexwlchan: If you're tweaking, rebase out this change.

.travis/install.sh
```diff
@@ -2,7 +2,7 @@
 
 set -o errexit
 set -o nounset
-set -o verbose
+set -o xtrace
```
alexwlchan: Scrub the change from this patch.

.travis.yml
```diff
@@ -8,6 +8,24 @@ branches:
...
+    then
+      git fetch origin
+      if ! python ./scripts/should_rerun_tests.py
+      then
+        echo "No changes that require us to re-run tests, skipping to the end"
+        exit 0
+      fi
+    fi
+
+
```
alexwlchan: nit: newline

kenoir:
@alexwlchan This is some real nice bash you got here.

PR #320, opened 2017-05-25
Loading updated version of works ontology

ChristopherHilton:

Material relating to Identifiers has been updated: in this version, there are only two types.  These are:
 - identifier: covering unique system-generated identifiers, and identifiers within public metadata that are preferred for public citation purposes
- alternativeIdentifier: covering identifiers that are not preferred for citation purposes, such as previous identifiers that may occur still in outdated finding aids

- [ ] Reviewed by @jtweed


jtweed:
I'm completely confused by this pull request.

Because the file was removed, then fully replaced, rather than updated in place, it's really hard to see what's been changed.

And I'm not sure what's in the comment above is actually reflected in the ontology. Also one of the commits looked to have changed the dates, but they are still xsd:Date and not ww:Period.

ChristopherHilton:
I think I need to go back and do this one again - @kenoir and I were struggling with getting my local repository to synch this morning and it looks like this is the old version, not the new one.  I'll try to get the new version in tomorrow (and will make sure we have ww:Period).

PR #337, opened 2017-05-26
Add includes to swagger docs for works endpoint

alicefuzier:
What it says in the title
### Who is this change for?
Anyone who wants to know how to use our API

- [ ] Reviewed by @jtweed

- [ ] Deployed new versions


alexwlchan:
Question for @jtweed: how do people learn that these fields exist? Is there a special endpoint they query? Something in our docs?

jtweed:
Docs. We should be documenting the available includes. Probably as a separate section for each type, within ReadMe.

PR #341, opened 2017-05-26
Fix a silly syntax error in the Travis check scripts

alexwlchan:
Also change it so that an error in the scripts (= exit code 1)
means everything should be rebuilt, so a failure in the scripts
mean we just rebuild everything.

Get the master builds running again.

### Who is this change for?

Developers who like working builds.


alexwlchan:
(This is what happens when I merge w/o proper code review.)

PR #347, opened 2017-05-30
Serialisation changes for the API

alexwlchan:

Drop the "hasFoo" naming style in favour of "foos", and update the JSON-LD context.

### Who is this change for?

Users of our API.

- [x] Reviewed by @jtweed
- [ ] Deployed new versions

### DIFF
--- a/api/src/main/resources/context.json
+++ b/api/src/main/resources/context.json
@@ -5,7 +5,15 @@
       "@vocab": "http://wellcomecollection.org/ontologies/works/",
       "id": "@id",
       "type": "@type",
-      "hasCreator": {
+      "creators": {
+        "@type": "wa:agent",
+        "@container": "@list"
+      },
+      "createdDate": {
+        "@type": "wa:period"
+      },
+      "identifiers": {
+        "@type": "wa:identifier",
         "@container": "@list"
       }
     },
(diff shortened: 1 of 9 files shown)


api/src/main/resources/context.json
```diff
@@ -5,7 +5,15 @@
       "@vocab": "http://wellcomecollection.org/ontologies/works/",
       "id": "@id",
       "type": "@type",
-      "hasCreator": {
+      "creators": {
+        "@type": "wa:agent",
```
jtweed: This should be `"@id": "wa:hasAgent"`

api/src/main/resources/context.json
```diff
@@ -5,7 +5,15 @@
...
+        "@container": "@list"
+      },
+      "createdDate": {
+        "@type": "wa:period"
```
jtweed: This would be `"@type": "wa:Period"`, but I don't think we need it, as our style is to include the type in the response.

api/src/main/resources/context.json
```diff
@@ -5,7 +5,15 @@
...
+      },
+      "identifiers": {
+        "@type": "wa:identifier",
```
jtweed: This should be `"@id": "hasIdentifier"`

alexwlchan:
@kenoir Can I get a review from you for the code side please? :)

PR #348, opened 2017-05-30
Rename Reindex case class members to match table props

kenoir:

A functioning reindexer.

### Who is this change for?

Folk who want the reindexer to work

- [ ] Deployed new versions


kenoir:
@alexwlchan I cheated and merged this myself as there was no one around and it's a pretty small change to an already broken thing.

Issue #343, opened 2017-05-30
Prevent ID minter from minting duplicate IDs for Miro records


alexwlchan:
😱

alexwlchan:
Example: V0011478

PR #352, opened 2017-05-31
Add a script for spotting doubly-minted identifiers

alexwlchan:

Give us a way to see whether the ID minter has minted more than one canonical ID for a given source identifier. Helps us to diagnose #343.

### Who is this change for?

Folks who like the conventional meaning of the word “unique”, for example in that common household phrase _“there is a unique canonical ID for each source identifier”_.

### DIFF
--- a/scripts/find_duplicate_identifiers.py
+++ b/scripts/find_duplicate_identifiers.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python3
+# -*- encoding: utf-8 -*-
+"""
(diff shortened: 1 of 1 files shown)


alexwlchan:
You can string this together with shell pipelines to get a tally of how many duplicates we have:

```console
$ python scripts/find_duplicate_identifiers.py MiroID | awk 'BEGIN{FS=","} {print NF?NF:1}' | sort | uniq -c
2328 2
  87 3
```

PR #363, opened 2017-06-01
Improve the design of the TryBackoff

alexwlchan:

* Ensure that TryBackoff really does increase wait times between attempts
* Make it easier to change the outward behaviour (time to wait after first failure, total time to wait) without having to do any sums
* Add a `continuous` parameter so the TryBackoff can be stopped after the first success if necessary
* Log an error from the function wrapped by TryBackoff

### Who is this change for?

Devs who work with the TryBackoff code.

- [ ] Deployed new versions

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/utils/TryBackoff.scala
+++ b/common/src/main/scala/uk/ac/wellcome/utils/TryBackoff.scala
@@ -7,6 +7,7 @@ import uk.ac.wellcome.utils.GlobalExecutionContext.context
 import scala.concurrent.duration._
 import scala.math.pow
 import scala.util.{Failure, Success, Try}
+import scala.util.control.Breaks.break
 
 /** This trait implements an exponential backoff algorithm.  This is useful
   * for wrapping an operation that is known to be flakey/unreliable.
@@ -26,6 +27,13 @@ import scala.util.{Failure, Success, Try}
   *     @param totalWaitMillis how many milliseconds should we wait before
   *                            giving up on the operation
   *
+  * Additionally, the operation can run again after it succeeds (reverting
+  * to the base wait time), or give up after the first success.  This is
+  * controlled by a third attribute:
+  *
+  *     @param continuous      true if the operation should repeat, false
+  *                            if it should return after the first success
+  *
   * For example, to wait 1 second after the first failure and give up after
   * five minutes, we would set
   *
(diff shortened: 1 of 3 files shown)


kenoir:
Perhaps the reindexer need to implement the `continuous` param to indicate it should stop when done.

common/src/main/scala/uk/ac/wellcome/utils/TryBackoff.scala
```diff
@@ -7,6 +7,7 @@ import uk.ac.wellcome.utils.GlobalExecutionContext.context
 import scala.concurrent.duration._
 import scala.math.pow
 import scala.util.{Failure, Success, Try}
+import scala.util.control.Breaks.break
```
kenoir: :(

common/src/main/scala/uk/ac/wellcome/utils/TryBackoff.scala
```diff
@@ -19,30 +59,61 @@ trait TryBackoff extends Logging {
...
+      case Failure(e) =>
+        error(s"Failed to run (attempt: $attempt)", e)
         attempt + 1
     }
 
-    if (numberOfAttempts > maxAttempts)
+    // If we're in non-continuous mode and the last attempt succeeded,
+    // we should return immediately.
+    if (numberOfAttempts == 0 && !continuous) {
+      return
```
alicefuzier: Sorry, last comment: this is also not very functional...We should avoid returning in the middle of a function if possible. Try to use a value instead maybe? Something like:
```val shouldReSchedule = numberOfAttempts != 0 || continuous```
And then down below you use it to decide whether to schedule the function or not like so:
```
if(shouldReSchedule) {
      val cancellable = system.scheduler.scheduleOnce(waitTime milliseconds)(
        run(f, system, attempt = numberOfAttempts))
      maybeCancellable = Some(cancellable)
    }
```

File: common/src/main/scala/uk/ac/wellcome/utils/TryBackoff.scala [hunk 3]
```diff
@@ -4,45 +4,109 @@ import akka.actor.{ActorSystem, Cancellable}
...
+ *
+ *  Reference: https://en.wikipedia.org/wiki/Exponential_backoff
+ *
+ */
 trait TryBackoff extends Logging {
-  val baseWaitMillis = 100
-  val maxAttempts = 75
+  def baseWait = 100 millis
+  def totalWait = 12 seconds
+  def continuous = true
```
kenoir: Should be `val` rather than `def`?

alexwlchan:
I’ve rebased against current master and tidied the commit history.

alexwlchan:
Travis seems to be having some trouble with this patch. 😕

PR #364, opened 2017-06-02
Adds continuous param to TryBackoff so that we can stop after success

kenoir:

Prevents winning _too much_

### Who is this change for?

- [ ] Deployed new versions

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/utils/TryBackoff.scala
+++ b/common/src/main/scala/uk/ac/wellcome/utils/TryBackoff.scala
@@ -11,18 +11,19 @@ import scala.util.{Failure, Success, Try}
 trait TryBackoff extends Logging {
   val baseWaitMillis = 100
   val maxAttempts = 75
+  val continuous: Boolean = true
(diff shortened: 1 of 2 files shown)


alexwlchan:
Also 👍 for neat small PRs (this is functionality extracted from #363).

common/src/main/scala/uk/ac/wellcome/utils/TryBackoff.scala
```diff
@@ -32,9 +33,12 @@ trait TryBackoff extends Logging {
       if (attempt > 0) increaseWaitTimeExponentially(attempt)
       else baseWaitMillis
 
-    val cancellable = system.scheduler.scheduleOnce(waitTime milliseconds)(
-      run(f, system, numberOfAttempts))
-    maybeCancellable = Some(cancellable)
+    if (continuous && !succeeded) {
```
alexwlchan: This condition is still wrong, I think.

If it was continuous _or_ we failed, we should retry.

kenoir:
Closing as to wait for tests around TryBackoff

PR #365, opened 2017-06-02
Ensures reindexer doesn't retry when successful

kenoir:

Prevents winning _too much_

### Who is this change for?

- [ ] Deployed new versions


alexwlchan:
apologies, my git client borked

PR #367, opened 2017-06-02
Ensures reindexer doesn't retry when successful

alicefuzier:
Prevents winning too much
### Who is this change for?

- [ ] Deployed new versions

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/utils/TryBackoff.scala
+++ b/common/src/main/scala/uk/ac/wellcome/utils/TryBackoff.scala
@@ -54,7 +54,7 @@ trait TryBackoff extends Logging {
 
   def run(f: (() => Future[Unit]), system: ActorSystem, attempt: Int = 0): Unit = {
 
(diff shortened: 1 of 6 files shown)


common/src/main/scala/uk/ac/wellcome/utils/TryBackoff.scala
```diff
@@ -67,13 +67,13 @@ trait TryBackoff extends Logging {
         right => 0
       )
 
-      if (numberOfAttempts > maxAttempts) {
-        throw new RuntimeException("Max retry attempts exceeded")
-      }
-
-      val waitTime = timeToWaitOnAttempt(attempt)
+      val shouldReschedule = if (numberOfAttempts > maxAttempts) {
```
kenoir: Not for here, but worth thinking about: what would this look like if we had a "reschedule" function that returned an Option of Cancellable? Could we push some of this conditional logic into the `fold` above?

kenoir:

PR #372, opened 2017-06-06
Adds a query util for Scanamo to prevent OOM ...

kenoir:
... when performing transformations on large datasets.

Provide a better way to work on large queries from Scanamo to prevent OOM error.

### Who is this change for?

Developers

- [ ] Deployed new versions

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/utils/ScanamoUtils.scala
+++ b/common/src/main/scala/uk/ac/wellcome/utils/ScanamoUtils.scala
@@ -1,6 +1,16 @@
 package uk.ac.wellcome.utils
 
+import java.util
+
+import cats.free.Free
+import com.amazonaws.services.dynamodbv2.model.{AttributeValue, QueryResult}
+import com.gu.scanamo.{DynamoFormat, ScanamoFree}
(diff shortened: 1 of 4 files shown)


common/src/main/scala/uk/ac/wellcome/utils/ScanamoUtils.scala
```diff
@@ -17,3 +26,42 @@ object ScanamoUtils {
       .flatMap(_.right.toOption)
   }
 }
+
+object ScanamoQueryStream {
```
alexwlchan: Since this class is copied together from two classes in Scanamo (it looks like [DynamoResultStream.scala](https://github.com/guardian/scanamo/blob/f358cbfc6ff74c435600ddd2f6e9812aaf81b8e6/src/main/scala/com/gu/scanamo/DynamoResultStream.scala), but I haven't checked in detail), please could we get a comment that:

1. Links back to the original source of this code
2. Highlights anywhere we've deviated from the original

kenoir: 👍

kenoir:

PR #381, opened 2017-06-08
Fine-tuning works ontology

ChristopherHilton:
Adjustments to Works ontology as a result of considering how the Items ontology will look: Shelfmark has been removed from the Works ontology and will appear in Items instead, as hasShelfmark will be a local, Item property rather than one applying to an entire Work.  Also a minor verbal adjustment to the explanation here of what an Item is and how it differs from a Work.

Minor adjustments to ensure that the Works ontology is in line with what will shortly follow in the ontology for Items.

### Who is this change for?
All team members

### DIFF
--- a/ontologies/works.ttl
+++ b/ontologies/works.ttl
@@ -22,7 +22,7 @@ ww:Work rdf:type owl:Class ;
 	
 ww:Item rdf:type owl:Class ;
 	rdfs:label "Item"@en ;
-	rdfs:comment "A specific instance of a work: for instance, an individual physical copy with its own location.  This corresponds to the bottom, Instance, layer of FRBR.  Works that are unique - that is, have only one copy - are described in terms of a Work for their intellectual content, linked to an Item describing their physical characteristics."@en ;
+	rdfs:comment "A specific instance of a work: for instance, an individual physical copy with its own location.  This corresponds to the bottom, Instance, layer of FRBR.  Works that are unique - that is, have only one copy - are described in terms of a Work for their intellectual content, linked to an Item describing their local (usually physical) characteristics."@en ;
 	rdf:isDefinedBy <http://www.wellcomecollection.org/ontologies/works> .	
 	
 ww:Identifier rdf:type owl:Class ;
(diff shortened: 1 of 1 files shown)


pollecuttn:
@ChristopherHilton I'm thinking again about that Dewey mention and think it might be confusing in that it might cause people to look for Dewey classifications that don't exist in our data. The way to find out is by you user test the wording. I'm sure people in the Experience will help you with how to do that.

Also in the ontology re hasShelfmark: is that for just the classification, i.e. YHX.AK, or the call number/shelfmark i.e. Quick Ref. Collection YHX.AK ? Are you planning on having both so you can find all things co-located on a physical shelf, and all things sharing a classmark?

ChristopherHilton:
Fair enough re Dewey - the idea was simply to make reference to a classification that existed widely outside our own institution so that it would be comprehensible to outsiders, but if I replace that with NLM then it kills two birds with one stone, referring to something known outside that's also represented here.

Re Shelfmark, yes, that would just be for the classification YHX.AK - whether closed stack or open shelves - but there will also be a Location property, modelled on that in the Sierra item records, which says it's in e.g. History of Medicine collection.  So we can bring all things with a particular classification together regardless of physical location, or all things sharing a physical location, or both.

pollecuttn:
Re NLM: yep, sounds good.

I think in that case, hasShelfmark would be better called hasClassmark or hasClassification or something like that which removes the shelfmark/call number naming.

That also has the bonus of then having shelfmark available for location + classification.

ChristopherHilton:
That makes sense, yes!

pollecuttn:
Okay, rethinking this again: what are you expecting the currently-named hasShelfmark give for b1564172 (http://catalogue.wellcomelibrary.org/record=b1564172)?

The item record locations and shelfmarks are:
History of Medicine		BA /POR
Student Collection		/POR
Student Collection		/POR
Student Collection		/POR
Student Collection		/POR
Student Collection		/POR
Student Collection		/POR
Student Collection		/POR
Closed stores Professional Collection	/POR

The bib record has a local classification field 698 BA, i.e. only one item has a classification, the others just have a shelfmark from the author's surname.

I expect there are lots of other examples like this in our data.

ChristopherHilton:
That sort of scenario - multiple copies with different shelfmarks - was what decided me that this was an Item property rather than a Works one; so in these instances, what I'd envisage was 
Location: History of Medicine,  Classmark: BA /POR
Location: Student Collection, Classmark:  /POR
Location: Closed stores Professional Collection, Classmark:  /POR
etc.
This means of course that some classmarks are not very informative, just the first three letters of the author's name: but that's how it is, after all, and that's why the description of Classmark says "often _but not always_ relating to a subject-based scheme of classification and arrangement such as Barnard..."

pollecuttn:
Yes, I'm not sure we could do anything else. I wonder what it might mean if we wanted to do anything clever with classifications, e.g. "show me all things classified x". Authors initials aren't a classification (I guess that's why it's labelled shelfmark!) so would add noise rather than signal.

Might be worth talking it over with Silver when he's in, and the devs, so that everyone's aware of what is going to turn up there.

ChristopherHilton:
Yes, something to discuss.  Silver's now booked in to visit on Monday 26th June am, btw.
We can, I suppose, identify those "useless" bits of the shelfmark field by the fact they have a slash before them, so maybe could instruct the stuff coming out of Sierra to ditch everything after a /; though we'd have to check that this wouldn't lead to trouble with any other types of material (at least CALM references with all their slashes won't be coming down the Sierra pipeline).

PR #383, opened 2017-06-08
Adds Metrics to Ingestor

kenoir:

Better visibility of Ingestor activity

### Who is this change for?

Devs

- [ ] Deployed new versions

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/finatra/modules/AmazonCloudWatchModule.scala
+++ b/common/src/main/scala/uk/ac/wellcome/finatra/modules/AmazonCloudWatchModule.scala
@@ -10,11 +10,14 @@ import com.twitter.inject.TwitterModule
 import uk.ac.wellcome.models.aws.AWSConfig
 
 object AmazonCloudWatchModule extends TwitterModule {
-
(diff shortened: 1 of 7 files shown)


kenoir:
Related: https://github.com/wellcometrust/platform-api/issues/325

common/src/main/scala/uk/ac/wellcome/finatra/modules/AmazonCloudWatchModule.scala
```diff
@@ -10,11 +10,14 @@ import com.twitter.inject.TwitterModule
 import uk.ac.wellcome.models.aws.AWSConfig
 
 object AmazonCloudWatchModule extends TwitterModule {
-
-  val awsEndpoint = flag[String](
+  private val awsNamespace = flag[String]("aws.metrics.namespace",
+                                          "",
+                                          "Namespace for cloudwatch metrics")
```
alexwlchan: Is there a corresponding infra patch to add this flag?

kenoir: Good point. 👍

PR #384, opened 2017-06-08
Reduce memory pressure in reindexer by using smaller objects

kenoir:

Reduce memory pressure in reindexer by using smaller objects

### Who is this change for?

Thrifty developers

- [ ] Deployed new versions


kenoir:

PR #388, opened 2017-06-08
Set reindex version to the value set in ReindexTracker table

alicefuzier:
avoid multiple runs if there are records that have an older reindex version (possible if we stop the reindexer halfway through)
### Who is this change for?
Devs

- [ ] Deployed new versions

### DIFF
--- a/reindexer/src/main/scala/uk/ac/wellcome/platform/reindexer/services/ReindexTargetService.scala
+++ b/reindexer/src/main/scala/uk/ac/wellcome/platform/reindexer/services/ReindexTargetService.scala
@@ -41,7 +41,8 @@ abstract class ReindexTargetService[T <: Reindexable[String]](
 
   protected val scanamoQueryStreamFunction: ScanamoQueryStreamFunction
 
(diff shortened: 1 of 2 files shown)


reindexer/src/main/scala/uk/ac/wellcome/platform/reindexer/services/ReindexTargetService.scala
```diff
@@ -41,22 +41,24 @@ abstract class ReindexTargetService[T <: Reindexable[String]](
 
   protected val scanamoQueryStreamFunction: ScanamoQueryStreamFunction
 
-  private def updateVersion(resultGroup: List[ScanamoQueryResult]): Boolean = {
+  private def updateVersion(newReindexVersion: Int,
```
kenoir: I would use multiple parameters here `updateVersion(requestedVersion: Int)(resultGroup: List[ScanamoQueryResult])`, so it can be called later `scanamoQueryStreamFunction(
      scanamoQueryRequest,
      updateVersion(requestedVersion))`

PR #394, opened 2017-06-09
Setting up alarms when DLQs are not empty

alicefuzier:
Be notified when there is stuff in the dlqs
### Who is this change for?
Devs

- [x] Run `terraform apply`.

### DIFF
--- a/terraform/sqs/sqs.tf
+++ b/terraform/sqs/sqs.tf
@@ -7,18 +7,3 @@ resource "aws_sqs_queue" "q" {
 resource "aws_sqs_queue" "dlq" {
   name = "${var.queue_name}_dlq"
 }
-
-resource "aws_cloudwatch_metric_alarm" "dlq_not_empty" {
-  alarm_name          = "${aws_sqs_queue.dlq.name}_not_empty"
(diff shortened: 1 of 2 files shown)


alexwlchan:
Putting it in the module – nice idea. 👍 

Please could you run `terraform fmt` over the patch please?

terraform/sqs/sqs.tf
```diff
@@ -7,3 +7,18 @@ resource "aws_sqs_queue" "q" {
...
 }
+
+resource "aws_cloudwatch_metric_alarm" "dlq_not_empty" {
+  alarm_name = "${aws_sqs_queue.dlq.name}_not_empty"
+  comparison_operator = "GreaterThanOrEqualToThreshold"
+  evaluation_periods = 1
+  metric_name = "NumberOfMessagesSent"
+  namespace = "SQS"
+  period = 60
+  threshold = 1
```
alexwlchan: Maybe `GreaterThanThreshold` and `threshold = 0`?

I’m not sure it makes a meaningful difference, that’s just what I might have reached for first.

kenoir:
Looks good.

Issue #325, opened 2017-05-25
Monitoring/Alerting/Logging Epic


alexwlchan:
Logging:
- [x] ID Minter:  Log failures
- [x] Ingestor:  Log failures
- [x] API: Log requests

Monitoring:
- [x] Visibility of service status: 
  - https://github.com/wellcometrust/platform-api/pull/397
  - ^ Add UI to JSON view
- [x] Visibility of dynamo capacity
- [x] Report counts in reindexer and transformer accurately.
- [x] Ingestor; Add metrics for items ingested: https://github.com/wellcometrust/platform-api/pull/383
- [x] ID Minter; Add metrics for IDs minted: https://github.com/wellcometrust/platform-api/pull/389
- [x] API: Request time/status metrics.
- [x] Notification of failed deploys: See  https://github.com/wellcometrust/platform-api/issues/397
- [x] Monitoring cluster hosts status (disk/memory/cpu): https://github.com/wellcometrust/platform-infra/pull/107
- [x] Monitoring container status (disk/memory/cpu): https://github.com/wellcometrust/platform-infra/pull/107

Alerts:
- [x] ID Minter DLQ > 0
- [x] ES Ingest DLQ > 0 
- [ ] ID Minter Queue Size > n
- [ ] ES Ingest Queue Size > n
- [ ] MiroData: Dynamo R/W exceeded
- [ ] Identifiers: Dynamo R/W exceeded
- [ ] ReindexTracker: Dynamo R/W exceeded
- [ ] Dynamo Stream Tables!
- [ ] Transformer failed to process a record

Alerts Visibility:
- [ ] Slackbot
- [x] Dashboards reporting counts in pipeline / API request status / frequency https://github.com/wellcometrust/platform-api/pull/403 https://github.com/wellcometrust/platform-api/pull/422

kenoir:
Related: https://github.com/wellcometrust/platform-api/pull/303

kenoir:
Closing!

pollecuttn:
@kenoir There are unchecked checkboxes on this still for alerts - do we not need to worry about those? Why are we not worrying about them? Ta!

kenoir:
Not for now. We're done as far as proactive alerting is concerned. I think there's more value in adding further alerts as we see problems.

PR #403, opened 2017-06-13
Deploy Grafana

alicefuzier:
deploy grafana on our AWS cluster so that we can use it to visualize our metrics. Related to #325 
### Who is this change for?
Devs and everyone who is interested in the state of our platform

- [ ] Run `terraform apply`.


terraform/alb.tf
```diff
@@ -7,6 +7,15 @@ module "services_alb" {
   vpc_id                       = "${module.vpc_services.vpc_id}"
 }
 
+module "monitoring_alb" {
+  source                       = "./ecs_alb"
+  name                         = "services"
```
alexwlchan: Should the name be different?

alicefuzier: oops

terraform/grafana.tf
```diff
@@ -0,0 +1,116 @@
...
+}
+
+resource "aws_ecs_task_definition" "task" {
+  family                = "grafana_task_definition"
+  container_definitions = "${data.template_file.definition.rendered}"
+  task_role_arn         = "${module.ecs_grafana_iam.task_role_arn}"
+
+  volume {
+    name = "grafana"
+    host_path = "/ecs/grafana"
```
alicefuzier: As far as I understand this is not persistent, i.e. if the EC2 instance goes down this is lost. Also this is not synchronized across instances of the same ECS cluster, so, if we have more than one EC2 instance, the container might not have the most up to date files depending on which instance it is started on. Is there a way to fix this? EFS maybe?

kenoir: EFS (multiple instance)/EBS (single instance)  is the only solution (or FUSE to S3 which I _think_ is not recommended), unless Grafana itself supports a non local FS datastore.

kenoir: If Grafana cannot function without a local filesystem then we should talk further about the options.

alicefuzier: I think that by default it uses a sqlite database - so file based - but it can be redirected at any sql database so we can change the config to use our Aurora. Definitely easier, but I'm not sure how expensive it is? It should use very little data though

kenoir: Sounds reasonable.  Not sure how sensible it is to add contention to the Aurora DB for the ID minter.

alicefuzier: Yeah, I don't think it should use the id_minter database. Creating another one just for this seems unreasonably expensive though... Would it be so bad to mount an EFS volume? It seems the cheapest option although not the easiest. 
Alternatively we could spin up a t2.nano or t2.small ec2 instance with mysql for this. It's probably cheaper than aurora?

terraform/asg.tf
```diff
@@ -6,11 +6,29 @@ module "services_cluster_asg" {
...
+  asg_name              = "monitoring-cluster"
+  subnet_list           = ["${module.vpc_monitoring.subnets}"]
+  key_name              = "${var.key_name}"
+  instance_profile_name = "${module.ecs_monitoring_iam.instance_profile_name}"
+  user_data             = "${module.monitoring_userdata.rendered}"
+  vpc_id                = "${module.vpc_monitoring.vpc_id}"
+  instance_type         = "t2.medium"
+  admin_cidr_ingress = "${var.admin_cidr_ingress}"
+  # Grafana containers persist information about configured dashboards on a volume exposed by the EC2
+  # instance they are running on. Mounted volumes are not syncrnized across different EC2 instances of an
```
alexwlchan: /s/syncrnized/synchronized

terraform/asg.tf
```diff
@@ -6,11 +6,29 @@ module "services_cluster_asg" {
...
+module "monitoring_cluster_asg" {
+  source                = "./ecs_asg"
+  asg_name              = "monitoring-cluster"
+  subnet_list           = ["${module.vpc_monitoring.subnets}"]
+  key_name              = "${var.key_name}"
+  instance_profile_name = "${module.ecs_monitoring_iam.instance_profile_name}"
+  user_data             = "${module.monitoring_userdata.rendered}"
+  vpc_id                = "${module.vpc_monitoring.vpc_id}"
+  instance_type         = "t2.medium"
+  admin_cidr_ingress = "${var.admin_cidr_ingress}"
```
alexwlchan: this needs a `terraform fmt`

terraform/grafana.tf
```diff
@@ -0,0 +1,120 @@
...
+data "template_file" "definition" {
+  template = "${file("grafana.json.template")}"
+
+  vars {
+    log_group_region = "${var.aws_region}"
+    log_group_name   = "${aws_cloudwatch_log_group.task.name}"
+    anonymous_enabled = "${var.grafana_anonymous_enabled}"
+    anonymous_role = "${var.grafana_anonymous_role}"
+    admin_user = "${var.grafana_admin_user}"
+    admin_password = "${var.grafana_admin_password}"
```
alexwlchan: `terraform fmt`

terraform/grafana.tf
```diff
@@ -0,0 +1,120 @@
...
+      "elasticloadbalancing:Describe*",
+      "elasticloadbalancing:RegisterInstancesWithLoadBalancer",
+      "elasticloadbalancing:RegisterTargets",
+    ]
+
+    resources = [
+      "*",
+    ]
+  }
+}
```
alexwlchan: Blergh this repetition. I’d like to see us pull this back into the `services` module at some point, and just expose a couple of extra flags that Grafana needs.

terraform/asg.tf
```diff
@@ -6,11 +6,30 @@ module "services_cluster_asg" {
...
+  instance_profile_name = "${module.ecs_monitoring_iam.instance_profile_name}"
+  user_data             = "${module.monitoring_userdata.rendered}"
+  vpc_id                = "${module.vpc_monitoring.vpc_id}"
+  instance_type         = "t2.medium"
+  admin_cidr_ingress    = "${var.admin_cidr_ingress}"
+
+  # Grafana containers persist information about configured dashboards on a volume exposed by the EC2
+  # instance they are running on. Mounted volumes are not synchronized across different EC2 instances of an
+  # ECS cluster. If we have more than one EC2 instance we cannot guarantee that the files will be where
+  # they need to be or that they will be up to date
```
alexwlchan: tiny nit: missing full stop

PR #408, opened 2017-06-14
Update information on MIRO notes field

ChristopherHilton:
Adding information about the <image_supp_notes> field: it is a mixture of public and staff-only content and could therefore not be made public without clean-up.
### Who is this change for?
All Platform team (and potentially, long-term, the Delivery team who may be involved in cleaning up the data)


ChristopherHilton:
Thanks – I’ll keep ‘em coming

Issue #407, opened 2017-06-14
How do we notice deleted CALM records?


alexwlchan:
Prompted by a thread on the CALM-users mailing list: our OAI will silently drop records from the OAI-PMH when they’re deleted in CALM. We should probably notice this and do something sensible.

alexwlchan:
Closing this until we’re actually working with Calm data!

PR #412, opened 2017-06-15
Turn our ?includes query parameter into a typed structure

alexwlchan:
This turns it into a series of boolean switches which can be parsed in the WorksController, simplifying the logic in downstream classes. Additionally, this gives us a good point to check whether the includes parameter contains valid values.

A first step towards #407.

Tidy up some of our internal logic around ?includes.

### Who is this change for?

Developers who like tidy code. No externally visible changes.

- [ ] Deployed new versions

### DIFF
--- a/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
+++ b/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
@@ -8,6 +8,7 @@ import com.twitter.finatra.request.{QueryParam, RouteParam}
 import com.twitter.finatra.validation._
 import com.twitter.inject.annotations.Flag
 import uk.ac.wellcome.platform.api.ApiSwagger
+import uk.ac.wellcome.platform.api.models.WorksIncludes
 import uk.ac.wellcome.platform.api.responses.{ResultListResponse, ResultResponse}
 import uk.ac.wellcome.platform.api.services.WorksService
 import uk.ac.wellcome.utils.GlobalExecutionContext.context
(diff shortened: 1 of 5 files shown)


alexwlchan:
A side benefit I forgot to put in the commit message: all our includes are now defined in one place. Previously the list of valid includes could be scattered around the codebase. 😢

api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
```diff
@@ -130,4 +132,14 @@ class WorksController @Inject()(@Flag("api.prefix") apiPrefix: String,
     }
 
   }
+
+  /// Parse a comma-separated ?includes query parameter.
+  ///
+  /// TODO: Throw an exception if we get an unrecognised include.
```
alexwlchan: I’m not planning to address this TODO in this patch – keep the refactor and feature addition separate!

api/src/main/scala/uk/ac/wellcome/platform/api/models/DisplayWork.scala
```diff
@@ -6,6 +6,8 @@ import org.elasticsearch.action.get.GetResponse
 import uk.ac.wellcome.models._
 import uk.ac.wellcome.utils.JsonUtil
 
+case class WorksIncludes(identifiers: Boolean = false)
```
alexwlchan: Two things I’m not sure about: is this the right place to define this class? And is this the right name?

kenoir: I think yes to both, if the class grows a companion object then evict it to another file.

alexwlchan:
Sounds good, I’ll resolve merge conflicts and merge in the morning.

PR #416, opened 2017-06-19
Throw a 400 Bad Request error if the user asks for undefined includes

### LINKED ISSUE #411: Throw an error on undefined ?includes parameters
e.g. if somebody asks for

```
https://api.wellcomecollection.org/catalogue/v0/works?includes=hasIdentifiers
```

we should spit out an error response, maybe including a list of valid identifiers. Related to #340.

alexwlchan:

If a user tries to pass unrecognised `?includes=` parameters to the API, return a 400 Bad Request error and tell them which errors we didn’t like. The error response will look similar to the response returned on invalid pages:

```json
{"errors":["includes: [\"foo\", \"bar\"] are not recognised includes parameters"]}
```

Resolves #411.

### Who is this change for?

Users of the API.

- [ ] Reviewed by @jtweed
- [ ] Deployed new versions

### DIFF
--- a/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
+++ b/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
@@ -41,12 +41,13 @@ object IncludesValidation {
 
   /// Check if the provided ?includes parameters contain fields we recognise.
   def validateIncludes(queryParam: Option[String]): ValidationResult =
-    WorksIncludes.validate(queryParam) match {
-      case Left(_) => Valid
-      case Right(badIncludes) =>
-        badIncludes.length match {
-          case 1 => Invalid(s"includes: '${badIncludes.head}' is not a valid include")
-          case _ => Invalid(s"includes: ${badIncludes.mkString("'", "', '", "'")} are not valid includes")
+    WorksIncludes.create(queryParam) match {
+      case Right(_) => Valid
+      case Left(badIncludes) =>
+        if (badIncludes.length == 1) {
+          Invalid(s"includes: '${badIncludes.head}' is not a valid include")
+        } else {
+          Invalid(s"includes: ${badIncludes.mkString("'", "', '", "'")} are not valid includes")
         }
     }
 }
--- a/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
+++ b/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
@@ -21,36 +21,13 @@ case class MultipleResultsRequest(
   @RouteParam id: Option[String],
   @QueryParam query: Option[String],
   @QueryParam _index: Option[String]
-) {
-
-  // @MethodValidation
-  // def validateIncludes = IncludesValidation.validateIncludes(includes)
-}
+)
 
 case class SingleWorkRequest(
   @RouteParam id: String,
   @QueryParam includes: Option[WorksIncludes],
   @QueryParam _index: Option[String]
-) {
-
-  // @MethodValidation
-  // def validateIncludes = IncludesValidation.validateIncludes(includes)
-}
-
-object IncludesValidation {
-
-  /// Check if the provided ?includes parameters contain fields we recognise.
-  def validateIncludes(queryParam: Option[String]): ValidationResult =
-    WorksIncludes.create(queryParam) match {
-      case Right(_) => Valid
-      case Left(badIncludes) =>
-        if (badIncludes.length == 1) {
-          Invalid(s"includes: '${badIncludes.head}' is not a valid include")
-        } else {
-          Invalid(s"includes: ${badIncludes.mkString("'", "', '", "'")} are not valid includes")
-        }
-    }
-}
+)
 
 @Singleton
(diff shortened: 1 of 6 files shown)


api/src/test/scala/uk/ac/wellcome/platform/api/ApiWorksTest.scala
```diff
@@ -543,4 +558,67 @@ class ApiWorksTest
...
+    eventually {
+      server.httpGet(
+        path = s"/$apiPrefix/works?includes=foo,identifiers,bar",
+        andExpect = Status.BadRequest,
+        withJsonBody = """{"errors" : ["includes: 'foo', 'bar' are not valid includes"]}"""
+      )
+    }
+  }
+
+  it("should return multiple errors if appropriate") {
```
alexwlchan: This test is the only one still failing, for reasons I’m still digging into. 😕

api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
```diff
@@ -44,36 +43,19 @@ object IncludesValidation {
...
       case Some(includesString) => {
-        val includesList = includesString.split(",").toList
-        val recognisedIncludes = getCaseClassFieldNames[WorksIncludes]
-        val badIncludes = includesList
-          .filterNot(recognisedIncludes.contains)
-        badIncludes.length match {
-          case 0 => Valid
-          case 1 => Invalid(s"includes: '${badIncludes.head}' is not a valid include")
-          case _ => Invalid(s"includes: ${badIncludes.mkString("'", "', '", "'")} are not valid includes")
+        WorksIncludes.apply(includesString) match {
```
kenoir: This is a slight misuse of `apply` - generally the apply method is intended to be used transparently, e.g.

See: https://twitter.github.io/scala_school/basics2.html#apply

For example:

```scala
case class Foo(bar: String, bat: String)

// This is a companion object (see: http://daily-scala.blogspot.co.uk/2009/09/companion-object.html)
object Foo {
  def apply(baz: String): Foo =  {
     Foo(baz, baz) 
  }
}

File: api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala [hunk 3]
```diff
@@ -44,36 +43,19 @@ object IncludesValidation {
...
+          case Left => Valid
```
kenoir: As a consequence of the above, you might end up wrapping the creation in a `Try` - alternatively if you see fit create a different (non-apply) factory method that returns an `Either`.

api/src/main/scala/uk/ac/wellcome/platform/api/models/WorksIncludes.scala
```diff
@@ -0,0 +1,47 @@
...
+  /// is a known include.  Returns a Left if this is true, or a Right
+  /// containing a list of any unrecognised includes.
+  def validate(queryParam: String): Either[Unit, List[String]] = {
+    val includesList = queryParam.split(",").toList
+    val unrecognisedIncludes = includesList
+      .filterNot(recognisedIncludes.contains)
+    unrecognisedIncludes.length match {
+      case 0 => Left(Unit)
+      case _ => Right(unrecognisedIncludes)
+    }
```
alexwlchan: Use an `if … else` here.

api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
```diff
@@ -20,13 +21,35 @@ case class MultipleResultsRequest(
...
+
+  /// Check if the provided ?includes parameters contain fields we recognise.
+  def validateIncludes(queryParam: Option[String]): ValidationResult =
+    WorksIncludes.validate(queryParam) match {
+      case Left(_) => Valid
+      case Right(badIncludes) =>
+        badIncludes.length match {
+          case 1 => Invalid(s"includes: '${badIncludes.head}' is not a valid include")
+          case _ => Invalid(s"includes: ${badIncludes.mkString("'", "', '", "'")} are not valid includes")
+        }
```
alexwlchan: if … else

api/src/main/scala/uk/ac/wellcome/platform/api/models/WorksIncludes.scala
alexwlchan: Also, wrong way round – left is bad, right is good

api/src/main/scala/uk/ac/wellcome/platform/api/models/WorksIncludes.scala
```diff
@@ -0,0 +1,47 @@
...
+      case None => WorksIncludes()
+    }
+  }
+
+  /// Validate an ?includes query-parameter string.
+  ///
+  /// Specifically, this checks that every element in the query parameter
+  /// is a known include.  Returns a Left if this is true, or a Right
+  /// containing a list of any unrecognised includes.
+  def validate(queryParam: String): Either[Unit, List[String]] = {
```
alexwlchan: /s/validate/create, return the actual object – then use _this_ instead of apply

kenoir:
@alexwlchan some tests failing in CI, otherwise good to go.

PR #418, opened 2017-06-19
Adding "edition" property to Works

ChristopherHilton:
Correcting oversight - Works previously had publisher and publication date etc but no way of indicating which of variant editions this data might describe.
### Who is this change for?
Platform team


ChristopherHilton:
Thanks!

PR #421, opened 2017-06-19
Adding general note

ChristopherHilton:
Both CALM and Sierra have fields for general miscellaneous notes: this adds this concept to the ontology (woolly though it is)
### Who is this change for?
Platform team


kenoir:
I'm a bit wary of "general notes" as they are essentially a sink for metadata that could be more narrowly classified. I'm also aware that this might be a necessity to prevent other fields being misused where there isn't an appropriate field. 

TLDR; I don't like it, but I accept this as a necessity.

ChristopherHilton:
Absolutely: they’re a recipe for trouble, aren’t they, and one would never set them up on purpose…. Adding this with gritted teeth purely to accommodate the existing data, which alas is a bit too various to put consistently into other existing fields.  But yes, one holds one’s nose whilst adding this.

PR #423, opened 2017-06-19
JSON example of Sierra book

ChristopherHilton:
A third pass at a JSON file that maps ontology to real data: this is a book described in Sierra
### Who is this change for?
Platform team

### DIFF
--- a/Sierra_book.json
+++ b/Sierra_book.json
@@ -10,8 +10,6 @@
       }
     ],
   "label": "The Arctic grail: the quest for the North West Passage and the North Pole, 1818-1909",
-  "description": "A lecture on pneumatics at the Royal Institution, London. Coloured etching by J. Gillray, 1802.",
-  "lettering": "Scientific researches! -New discoveries in pneumaticks! -or- an experimental lecture on the powers of air.- Js. Gillray inv. & fect.",
   "creators": 
     [
       {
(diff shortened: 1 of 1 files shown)


Sierra_book.json
```diff
@@ -0,0 +1,96 @@
...
+  "isbn": "0670824917"
+  "physicalDescription": "xii, 672 pages.  Other physical details: illustrations, map.  Dimensions: 25 cm",
+  "workType": "Books",
+  "subjects": 
+    [
+     {
+        "label": "Northwest Passage",
+        "authority": "LCSH" 
+      },
+; Comment: Source is identified as LCSH by the second indicator 0.  NB use of MARC $651 indicates this (and the next two subjects) is a geographic subject:
```
kenoir: Would examples be better in a markdown file with commented JSON code blocks? Perhaps YAML might be a useful tool here (being essentially reducible to JSON, easier to read, more compact and with comments).

kenoir: Here's this file converted to YAML for reference: https://gist.github.com/kenoir/94fe02f938c53accb961c1f8fddb8cd0

alexwlchan: 👎 on YAML

I’d much rather keep these in JSON, so they mirror the API responses as closely as possible.

PR #425, opened 2017-06-20
Adds cluster view

kenoir:

Display the different clusters ringfencing each service:

status_clusters

### Who is this change for?

Folks who want more visibility of service status.

- [ ] Deployed new versions

**Depends:** https://github.com/wellcometrust/platform-api/pull/420


kenoir:
For: https://github.com/wellcometrust/platform-api/issues/426

PR #427, opened 2017-06-20
[WIP] Services refactoring

alicefuzier:
Reduce code duplication in services definition
### Who is this change for?
Devs

- [ ] Run `terraform apply`.


kenoir:
Why closed?

PR #428, opened 2017-06-20
Adds Events view

kenoir:

More visibility of the cluster status.

adding events

### Who is this change for?

Folks who want visibility of service status.

- [ ] Deployed new versions


kenoir:
From: https://github.com/wellcometrust/platform-api/issues/426

PR #435, opened 2017-06-21
JSON examples: moving images

ChristopherHilton:
Attached is the latest JSON sample, for moving image material; plus comments, in a separate Markdown file.

### Who is this change for?
Platform team

### DIFF
--- a/ontologies/Sierra_film.json
+++ b/ontologies/Sierra_film.json
@@ -1,0 +1,180 @@
+{
+  "@context": "https://api.wellcomecollection.org/catalogue/v0/context.json",
+  "id": "8rm28aw",
+  "type": "Work",
+  "identifiers": [
+    {
+      "label": "b24993207",
+      "authority": "Sierra"
+    }
+  ],
+  "label": "Are health tests a good idea? An Horizon special",
+  "creators": [
+    {
+      "label": "Davies, Serena, television producer, television director",
+      "type": "Agent"
+    },
+    {
+      "label": "Mosley, Michael, television presenter",
+      "type": "Agent"
+    },
+    {
+      "label": "Levy, Daniel, contributor",
+      "type": "Agent"
+    },
+    {
+      "label": "Jenkins, Paul, contributor",
+      "type": "Agent"
+    },
+    {
+      "label": "Dymond, Duncan, contributor",
+      "type": "Agent"
+    },
+    {
+      "label": "Dweck, Marc, contributor",
+      "type": "Agent"
+    },
+    {
+      "label": "Wilson, Robin, contributor",
+      "type": "Agent"
+    },
+    {
+      "label": "Heath, Iona, contributor",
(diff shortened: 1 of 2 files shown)


ontologies/Sierra_film.json
```diff
@@ -0,0 +1,180 @@
...
+  "identifiers": [
+    {
+      "label": "b24993207",
+      "authority": "Sierra"
+    }
+  ],
+  "label": "Are health tests a good idea? An Horizon special",
+  "creators": [
+    {
+      "label": "Davies, Serena, television producer, television director",
```
alexwlchan: At some point, do we want the Agent ontology to have some notion of "role"? So this could eventually become something like:

```json
{
  "label": "Davies, Serena",
  "role": [
    "television producer",
    "television directory"
  ],
  "type": "Agent"
}
```

So if Serena has contributed in another capacity, there's still a way to find all their images in one go.

(I’m aware CALM doesn’t actually expose that info.)

ontologies/Sierra_film.json
```diff
@@ -0,0 +1,180 @@
...
+    },
+    {
+      "label": "two-dimensional moving image: content"
+    },
+    {
+      "label": "video: media"
+    },
+    {
+      "label": "videodisc: carrier"
+    }
```
alexwlchan: Why are these single item dictionaries and not just a flat list? When would `format` have an additional field?

ontologies/Sierra_film.md
```diff
@@ -0,0 +1,12 @@
+# JSON examples: Sierra data describing moving images
+
+Comment at lines 90-109: these various notes relating to the format of the item come out of slightly different MARC fields.
```
kenoir: Nice!

Don't forget you can also encapsulate json in markdown (at least in Github Markdown).

e.g.

Here is some json:

```json
{ 
  "some": "json"
}
```

ontologies/Sierra_film.json
ChristopherHilton: Yes, absolutely: it's something that the new cataloguing paradigms for books (RDA) and archives (RiC) are both keen on, and I keep banging on to Axiell that future versions of CALM are going to have to support this.  Even if the data currently coming down the pipe from CALM doesn't show this information, I'm hopeful that it will.  So yes, Agents ontology will need to include this!

ontologies/Sierra_film.json
ChristopherHilton: Basically a bit of a placeholder: they do come out of slightly differing MARC fields, so it's a matter of the extent to which we would want to expose that sort of subtlety to the reader - in that case I'm assuming that we might acquire a second item in the dictionary, setting out what sort of Format information each might have?

ontologies/Sierra_film.md
ChristopherHilton: Thanks!

Issue #424, opened 2017-06-20
Swap EC2 instances in the ECS cluster without breaking services running


alicefuzier:
If we change the cloud config for an EC2 instance that's part of our ECS cluster, AWS starts the new one and shuts down the old one, but it doesn't check that the new one has an ECS agent running and it doesn't migrate containers running in the old instance to the new one before shutting down the old one. 
This means that if we want to make changes to our underlying Ec2 instances we will have service outages, which is obviously not acceptable, especially for the API

alicefuzier:
This might help: https://aws.amazon.com/blogs/compute/how-to-automate-container-instance-draining-in-amazon-ecs/

PR #436, opened 2017-06-22
Set up a way of doing blue green deployments of EC2 instances

alicefuzier:
Related to #424 . Make our applications not go down if we change the underlying Ec2 instances
### Who is this change for?
Devs and everyone who cares about infrastructure reliability

- [x] Run `terraform apply`.


kenoir:
Probably worth mentioning somewhere this is cribbed from https://aws.amazon.com/blogs/compute/how-to-automate-container-instance-draining-in-amazon-ecs/.

alicefuzier:
Oh I put it in the issue, but probably it makes more sense in the PR

lambdas/drain_ecs_container_instance/drain_ecs_container_instance.py
```diff
@@ -0,0 +1,64 @@
...
+            asg_client.record_lifecycle_action_heartbeat(
+                LifecycleHookName=lifecycle_hook_name,
+                AutoScalingGroupName=asg_group_name,
+                LifecycleActionToken=lifecycle_action_token,
+                InstanceId=ec2_instance_id
+            )
+            container_instance_info = ecs_client.describe_container_instances(cluster=cluster_arn,containerInstances=[ecs_container_instance_arn])
+            if container_instance_info['containerInstances'][0]['status'] != 'DRAINING':
+                set_container_instance_to_draining(cluster_arn, ecs_client, ecs_container_instance_arn)
+            publish_sns_message(topic_arn, message_data)
```
alicefuzier: Sleep for a minute

lambdas/drain_ecs_container_instance/drain_ecs_container_instance.py
```diff
@@ -0,0 +1,64 @@
+import json
+import pprint
+import boto3
+from sns_utils import publish_sns_message
```
alicefuzier: Fix imports

lambdas/drain_ecs_container_instance/drain_ecs_container_instance.py
```diff
@@ -0,0 +1,98 @@
...
+        'cluster_arns': cluster_arns,
+        'ecs_container_instance_arns': ecs_container_instance_arns
+    }
+
+
+def main(event, _):
+    asg_client = boto3.client("autoscaling")
+    ec2_client = boto3.client("ec2")
+    ecs_client = boto3.client('ecs')
+    print(f'Received event:\n{pprint.pformat(event)}')
```
kenoir: Can we get some newlines in here to make it a bit more readable?

terraform/ecs_asg/variables.tf
```diff
@@ -51,3 +51,11 @@ variable "admin_cidr_ingress" {
   default     = "0.0.0.0/0"
   description = "CIDR for SSH access to EC2 instances"
 }
+
+variable "sns_topic_arn" {
+  description = "ARN of the topic where to push notifications when an EC2 instance is set to terminating state"
+}
+
+variable "publish_to_sns_policy" {
+  description = "Policy document to publish to the sns topic"
```
alexwlchan: Is this something that's going to be published _to_ the SNS document, or something that allows us to publish to the SNS topic? I’m not quite clear.

terraform/ecs_asg/iam.tf
```diff
@@ -0,0 +1,23 @@
+resource "aws_iam_role_policy" "publish_to_sns" {
+  name   = "publish_to_sns"
+  role   = "${aws_iam_role.role.id}"
+  policy = "${var.publish_to_sns_policy}"
+}
```
alexwlchan: Where is this resource used?

Separately, giving it a fixed name may cause issues – names are in a global namespace, I think, so you’d end up blatting over policy from a different instance of the module.

terraform/ecs_asg/asg.tf
```diff
@@ -12,11 +12,13 @@ data "template_file" "cluster_asg" {
...
-    asg_min_size        = "${var.asg_min}"
-    asg_desired_size    = "${var.asg_desired}"
-    asg_max_size        = "${var.asg_max}"
+    launch_config_name   = "${aws_launch_configuration.launch_config.name}"
+    vpc_zone_identifier  = "${jsonencode(var.subnet_list)}"
+    asg_min_size         = "${var.asg_min}"
+    asg_desired_size     = "${var.asg_desired}"
+    asg_max_size         = "${var.asg_max}"
+    sns_topic_arn        = "${var.sns_topic_arn}"
+    sns_publish_role_arn = "${aws_iam_role.role.arn}"
```
alexwlchan: Should this be pointing to something SNS related?

terraform/iam_role_policy.tf
```diff
@@ -160,3 +160,33 @@ resource "aws_iam_role_policy" "ecs_grafana_task_cloudwatch_read" {
...
+  name   = "drain_ecs_container_instance_ec2_describe"
+  role   = "${module.lambda_drain_ecs_container_instance.role_name}"
+  policy = "${data.aws_iam_policy_document.ec2_describe_instances.json}"
+}
+
+resource "aws_iam_role_policy" "drain_ecs_container_send_asg_heartbeat" {
+  name   = "drain_ecs_container_instance_send_asg_heartbeat"
+  role   = "${module.lambda_drain_ecs_container_instance.role_name}"
+  policy = "${data.aws_iam_policy_document.send_asg_heartbeat.json}"
+}
```
alexwlchan: There are comments in the rest of this file to break it up, telling us where one set of role policies end and another begins. Could we get one here please?

terraform/lambda/trigger_cloudwatch/cloudwatch.tf
```diff
@@ -21,6 +21,6 @@ resource "aws_cloudwatch_event_target" "event_trigger_custom" {
 resource "aws_cloudwatch_event_target" "event_trigger" {
   count = "${1 - var.custom_input}"
```
alexwlchan: We can lose the newline, can’t we?

terraform/lambda/variables.tf
```diff
@@ -19,3 +19,8 @@ variable "environment_variables" {
     EMPTY_VARIABLE = ""
   }
 }
+
+variable "timeout" {
+  description = "The amount of time your Lambda Function has to run in seconds. Defaults to 3"
```
alexwlchan: nit: "Function" doesn't need to be capitalised. Don't repeat the default in the description.

lambdas/drain_ecs_container_instance/drain_ecs_container_instance.py
```diff
@@ -0,0 +1,103 @@
...
+        'ecs_container_instance_arns': ecs_container_instance_arns
+    }
+
+
+def main(event, _):
+    asg_client = boto3.client("autoscaling")
+    ec2_client = boto3.client("ec2")
+    ecs_client = boto3.client('ecs')
+
+    print(f'Received event:\n{pprint.pformat(event)}')
```
alexwlchan: This should be the first thing we do in the `main()` function, in case something goes wrong setting up the `boto3` clients (e.g. bad credentials).

lambdas/drain_ecs_container_instance/drain_ecs_container_instance.py
```diff
@@ -0,0 +1,103 @@
...
+
+    topic_arn = event['Records'][0]['Sns']['TopicArn']
+    message = event['Records'][0]['Sns']['Message']
```
alexwlchan: Let’s log both of these out – a lot of downstream code depends on these being correct, so it would be good to have a check on them.

lambdas/drain_ecs_container_instance/drain_ecs_container_instance.py
```diff
@@ -0,0 +1,103 @@
...
+    lifecycle_action_token = message_data['LifecycleActionToken']
+
+    if lifecycle_transition == 'autoscaling:EC2_INSTANCE_TERMINATING':
+        ecs_info = get_ecs_info_from_tags(ec2_client, ec2_instance_id)
+
+        if not ecs_info['cluster_arns'] and not ecs_info['ecs_container_instance_arns']:
+            continue_lifecycle_action(asg_group_name, ec2_instance_id, lifecycle_hook_name, asg_client)
+            return
+
+        cluster_arn=ecs_info['cluster_arns'][0]
```
alexwlchan: nit: space around equals sign when used for assignment

alexwlchan: Also, what if this list contains more than one element?

lambdas/drain_ecs_container_instance/drain_ecs_container_instance.py
```diff
@@ -0,0 +1,103 @@
...
+        ecs_container_instance_arn=ecs_info['ecs_container_instance_arns'][0]
+        running_tasks = ecs_client.list_tasks(
+            cluster=ecs_info['cluster_arns'][0],
+            containerInstance=ecs_info['ecs_container_instance_arns'][0]
+        )
+        print(f"running tasks: {running_tasks['taskArns']}")
+
+        if not running_tasks['taskArns']:
```
alexwlchan: 👍 nice use of the implicit truthiness of lists

lambdas/drain_ecs_container_instance/drain_ecs_container_instance.py
alicefuzier: They are part of the event that's printed above...I'm worried that if we add this we will have too much clutter in our logs..

lambdas/drain_ecs_container_instance/drain_ecs_container_instance.py
alicefuzier: it should not happen..I guess the lambda should crash if that's the case?

terraform/ecs_asg/variables.tf
alicefuzier: It's the policy document to give permission to publish to the sns topic

terraform/ecs_asg/iam.tf
alicefuzier: It's only used there. It's to give permission the tole declared below to publish to use the policy passed in in the variables

alicefuzier: I'll change the name btw

terraform/lambda/variables.tf
alicefuzier: Literally copied and pasted the terraform documentation

lambdas/drain_ecs_container_instance/drain_ecs_container_instance.py
alicefuzier: Actually that cannot happen because the format of what's returned by EC2 is in key, value pairs format. The list comes form the filtering in the `get_ecs_info_from_tags` function, so it will always have size zero or one

terraform/iam_role_policy.tf
alicefuzier: 👍

terraform/lambda/trigger_cloudwatch/cloudwatch.tf
alicefuzier: nope, that's terraform fmt adding it

lambdas/drain_ecs_container_instance/drain_ecs_container_instance.py
alexwlchan: Good point, we can always re-run the parsing code locally.

terraform/ecs_asg/variables.tf
alexwlchan: Okay, could we put that comment in the description please? I think that’s a clearer way of putting it.

lambdas/drain_ecs_container_instance/drain_ecs_container_instance.py
```diff
@@ -0,0 +1,100 @@
...
+    pprint.pprint(response)
+
+
+def get_ecs_info_from_tags(ec2_client, ec2_instance_id):
+    ec2_instance_info = ec2_client.describe_instances(InstanceIds=[
+        ec2_instance_id,
+    ])
+    tags = ec2_instance_info['Reservations'][0]['Instances'][0]['Tags']
+    ecs_container_instance_arns = [x['Value'] for x in tags if x['Key'] == 'containerInstanceArn']
+    cluster_arns = [x['Value'] for x in tags if x['Key'] == 'clusterArn']
```
alexwlchan: An alternative way we could do this would be to unpack this back into a proper dictionary object:

```python
tag_dict = {t['Key']: t['Value'] for t in tags}
```

You could then query individual items like a dict:

```python
ecs_container_instance_arn = tag_dict['containerInstanceArn']
cluster_arn = tag_dict['clusterArn']
```

and catch the `KeyError` if the key is missing. Might that be simpler?

terraform/ecs_asg/iam.tf
```diff
@@ -0,0 +1,23 @@
...
+  role   = "${aws_iam_role.sns_publish_role.id}"
+  policy = "${var.publish_to_sns_policy}"
+}
+
+resource "aws_iam_role" "sns_publish_role" {
+  name               = "${var.asg_name}_role"
+  assume_role_policy = "${data.aws_iam_policy_document.assume_ec2_role.json}"
+}
+
+data "aws_iam_policy_document" "assume_ec2_role" {
```
alexwlchan: Is this assuming an EC2 role, or an autoscaling role?

lambdas/drain_ecs_container_instance/drain_ecs_container_instance.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def get_ec2_tags(ec2_client, ec2_instance_id):
+    ec2_instance_info = ec2_client.describe_instances(InstanceIds=[
+        ec2_instance_id,
+    ])
+    tags = ec2_instance_info['Reservations'][0]['Instances'][0]['Tags']
+    tag_dict = {t['Key']: t['Value'] for t in tags}
+    print(f"tag_dict: {tag_dict}")
+    return tag_dict
```
alexwlchan: 👍 nice tidy change

lambdas/drain_ecs_container_instance/drain_ecs_container_instance.py
```diff
@@ -0,0 +1,97 @@
...
+    lifecycle_transition = message_data['LifecycleTransition']
+    lifecycle_action_token = message_data['LifecycleActionToken']
+
+    if lifecycle_transition == 'autoscaling:EC2_INSTANCE_TERMINATING':
+        tags_dict = get_ec2_tags(ec2_client, ec2_instance_id)
+
+        try:
+            cluster_arn = tags_dict['clusterArn']
+            ecs_container_instance_arn = tags_dict['containerInstanceArn']
+        except KeyError as e:
```
alexwlchan: no need for `as e:`, we're not using the variable

lambdas/drain_ecs_container_instance/drain_ecs_container_instance.py
```diff
@@ -0,0 +1,97 @@
...
+            return
+
+        running_tasks = ecs_client.list_tasks(
+            cluster=cluster_arn,
+            containerInstance=ecs_container_instance_arn
+        )
+        print(f"running tasks: {running_tasks['taskArns']}")
+
+        if not running_tasks['taskArns']:
+            continue_lifecycle_action(asg_client, asg_group_name, ec2_instance_id, lifecycle_hook_name)
```
alexwlchan: maybe just stick a `return` here and dedent the `else:` block?

PR #442, opened 2017-06-22
JSON book example

ChristopherHilton:
Removal of comments into Markdown, correction of JSON errors
### Who is this change for?
Platform team

### DIFF
--- a/MIRO_sample_image_of_digital-only_image.md
+++ b/MIRO_sample_image_of_digital-only_image.md
@@ -1,0 +1,22 @@
+
+# JSON samples: MIRO images that are free-standing, without an analogue in other Library systems
+
+Comment on lines 12-13: 
+This is a non-standard reference, provided by the organisation that provided the image.  
+I assume here that we would be able to set up something where if the label began "Contributor Reference" and the data is from MIRO, we could supply
+"MIRO contributor" as an authority
+
+Comment, line 13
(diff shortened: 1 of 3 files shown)


alexwlchan:
This doesn’t look right to me – the comments in the Markdown file don’t go with this JSON file.

This is probably going to be easier to work through in person – @ChristopherHilton, let’s sit down tomorrow and work through the pile of patches together. Content all looks good, but making sure they’re easily reviewable is important too. 😄

ChristopherHilton:
Grand, thanks: I was thinking as I put this blizzard of stuff in that it could do with being neatly organised so that JSON and comments are easily linked.  I’m writing up a Library blog thing at home today but let’s talk about it on Monday?
Chris

alexwlchan:
> I’m writing up a Library blog thing at home today but let’s talk about it on Monday?

Sounds good to me!

ChristopherHilton:
BTW, while this is fresh in my mind I noticed yesterday that I’d put the JSON Sierra Book sample (https://github.com/wellcometrust/platform-api/blob/master/Sierra_book.json) into the master top-level repository rather than the Ontologies folder - would be useful to shift this as part of the same chat.  (The mapping of the repository onto my shared O drive doesn’t update, so I don’t have the option of shifting it via Windows Explorer without a long preceding dialogue with IT – if there are ways within Github of moving it that sounds a lot simpler.)
Chris

PR #444, opened 2017-06-23
Fix the Docker build on Travis

alexwlchan:

At some point, our build step to install Docker on Travis stopped working. This is blocking #443, and will start breaking any build which touches the core application code. This needs fixing for builds to work. First try: do we still need that custom Docker install code?

(We haven’t touched application code in a week, so we haven’t been hitting this code path.)

### Who is this change for?

Developers who like working builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: scala
 
 sudo: required
 
+services:
+  - docker
+
 dist: trusty
 
 branches:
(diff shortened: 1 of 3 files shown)


alexwlchan:
Ah, I wonder if [the new build images](https://blog.travis-ci.com/2017-06-19-trusty-updates-2017-Q2) are to blame.

PR #452, opened 2017-06-26
Refactor to use Python classes in blue-green deploy lambda

kenoir:


kenoir:
@alicefuzier I did a bit of an experiment with extending `namedtuple` into a class. Gives you an (immutable-ish) case-class-esque _thing_ that I think makes it easier to follow what's going on. I don't think this is that weird from a Python point of view (/cc @alexwlchan).

I've also run `flake8` over the code (and fixed the errors - they were all whitespace or line length) on @alexwlchan's advice.

There were some places where you were doing:

```py
if true:
  wrap_main_execution_in_if_statement()
  etc()
```

When I think it's more idiomatic to just `return` on the "bad" case:

```py
if false:
  return
```

lambdas/drain_ecs_container_instance/drain_ecs_container_instance.py
```diff
@@ -27,27 +75,112 @@ def set_container_instance_to_draining(cluster_arn, ecs_container_instance_arn,
...
+
+LifeCycleInfo = collections.namedtuple(
+    'LifecycleInfo',
+    'hook_name transition action_token')
+
+class LifecycleEvent(collections.namedtuple(
+    'LifecycleEvent',
+    'ec2_info lifecycle_info topic_arn')):
+
+    @classmethod
```
kenoir: @alexwlchan should these be `staticmethod`?

lambdas/drain_ecs_container_instance/drain_ecs_container_instance.py
```diff
@@ -57,47 +190,44 @@ def main(event, _):
...
+        print(f'continue_lifecycle_action received:\n{pprint.pformat(response)}')
+
+    asg_client.record_lifecycle_action_heartbeat(
+        LifecycleHookName=lifecycle_event.lifecycle_info.hook_name,
+        AutoScalingGroupName=lifecycle_event.ec2_info.asg_group_name,
+        LifecycleActionToken=lifecycle_event.lifecycle_info.action_token,
+        InstanceId=lifecycle_event.ec2_info.ec2_instance_id,
+    )
+
+    if not ecs_info.is_draining:
```
kenoir: maybe flip the logic here so it's clearer

lambdas/drain_ecs_container_instance/drain_ecs_container_instance.py
```diff
@@ -57,47 +190,44 @@ def main(event, _):
...
+    ecs_info = ECSInfo.create(
+        ecs_client,
+        ec2_client,
+        lifecycle_event.ec2_info.ec2_instance_id
+    )
 
-        if not ecs_info['cluster_arns'] and not ecs_info['ecs_container_instance_arns']:
-            continue_lifecycle_action(asg_group_name, ec2_instance_id, lifecycle_hook_name, asg_client)
-            return
+    def _continue_lifecycle_action():
```
kenoir: Is nested def here necessary?

kenoir:
Closing in favour of existing PR.

PR #456, opened 2017-06-28
Adds pagination links to the API results where appropriate

kenoir:

In order to move the burden of working out where the next page of results is from the client to the server we will provide next/prev links on API results where appropriate so that the client can simplify pagination logic.

### Who is this change for?

Clients consuming multi-page query requests - specifically the Digital Experience team.

- [x] Reviewed by @jtweed
- [x] Updated the Swagger documentation
- [X] Added tests
- [x] Deployed new versions

Fixes: https://github.com/wellcometrust/platform-api/issues/453

An example API response:

`GET https://localhost:8888/catalogue/v0/works?page=2&pageSize=1`

```json
{
  "@context" : "https://localhost:8888/catalogue/v0/context.json",
  "pageSize" : 1,
  "totalPages" : 3,
  "totalResults" : 3,
  "results" : [
    {
      "id" : "2-1234",
      "label" : "2-this is the first image title",
      "description" : "2-this is a description",
      "lettering" : "2-some lettering",
      "createdDate" : {
        "label" : "2-the past",
        "type" : "Period"
      },
      "creators" : [
        {
          "label" : "2-a person",
          "type" : "Agent"
        }
      ],
      "type" : "Work"
    }
  ],
  "prev" : "https://localhost:8888/catalogue/v0/works?page=1&pageSize=1",
  "next" : "https://localhost:8888/catalogue/v0/works?page=3&pageSize=1",
  "type" : "ResultList"
}
```

### DIFF
--- a/api/src/main/scala/uk/ac/wellcome/platform/api/responses/Responses.scala
+++ b/api/src/main/scala/uk/ac/wellcome/platform/api/responses/Responses.scala
@@ -18,8 +18,8 @@ case class ResultListResponse(
   totalPages: Int = 10,
   totalResults: Int = 100,
   results: Array[_ <: Any],
-  prev: Option[String] = None,
-  next: Option[String] = None
+  prevPage: Option[String] = None,
+  nextPage: Option[String] = None
 ) {
   @JsonProperty("type") val ontologyType: String = "ResultList"
 }
@@ -61,8 +61,8 @@ object ResultListResponse {
       pageSize = displaySearch.pageSize,
       totalPages = displaySearch.totalPages,
       totalResults = displaySearch.totalResults,
-      prev = prevLink,
-      next = nextLink
+      prevPage = prevLink,
+      nextPage = nextLink
     )
   }
 }
--- a/api/src/main/scala/uk/ac/wellcome/platform/api/responses/Responses.scala
+++ b/api/src/main/scala/uk/ac/wellcome/platform/api/responses/Responses.scala
@@ -1,6 +1,10 @@
 package uk.ac.wellcome.platform.api.responses
 
 import com.fasterxml.jackson.annotation.{JsonProperty, JsonUnwrapped}
+import com.twitter.finagle.http.ParamMap
+import uk.ac.wellcome.platform.api.requests.{ApiRequest, MultipleResultsRequest}
+import uk.ac.wellcome.platform.api.models.DisplaySearch
+
 import scala.language.existentials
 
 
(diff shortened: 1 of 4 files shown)


api/src/main/scala/uk/ac/wellcome/platform/api/responses/Responses.scala
```diff
@@ -14,7 +18,64 @@ case class ResultListResponse(
...
+  results: Array[_ <: Any],
+  prev: Option[String] = None,
+  next: Option[String] = None
 ) {
   @JsonProperty("type") val ontologyType: String = "ResultList"
 }
+
+object ResultListResponse {
+
+  private def getParamMapFromCustomCCRequestParams(cc: ApiRequest): ParamMap =
```
kenoir: There is some nasty reflection going on here to get back a Finagle ParamMap from our "Custom Request Case Class" - I couldn't find a publicly accessible way of doing this in the Finatra code.

This code will work for _single level_ case classes, but will fall over when asked to perform this operation on a nested case class.

api/src/main/scala/uk/ac/wellcome/platform/api/responses/Responses.scala
```diff
@@ -14,7 +18,64 @@ case class ResultListResponse(
...
+    ParamMap((Map[String, String]() /: cc.getClass.getDeclaredFields) {(a, f) =>
+      f.setAccessible(true)
+
+      if(f.getName != "request") {
```
kenoir: The "Custom Request Case Class" can have a request object on it (this is how you get back a plain request when using a custom case class) which Finatra will populate with the original request object - we don't want that to appear in the api link we return to the user!

api/src/main/scala/uk/ac/wellcome/platform/api/responses/Responses.scala
```diff
@@ -14,7 +18,64 @@ case class ResultListResponse(
...
+          case None => a
+          case Some(o) => a + (f.getName -> o.toString)
+          case o => a + (f.getName -> o.toString)
+        }
+      } else {
+        a
+      }
+    })
+
+  private def apiLink(
```
kenoir: Perhaps this should actually be a separate object? i.e.

```scala
case class ApiLink(link: String) {
  def toString = link
}
object ApiLink {
  private def getParamMapFromCustomCCRequestParams(cc: ApiRequest): ParamMap = ...

  def apply(requestBaseUri: String , apiRequest: ApiRequest): ApiLink = {
    val customRequestParamMap = getParamMapFromCustomCCRequestParams(apiRequest)

    ApiLink(s"${requestBaseUri}${apiRequest.request.path}${customRequestParamMap.toString()}")
  }
}
```

api/src/main/scala/uk/ac/wellcome/platform/api/responses/Responses.scala
```diff
@@ -14,7 +18,64 @@ case class ResultListResponse(
...
+              displaySearch: DisplaySearch,
+              multipleResultsRequest: MultipleResultsRequest,
+              requestBaseUri: String
+            ): ResultListResponse = {
+
+    val isLastPage = displaySearch.totalPages == multipleResultsRequest.page
+    val isFirstPage = multipleResultsRequest.page == 1
+    val isOnlyPage = displaySearch.totalPages <= 1
+
+    val prevPageRequest = multipleResultsRequest.copy(page = multipleResultsRequest.page - 1)
```
kenoir: Do we care about generating this every time even though they might not be used?

api/src/test/scala/uk/ac/wellcome/platform/api/ApiWorksTest.scala
```diff
@@ -145,10 +145,11 @@ class ApiWorksTest
   }
 
   it(
-    "should return the requested page of results when requested with page & pageSize") {
```
kenoir: Squeezed the links tests in here. I think it makes sense as this test should really be covering the edge cases (first/last) too.

kenoir:
An alternative to this approach would be to put the pagination links in their own block:

```json
{
  "...": "...",
  "links": [
     { "type": "next", "href": "http://example.com?page=42" } 
  ]
}
```

This is arguably more descriptive and extendable but YMMV.

api/src/main/scala/uk/ac/wellcome/platform/api/responses/Responses.scala
kenoir: I'm aware this might be better off as a comment actually in the code. How non-obvious is this do we think?

api/src/main/scala/uk/ac/wellcome/platform/api/responses/Responses.scala
jamesgorrie: Should this note be in the code/

jtweed:
I'm not against thinking about something more extensible, but if not then I think we might be better to call them `nextPage` and `prevPage`.

kenoir:
@jtweed will add these with the suffix "Page" for now and consider adding an extra block if the need arises.

PR #459, opened 2017-06-29
Script the creation of our nginx containers

alexwlchan:

Adds a script for creating our nginx images and pushing them to ECR.

### Who is this change for?

Me, because I am literally too lazy to do anything that involves running more than one command, and I was looking at writing a new variant for the IIIF images.


kenoir:
Nice. Step on the way to automating builds too.

PR #460, opened 2017-06-29
Run a docker container to flake8 the lambdas

kenoir:

Some CI checking of our Python lambdas for consistency and to prevent simple errors reaching prod.

### Who is this change for?

xoi6zg

- [ ] Deployed new versions


alexwlchan:
👍 on the idea, but I think we can just install flake8 directly in Travis (as an install step):

```console
$ pip3 install flake8
```

I wonder whether the Docker container adds anything?

Separately, don’t forget that the pre-flight checks in `.travis.yml` will skip the build if there are changes in the Lambda directories but not anywhere else. You’ll need to change that.

kenoir:
> I wonder whether the Docker container adds anything?

I think we should be moving in this direction as this makes our tests more consistent and portable.

alexwlchan:
Also, can you push a flake8 violation to prove it breaks the build please? 😄

kenoir:
@alexwlchan re re review pls!

docker/python3.6/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.6
+
+VOLUME ["/data"]
+
+RUN pip install flake8
+
+COPY run.sh /app/run.sh
+
+ENTRYPOINT /app/run.sh
```
alexwlchan: nit: I’d call this `flake8` instead of `python3.6`.

scripts/should_rebuild_master.py
```diff
@@ -33,6 +33,10 @@ def should_rebuild_project(changed_files, project):
         print("*** Changes to the scripts directory mean we should rebuild")
         return True
 
+    elif any(f.startswith('.travis/') for f in changed_files):
```
alexwlchan: I don't think you need the trailing slash – what about `.travis.yml`?

docker/python3.6/Dockerfile
kenoir: 👍 makes sense

scripts/should_rebuild_master.py
kenoir: Yup. I agree, as this doesn't regress functionality I would merge and then or another change.

PR #461, opened 2017-06-29
Redirect requests to iiif.wellcomecollection.org/image to loris

alicefuzier:
Make loris accessible from https://iiif.wellcomecollection.org/image
### Who is this change for?
Everyone accessing our iiif API

- [ ] Run `terraform apply`.

### DIFF
--- a/terraform/services/ecs_service/alb.tf
+++ b/terraform/services/ecs_service/alb.tf
@@ -12,7 +12,8 @@ resource "aws_alb_target_group" "ecs_service" {
   }
 }
 
-resource "aws_alb_listener_rule" "rule" {
+resource "aws_alb_listener_rule" "path_rule" {
+  count        = "${1 - var.use_host_condition}"
   listener_arn = "${var.listener_arn}"
   priority     = "${var.alb_priority}"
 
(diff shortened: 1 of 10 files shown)


terraform/services/ecs_service/alb.tf
```diff
@@ -12,7 +12,8 @@ resource "aws_alb_target_group" "ecs_service" {
   }
 }
 
-resource "aws_alb_listener_rule" "rule" {
+resource "aws_alb_listener_rule" "path_rule" {
+  count        = "${1 - var.use_host_condition}"
```
alexwlchan: I think we can use [conditionals](https://www.terraform.io/docs/configuration/interpolation.html#conditionals) to achieve the same effect, and do away with the need for a `host_condition` variable. Something like:

```hcl
count  = "${var.host_name == "" ? 0 : 1}"
```

kenoir: Nice. Didn't know these existed. Apparently they came in v0.8: https://www.hashicorp.com/blog/terraform-0-8/#conditional

alexwlchan: Yeah, I was looking at using them for the GSIs in DynamoDB, but you can’t stick a `count` on a GSI. 😕

PR #464, opened 2017-06-30
Loris nginx config

alicefuzier:
Make loris rewrite Urls for miro images so that our clients don't need to know the structure of the s3 bucket
### Who is this change for?
Users

- [ ] Run `terraform apply`.

### DIFF
--- a/docker/nginx/loris.nginx.conf
+++ b/docker/nginx/loris.nginx.conf
@@ -6,6 +6,17 @@ http {
   server {
     listen 9000;
 
+    # Requests for iiif.wellcomecollection.org/image should be forwarded to our IIIF Image API, Loris
+    #
+    # Wellcome Collection V images are currenly hosted in an s3 bucket and
+    # they are sharded based on their file name, so we need to construct the image path in
+    # the s3 bucket when we send the request to the iiif image API.
+    #
+    # File names start with V and are followed by 7 digits. The name of the directory in which each
+    # file is stored is based on the first 4 digits of its filename.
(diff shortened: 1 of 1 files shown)


docker/nginx/loris.nginx.conf
```diff
@@ -6,8 +6,8 @@ http {
   server {
     listen 9000;
 
-    location ~ ^/image/V[0-9]{4}.* {
-      rewrite ^/image/(V([0-9]{4}).*) /V$2000/$1 break;
+    location ~ ^/image/V[0-9]\{4\}.* {
+      rewrite ^/image/(?<filename>(?<shard>V[0-9]\{4\}).*) /s3:\k<shard>000/\k<filename> break;
```
alexwlchan: Could we get a comment explaining what the original URLs look like, and what we're trying to rewrite them too? That would make the regex much easier to follow.

docker/nginx/loris.nginx.conf
```diff
@@ -6,8 +6,8 @@ http {
   server {
     listen 9000;
 
-    location ~ ^/image/V[0-9]{4}.* {
-      rewrite ^/image/(V([0-9]{4}).*) /V$2000/$1 break;
+    location ~ "^/image/V[0-9]{4}.*" {
+      rewrite "^/image/(?<filename>(?<shard>V[0-9]{4}).*)" "/s3:${shard}000/${filename}" break;
```
alexwlchan: Can we add a quick comment explaining what sort of URLs we're trying to match, and how we're rewriting them? Make the regex comprehensible to the next person who comes along. :)

PR #468, opened 2017-06-30
Consolidate certificates into the ALB module

alexwlchan:
WIP.

PR #476, opened 2017-07-03
Redirect requests for http://wc.org to https://wc.org

### LINKED ISSUE #264: HTTP redirect to HTTPS isn't working
If I visit http://api.wellcomecollection.org/catalogue/v0/works, my browser hangs forever.

If I add HTTPS, the page returns immediately.

We should either return a response to HTTP requests, or redirect to HTTPS.

alexwlchan:

With the config on master, requests to http://wc.org will simply timeout. This redirects them to HTTPS.

Resolves #264. Builds atop #468.

### Who is this change for?

Users who want their connection secured with HTTPS, which protects it with words like TLS, RSA and ECC. I don’t know what they mean, but those are all very encouraging words!

(White lie: I do have some idea of what those mean, talk to me if you want an explanation.)

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


alexwlchan:
Okay, I think this is ready to review.

```console
$ curl -I 'http://iiif.wellcomecollection.org/image'
HTTP/1.1 301 Moved Permanently
Date: Mon, 03 Jul 2017 12:52:59 GMT
Content-Type: text/html
Content-Length: 185
Connection: keep-alive
Server: nginx/1.13.0
Location: https://iiif.wellcomecollection.org/image
Strict-Transport-Security: max-age=31536000; includeSubDomains;

$ curl -I 'https://iiif.wellcomecollection.org/image'
HTTP/1.1 200 OK
Date: Mon, 03 Jul 2017 12:53:07 GMT
Content-Type: text/plain
Content-Length: 220
Connection: keep-alive
Server: nginx/1.13.0
ETag: "1acdb93e43e76c08a2e2c037bad4c554"
Accept-Ranges: none
Strict-Transport-Security: max-age=31536000; includeSubDomains;
```

docker/nginx/api.nginx.conf
```diff
@@ -5,6 +5,21 @@ events { worker_connections 1024; }
...
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_set_header Host $http_host;
+    proxy_redirect off;
+    proxy_next_upstream error;
+
+    if ($http_x_forwarded_proto != "https") {
+      rewrite ^(.*)$ https://api.wellcomecollection.org$1 permanent;
+    }
+
+    add_header Strict-Transport-Security "max-age=31536000; includeSubDomains;";
```
kenoir: HSTS! Whoop. @jamesgorrie lookit 👀

jamesgorrie: [We're doing it here](https://github.com/wellcometrust/wellcomecollection.org/blob/74689a702475098683eed84d66fc073eae77eb91/server/middleware/enforce-ssl.js).
I'd like to move it to nginx so nginx is doing all things http.

PR #477, opened 2017-07-03
Lint JS

kenoir:

Checking for valid JS in CI where appropriate, specifically in our case ontology examples, in order that we publish valid examples!

### Who is this change for?

Developers (internal & external).


.travis.yml
```diff
@@ -83,6 +83,9 @@ matrix:
     - env:
         - BUILD_TYPE=python
         - PROJECT=lambdas
+    - env:
+        - BUILD_TYPE=js
```
alexwlchan: Minor: I’d call this `jslint` instead: we’re not doing any actual tests!

docker/js_ci/run.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
+
+set -o errexit
+set -o nounset
+set -o verbose
+
+find /data/*.js* | xargs jshint
```
alexwlchan: Since this is a single command, can we do it directly in the Dockerfile rather than adding another shell script?

docker/scala_ci/run.sh
```diff
@@ -0,0 +1 @@
+#!/usr/bin/env bash
```
alexwlchan: This doesn’t need to be here.

docker/js_ci/run.sh
kenoir: I'd rather try to extract the "run" step into a separate bash script for consistency (some builds will require multiple commands), and it's generally less fiddly to run shell commands in a shell.

.travis.yml
kenoir: 👍

docker/scala_ci/run.sh
kenoir: 👍

kenoir:
@alexwlchan "Makefile in the top-level of the repo" <- I like it.

.travis/run_sbt.sh
```diff
@@ -10,30 +10,31 @@
...
+#fi
 
-export VERSION="0.0.1"
-export BUILD_ENV="prod"
-export RELEASE_ID="$VERSION-$(git rev-parse HEAD)_$BUILD_ENV"
-export TAG="$AWS_ECR_REPO/uk.ac.wellcome/$PROJECT:$RELEASE_ID"
+#export VERSION="0.0.1"
+#export BUILD_ENV="prod"
+#export RELEASE_ID="$VERSION-$(git rev-parse HEAD)_$BUILD_ENV"
+#export TAG="$AWS_ECR_REPO/uk.ac.wellcome/$PROJECT:$RELEASE_ID"
```
alexwlchan: ❗️❗️❗️

PR #502, opened 2017-07-03
Swap the two linting tasks to use make

alexwlchan:

Start to define tasks in a Makefile, with a long-term view that all our build processes are:

* Run using Docker
* Triggered using Make

Right now just the two linting tasks, because those are the easy bits. Migrating sbt will likely be a bit harder. A first step towards #482.

### Who is this change for?

Developers who want consistent, easy-to-manage build environments.


.travis.yml
```diff
@@ -81,10 +81,10 @@ matrix:
         - BUILD_TYPE=sbt
         - PROJECT=reindexer
     - env:
-        - BUILD_TYPE=python
+        - TASK=lint-lambdas
```
kenoir: These were intentionally generic, I think the BUILD_TYPE should be a more general case rather than so closely linked to the project.

alexwlchan:
I agree in principle, the difficulty is that passing variables to Make tasks is typically Not The Done Thing. More common to define lots of individual tasks, even if it is a bit repetitive. And it’s nice to run (for example):

```console
$ make lint-ontologies
```

and not have to worry about environment variables, shell config, or anything like that. So I think this is fine as-is.

kenoir:
@alexwlchan make tasks with more specific names is fine - the BUILD_TYPES matrix is where I would prefer it to be generic/

alexwlchan:
I’m still not sure what you mean.

I feel like there’s a subtle difference of understanding here that’s going to be hard to resolve via GitHub comments. Since this isn’t urgent, how about we discuss this in-person tomorrow?

kenoir:
@alexwlchan ah, my fault. I didn't see BUILD_TYPE -> had become TASK. Makes a lot more sense now.

PR #503, opened 2017-07-03
Send slack notification for alarms

alicefuzier:
Send slack notification for alarms
### Who is this change for?
Devs

- [x] Run `terraform apply`.


lambdas/post_to_slack/post_to_slack.py
```diff
@@ -0,0 +1,55 @@
...
+    print(f'Received event:\n{pprint.pformat(event)}')
+    message = event['Records'][0]['Sns']['Message']
+    message_data = json.loads(message)
+    alarm_name = message_data['AlarmName']
+    namespace = message_data['Trigger']['Namespace']
+    metric_name = message_data['Trigger']['MetricName']
+    dimensions = message_data['Trigger']['Dimensions']
+    state_reason = message_data['NewStateReason']
+
+    slack_data = {'username': 'cloudwatch-alert',
```
kenoir: I wasn't sure whether it would make sense to encapsulate this logic in a class, but this Lambda is so succinct as it is it's probably unnecessary.

kenoir:
`flake8` is upset about whitespace

```sh
+docker run -v /home/travis/build/wellcometrust/platform-api/lambdas:/data python3.6_ci:latest
flake8 /data
/data/post_to_slack/post_to_slack.py:43:5: E122 continuation line missing indentation or outdented
/data/post_to_slack/post_to_slack.py:44:5: E122 continuation line missing indentation or outdented
```

lambdas/post_to_slack/post_to_slack.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python
+# -*- encoding: utf-8 -*-
+"""
+
+"""
+
+import json
+import pprint
+import os
```
kenoir: should be an extra newline between `import os` and `from botocore.vendored import requests`

lambdas/post_to_slack/post_to_slack.py
alexwlchan: We might want to encapsulate the `message_data` parsing on a class, and use the `@property` decorator we discussed last week. Something like:

```python
class Alarm:
    def __init__(self, json_message):
        self.message = json.loads(json_message)

    @property
    def name(self):
        return self.message['AlarmName']
```

thus allowing us to call (further down):

```python
alarm = Alarm(event['Records'][0]['Sns']['Message'])
{
    'fallback': alarm.name,
    'title': alarm.name,
}
```

and so on.

lambdas/post_to_slack/post_to_slack.py
```diff
@@ -0,0 +1,55 @@
...
+                          }
+                      ]
+                  }]}
+
+    response = requests.post(
+        os.environ["SLACK_INCOMING_WEBHOOK"],
+        data=json.dumps(slack_data),
+        headers={'Content-Type': 'application/json'}
+    )
+    if response.status_code != 200:
```
alexwlchan: Neat, but can be neater still: Response objects have a `raise_for_status()` method you can use instead:

```pycon
>>> import requests
>>> r = requests.get('https://api.wellcomecollection.org/doesnotexist')
>>> r
<Response [404]>
>>> r.raise_for_status()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/chana/.virtualenvs/api/lib/python3.6/site-packages/requests/models.py", line 909, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://api.wellcomecollection.org/doesnotexist
```

terraform/asg.tf
```diff
@@ -12,6 +12,7 @@ module "services_cluster_asg" {
   instance_type         = "t2.large"
   sns_topic_arn         = "${module.ec2_terminating_topic.arn}"
   publish_to_sns_policy = "${module.ec2_terminating_topic.publish_policy}"
+  alarm_topic_arn = "${module.ec2_instance_terminating_for_too_long_alarm.arn}"
```
alexwlchan: You need to run `terraform fmt`.

terraform/variables.tf
```diff
@@ -73,3 +73,8 @@ variable "grafana_admin_user" {
 variable "grafana_admin_password" {
   description = "The password of the default Grafana admin"
 }
+
+variable "slack_webhook" {
+  description = "Incoming Webhook URL to send slack notifications"
+  default = ""
```
alexwlchan: Should this have a default?

lambdas/post_to_slack/post_to_slack.py
jamesgorrie: I wonder if we can write a bot to make these comments?
But isn't that what linting is for?

lambdas/post_to_slack/post_to_slack.py
```diff
@@ -0,0 +1,70 @@
...
+    def metric_name(self):
+        return self.message['Trigger']['MetricName']
+
+    @property
+    def dimensions(self):
+        return self.message['Trigger']['Dimensions']
+
+    @property
+    def state_reason(self):
+        return self.message['NewStateReason']
```
alexwlchan: ✨

terraform/sqs/alarms.tf
```diff
@@ -1,7 +1,3 @@
-resource "aws_sns_topic" "topic" {
-  name = "${aws_sqs_queue.dlq.name}_alarm"
-}
```
alexwlchan: Could we use the topic level sns module here instead? Would that get us what we need w/o forcing it out of the module?

lambdas/post_to_slack/post_to_slack.py
kenoir: we've got a linter, but it needs a plugin to check this for us

PR #509, opened 2017-07-03
Splits SBT test and deploy scripts

kenoir:

Some groundwork for further extracting test/deploy (really build and push ECR container).

### Who is this change for?

Future developers who want a better build.

### DIFF
--- a/.travis/deploy_sbt.sh
+++ b/.travis/deploy_sbt.sh
@@ -19,8 +19,8 @@ then
   exit 0
 fi
 
-export VERSION="0.0.1"
-export BUILD_ENV="prod"
+export VERSION="${VERSION:-0.0.1}"
+export BUILD_ENV="${BUILD_ENV:-dev}"
 export RELEASE_ID="$VERSION-$(git rev-parse HEAD)_$BUILD_ENV"
 export TAG="$AWS_ECR_REPO/uk.ac.wellcome/$PROJECT:$RELEASE_ID"
 
--- a/.travis/deploy_sbt.sh
+++ b/.travis/deploy_sbt.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o verbose
+
+# If we're on the master branch and in an application project, we should
+# build a new Docker image and push it to ECR.
+
+if [[ "$BRANCH" != "master" ]]
+then
+  echo "Not on master (BRANCH=$BRANCH); skipping deploy..."
+  exit 0
+fi
+
+if [[ "$PROJECT" == "common" ]]
+then
+  echo "Common lib doesn't have a Docker container; skipping deploy..."
+  exit 0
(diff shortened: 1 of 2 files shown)


.travis/run_sbt.sh
```diff
@@ -12,28 +12,4 @@ set -o nounset
...
-export BUILD_ENV="prod"
-export RELEASE_ID="$VERSION-$(git rev-parse HEAD)_$BUILD_ENV"
-export TAG="$AWS_ECR_REPO/uk.ac.wellcome/$PROJECT:$RELEASE_ID"
-
-./scripts/build_docker_image.sh
-
-export AWS_DEFAULT_REGION=eu-west-1
-
-./scripts/push_docker_image_to_ecr.sh
+./.travis/run_sbt.sh
```
alexwlchan: I think you meant `deploy`, because this is going to call itself forever and never do anything useful. (I’ve cancelled the build on this PR.)

.travis/deploy_sbt.sh
```diff
@@ -0,0 +1,31 @@
...
+then
+  echo "Not on master (BRANCH=$BRANCH); skipping deploy..."
+  exit 0
+fi
+
+if [[ "$PROJECT" == "common" ]]
+then
+  echo "Common lib doesn't have a Docker container; skipping deploy..."
+  exit 0
+fi
```
alexwlchan: I would pull this logic upstream into the run script – this script just deploys a given project, and leave it to the caller to determine whether it’s appropriate. Keep this script self-contained.

.travis/deploy_sbt.sh
```diff
@@ -0,0 +1,31 @@
...
+
+export VERSION="0.0.1"
+export BUILD_ENV="prod"
```
alexwlchan: I might use bash’s ability to use variable defaults here:

```bash
export VERSION="${VERSION:-0.0.1}"
export BUILD_ENV="${BUILD_ENV:-dev}"
```

So this can be used to do dev deploys if necessary.

.travis/deploy_sbt.sh
kenoir: 👍 i'm not sure it should be in the `run_script.sh` but it's appealing to keep the logic out of this script.

.travis/deploy_sbt.sh
```diff
@@ -0,0 +1,22 @@
...
+set -o verbose
+
+if [[ "$PROJECT" == "common" ]]
+then
+  echo "Common lib doesn't have a Docker container; skipping deploy..."
+  exit 0
+fi
+
+export VERSION="${VERSION:-0.0.1}"
+export BUILD_ENV="${BUILD_ENV:-dev}"
```
alexwlchan: This defaults it to dev so we can’t deploy new prod images by accident, but as a result you’ll need to add an `export BUILD_ENV=prod` to the calling script.

PR #510, opened 2017-07-04
Adds Alarms for API 5xx/4xx

kenoir:

Alerting on 5xx and 4xx responses (i.e. _bad_ or _unexpected_) from the API so we can be notified as soon as possible if there is a problem with the API.

### Who is this change for?

Anyone concerned with API uptime!

- [ ] Run `terraform apply`.


terraform/alarms.tf
```diff
@@ -15,3 +15,75 @@ resource "aws_cloudwatch_metric_alarm" "notify_old_deploys" {
...
+  statistic           = "Sum"
+  threshold           = "1"
+
+  dimensions {
+    LoadBalancer = "${module.api_alb.cloudwatch_id}"
+  }
+
+  alarm_description = "This metric monitors the API ALB _target_ for 400s (API only)"
+  alarm_actions     = []
+}
```
alexwlchan: Occurs to me that we might want to consolidate these in some sort of "critical alarm" module? To cut down repetition. Otherwise LGTM.

PR #531, opened 2017-07-04
Move terraform operations inside container

kenoir:

Further extraction of build process into containers to increase reproducibility and portability.

### Who is this change for?

💻


kenoir:
Waiting for the official hashicorp/terraform image to be 0.9.11

alicefuzier:
I wonder whether this issue with there not being an official docker image for the latest version with terraform is a general problem we might encounter with terraform builds being in docker. 
I mean, could it be that anytime we upgrade terraform our build is broken for a few days until terraform updates the official docker image? Does it mean that before upgrading we should always check if a corresponding docker image exists?

alexwlchan:
> I mean, could it be that anytime we upgrade terraform our build is broken for a few days until terraform updates the official docker image? Does it mean that before upgrading we should always check if a corresponding docker image exists?

If we move entirely to using the Docker image locally as well, then we’d never upgrade ahead of the Docker image. Hopefully a non-problem if we go Docker Everywhere™?

kenoir:
@alicefuzier what @alexwlchan said. I imagined we'd be using the `make` commands ourselves, thereby avoiding the issue.

alicefuzier:
👍

alicefuzier:
Looks ok to me, but there seems to be some kind of conflict with master? I assume `docker/nginx/manage_images.sh` should not be part of this pull request..

kenoir:
replaced by https://github.com/wellcometrust/platform-api/pull/584

Issue #478, opened 2017-07-03
Modify Transformer to use SQS (Standardise service event streams)


alicefuzier:
The transformer reads the dynamo stream and publishes to an SNS topic. This is different from how the id minter and the ingestor work, in that their input is an SQS queue. 

This has some advantages, for example the fact that we have dead letter queues and alarms for each dead letter queue already provisioned in terraform. Having the transformer behave the same way would give us more information when investigating a failure and would avoid loosing the item which failed, so that we can easily resend it.

So, we should push the logic of parsing the dynamo db stream out into a lambda that then pushes to an SNS topic and into a SQS queue, which would be the input of the transformer

kenoir:
Done and deployed - some errors, will be fixed by https://github.com/wellcometrust/platform-api/issues/675.

PR #547, opened 2017-07-05
Create lamdba to push updates to dynamo tables to sqs

alicefuzier:
Push updates to dynamo table to sqs, so that the transformer can use the sqs queue as input instead of the dynamo stream. Partially addresses #478 
### Who is this change for?
Devs

- [ ] Run `terraform apply`.


lambdas/dynamo_to_sns/dynamo_to_sns.py
```diff
@@ -0,0 +1,17 @@
...
+import json
+import os
+
+from sns_utils import publish_sns_message
+
+
+def main(event, _):
+    print(f'Received event:\n{event}')
+    stream_topic_map = json.loads(os.environ["STREAM_TOPIC_MAP"])
+    new_image = event['Records'][0]['dynamodb']['NewImage']
```
alexwlchan: At some point we might want to write nice `@property`-based wrapper for Lambda events that come from DynamoDB, but this'll do for now.

terraform/lambdas.tf
```diff
@@ -238,3 +238,39 @@ module "trigger_post_to_slack_client_error_alb" {
...
+  starting_position = "LATEST"
+}
+
+module "trigger_dynamo_to_sns_calm" {
+  source            = "./lambda/trigger_dynamo"
+  stream_arn        = "${aws_dynamodb_table.calm_table.stream_arn}"
+  function_arn      = "${module.lambda_dynamo_to_sns.arn}"
+  function_role     = "${module.lambda_dynamo_to_sns.role_name}"
+  batch_size        = 1
+  starting_position = "LATEST"
```
alexwlchan: I wonder if we should make defaults for these two arguments in the module.

terraform/lambdas.tf
```diff
@@ -238,3 +238,39 @@ module "trigger_post_to_slack_client_error_alb" {
...
+  description = ""
+  source_dir  = "../lambdas/dynamo_to_sns"
+
+  environment_variables = {
+    STREAM_TOPIC_MAP = <<EOF
+      {
+        "${aws_dynamodb_table.miro_table.stream_arn}": "${module.miro_transformer_topic.arn}",
+        "${aws_dynamodb_table.calm_table.stream_arn}": "${module.calm_transformer_topic.arn}"
+      }
+      EOF
```
alexwlchan: This is a really neat solution.

terraform/lambdas.tf
```diff
@@ -238,3 +238,39 @@ module "trigger_post_to_slack_client_error_alb" {
   lambda_function_arn  = "${module.lambda_post_to_slack.arn}"
   sns_trigger_arn      = "${module.alb_client_error_alarm.arn}"
 }
+
+# Lambda for pushing updates to dynamo tables into sqs queues
+
+module "lambda_dynamo_to_sns" {
+  source      = "./lambda"
+  name        = "dynamo_to_sns"
+  description = ""
```
alexwlchan: Can we get a description please? :)

PR #549, opened 2017-07-05
Moves ALB alarms into ALB module

kenoir:

Move ALB alarms into ALB module so all ALBs get alarms by default.

### Who is this change for?

Developers who like alarms.

- [ ] Run `terraform apply`.


terraform/ecs_alb/alarms.tf
```diff
@@ -0,0 +1,71 @@
...
+  period              = "60"
+  statistic           = "Sum"
+  threshold           = "1"
+
+  dimensions {
+    LoadBalancer = "${aws_alb.ecs_service.arn_suffix}"
+  }
+
+  alarm_description = "This metric monitors the ${var.name} ALB _target_ for 400s (API only)"
+  alarm_actions     = []
```
alicefuzier: none of these alarms have actions, so I don't think anything will get sent to `alarm_topic_arn`, will it?

terraform/alb.tf
```diff
@@ -10,6 +10,8 @@ module "services_alb" {
 
   certificate_domain = "services.wellcomecollection.org"
   vpc_id             = "${module.vpc_services.vpc_id}"
+
```
alexwlchan: any reason for the empty line?

terraform/lambdas.tf
```diff
@@ -224,3 +224,10 @@ module "trigger_post_to_slack_esg_not_terminating" {
   lambda_function_arn  = "${module.lambda_post_to_slack.arn}"
   sns_trigger_arn      = "${module.ec2_instance_terminating_for_too_long_alarm.arn}"
 }
+
+module "trigger_post_to_slack_alb" {
+  source               = "./lambda/trigger_sns"
+  lambda_function_name = "${module.lambda_post_to_slack.function_name}"
+  lambda_function_arn  = "${module.lambda_post_to_slack.arn}"
+  sns_trigger_arn      = "${module.alb_alarm.arn}"
+}
```
alexwlchan: needs a `terraform fmt`, I think

terraform/ecs_alb/alarms.tf
```diff
@@ -0,0 +1,71 @@
...
+  namespace           = "AWS/ApplicationELB"
+  period              = "60"
+  statistic           = "Sum"
+  threshold           = "1"
+
+  dimensions {
+    LoadBalancer = "${aws_alb.ecs_service.arn_suffix}"
+  }
+
+  alarm_description = "This metric monitors the ${var.name} ALB for 500s (Loris & API)"
```
alexwlchan: Now that we're in the module, are these descriptions still accurate?

terraform/ecs_alb/alarms.tf
```diff
@@ -0,0 +1,71 @@
...
+  alarm_actions     = []
+}
```
alexwlchan: Maybe these could be forced down into a further module? Something like:

```hcl
module "alb_500_errors" {
  source        = "./alb_alarm"
  alarm_name    = "${var.name}-alb-500-errors"
  metric_name   = "HTTPCode_ELB_5XX_Count"
  load_balancer = "${aws_alb.ecs_service.arn_suffix}"
  description   = "This metric monitors the ${var.name} ALB for 500s"
  alarm_actions = []
}
```

terraform/ecs_alb/variables.tf
```diff
@@ -23,3 +23,7 @@ variable "health_check_path" {
 variable "vpc_id" {
   description = "ID of VPC to create ALB in"
 }
+
+variable "alarm_topic_arn" {
```
alicefuzier: Would it be worth having two different topics for alarms? I don't think we will treat 500 errors the same way as 400 errors

terraform/alb.tf
kenoir: ease of reading

terraform/lambdas.tf
```diff
@@ -224,3 +224,17 @@ module "trigger_post_to_slack_esg_not_terminating" {
...
 }
+
+module "trigger_post_to_slack_server_error_alb" {
+  source               = "./lambda/trigger_sns"
+  lambda_function_name = "${module.lambda_post_to_slack.function_name}"
+  lambda_function_arn  = "${module.lambda_post_to_slack.arn}"
+  sns_trigger_arn      = "${module.alb_server_error_alarm.arn}"
+}
+
+module "trigger_post_to_slack_client_error_alb" {
```
alicefuzier: do we want a slack notification for every 4xx?

kenoir: It's only a notification when it goes into ALARM state. I think right now we do (we shouldn't be seeing any really) - we can dial it back when the API is under more load.

PR #552, opened 2017-07-05
Push nginx image tags to S3 alongside our other release IDs

alexwlchan:

This is a fairly substantial rewrite of one of the deploy steps. In particular, it consolidates all the work for deploying an image to ECR and S3 into a single script, and allows us to simplify the calling scripts. It also means the image tags we use locally will be a bit more tractable (the ECR image URI is only prepended during a push, never any other time).

Key point of this is to bring the nginx images under the same roof as our other applications – we now push a release ID in S3 for them. In particular:

screen shot 2017-07-05 at 11 37 18

The argument signature for this script is awful, but we need to cart around quite a bit of information. Suggestions for simplification on a postcard. I was trying to avoid having mystery environment variables. 😕 

### Who is this change for?

This is the penultimate step in bringing our nginx image deployments in line with our other applications.


docker/nginx/manage_images.sh
```diff
@@ -9,44 +9,30 @@ TASK=${1:-BUILD}
 echo "*** Task is $TASK"
 
 export AWS_DEFAULT_REGION=${AWS_DEFAULT_REGION:-eu-west-1}
+export CONFIG_BUCKET=${CONFIG_BUCKET:-platform-infra}
```
alexwlchan: We probably don't need these

docker/nginx/manage_images.sh
```diff
@@ -9,44 +9,30 @@ TASK=${1:-BUILD}
...
+  TAG="nginx:$variant-$(git rev-parse HEAD)"
   echo "*** Image will be tagged $TAG"
 
   # Actually build the image
   docker build --build-arg conf_file="$conf_file" --tag "$TAG" .
 
   if [[ "$TASK" == "DEPLOY" ]]
   then
-    docker push "$TAG"
+    $ROOT/scripts/deploy_docker_image.sh "uk.ac.wellcome/nginx" "$TAG" "$CONFIG_BUCKET" "nginx_$variant" "$(git rev-parse HEAD)"
```
alexwlchan: Maybe this should be entirely separate?

Write to a `.releases` directory in the root of the repo, and have the deploy script auto-push it for us

scripts/deploy_docker_image.sh
```diff
@@ -0,0 +1,68 @@
...
+
+echo "*** Pushing image to ECR"
+docker tag "$IMAGE_ID" "$NEW_IMAGE_ID"
+trap "docker rmi '$NEW_IMAGE_ID'" INT TERM EXIT
+docker push "$NEW_IMAGE_ID"
+
+
+# Finally, upload the release ID string to S3.
+echo "*** Uploading release ID to S3"
+echo "$RELEASE_ID" | aws s3 cp - "s3://$CONFIG_BUCKET/releases/$PROJECT"
```
alexwlchan: blergh, bash. Python instead?

.travis/deploy_sbt.sh
```diff
@@ -13,10 +13,10 @@ fi
 export VERSION="${VERSION:-0.0.1}"
 export BUILD_ENV="${BUILD_ENV:-dev}"
 export RELEASE_ID="$VERSION-$(git rev-parse HEAD)_$BUILD_ENV"
-export TAG="$AWS_ECR_REPO/uk.ac.wellcome/$PROJECT:$RELEASE_ID"
+export TAG="$PROJECT:$RELEASE_ID"
```
alexwlchan: We no longer need to store AWS_ECR_REPO as a Travis variable as well

docker/nginx/manage_images.sh
```diff
@@ -9,44 +9,33 @@ TASK=${1:-BUILD}
...
-  # Actually build the image
   docker build --build-arg conf_file="$conf_file" --tag "$TAG" .
 
+  mkdir -p "$ROOT/.releases"
+  echo "$RELEASE_ID" > "$ROOT/.releases/nginx_$variant"
+
   if [[ "$TASK" == "DEPLOY" ]]
   then
-    docker push "$TAG"
+    $ROOT/scripts/deploy_docker_to_aws.py --docker-image="$TAG" --infra-bucket="$CONFIG_BUCKET"
```
alexwlchan: You don't even need this any more. :O

alexwlchan:
Okay, I think this is ready for another review.

alexwlchan:
Actually, this changes a whole bunch of stuff. I might go back and break it up a bit.

scripts/build_nginx_image.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python3
```
kenoir: I might move this into the same folder as the nginx Dockerfile as this is specific to images created with it.

alexwlchan: It’s here initially because it uses some of the tooling in the `scripts` folder. Can be moved, just a little more fiddly.

kenoir: 👍  leave it for now.

alexwlchan:
🙈 

Guess who forgot to update the make task for Travis…

PR #553, opened 2017-07-05
Move SQSWorker to common as Trait

kenoir:

Laying the groundwork for standardising our service workers to use SQS only as an event stream - moves SQSWorker to common and makes it into a Trait.

### Who is this change for?

Developers who want to reduce the types of event stream interface to our services.

- [ ] Deployed new versions


ingestor/src/main/scala/uk/ac/wellcome/platform/ingestor/modules/IngestorWorker.scala
```diff
@@ -0,0 +1,21 @@
+package uk.ac.wellcome.platform.ingestor.modules
+
+import com.amazonaws.services.sqs.model.{Message => AwsSQSMessage}
```
alexwlchan: Do you need this import?

kenoir: No. Will remove.

kenoir:
Part of https://github.com/wellcometrust/platform-api/pull/553

kenoir:
Part of https://github.com/wellcometrust/platform-api/issues/478

PR #559, opened 2017-07-05
Push terraform output to S3

kenoir:

In anticipation of moving build tasks into the `Makefile` it will be handy to have a source of infra variable that aren't dependent on ENV variables being set.

### Who is this change for?

💻

### DIFF
--- a/terraform/apply.sh
+++ b/terraform/apply.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+
+rm -f terraform.outputs.json
+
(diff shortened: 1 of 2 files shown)


kenoir:
I don't think we'll need this because of https://github.com/wellcometrust/platform-api/pull/552.

Rather than rely on terraform outputs we can inspect the AWS APIs!

alexwlchan:
> Rather than rely on terraform outputs we can inspect the AWS APIs!

🙈 

I made the same mistake of parsing terraform output instead of querying AWS recently. 😛

PR #562, opened 2017-07-05
Adds METS ingest bucket, user and policy

kenoir:

Add a bucket to accept METs files from third parties, along with (for now) an IAM user.

### Who is this change for?

_third parties_ who want to send us METS file.

- [x] Run `terraform apply`.


terraform/s3.tf
```diff
@@ -42,3 +42,12 @@ resource "aws_s3_bucket" "dashboard" {
...
 }
+
+resource "aws_s3_bucket" "mets-ingest" {
+  bucket = "mets-ingest"
+  acl    = "private"
+
+  lifecycle {
+    prevent_destroy = true
+  }
+}
```
pollecuttn: run terraform fmt

PR #583, opened 2017-07-05
Move root Dockerfile into /docker

kenoir:

All Dockerfiles should be in the same place

### Who is this change for?

💻


docker/scala_service/Dockerfile
```diff
@@ -1,10 +1,11 @@
 FROM anapsix/alpine-java
 
 ARG project
+ARG target
 
 RUN apk update && apk add python3 && pip3 install awscli && rm -rf /var/cache/apk
 
-ADD $project/target/universal/stage /opt/docker
+ADD $target /opt/docker
```
alexwlchan: 👍

PR #585, opened 2017-07-06
Run python tests in CI

alicefuzier:
Be able to write tests for our lambdas and run them in CI
### Who is this change for?
Everyone

- [ ] Run `terraform apply`.


docker/python3.6_ci/run.sh
```diff
@@ -4,4 +4,16 @@ set -o errexit
 set -o nounset
 set -o verbose
 
-flake8 /data
+export OP="${OP:-plan}"
```
kenoir: should the default be plan?

alicefuzier: oops copy paste mistake

PR #587, opened 2017-07-06
Fix transformer build

alicefuzier:
I _think_ transformer build is broken on master because of an error in our makefile. FIxing it but firste I'm triggering a rebuild of the transformer to make sure

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,7 +69,7 @@ matrix:
         - TASK=sbt-test-api
         - PROJECT=api
     - env:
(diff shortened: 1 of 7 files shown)


alexwlchan:
Oops.

🙈

alexwlchan:
(This is why I shouldn't edit code on my phone!)

pollecuttn:
... during leave! We can see you!

alexwlchan:
I was still in the building when I made that code change yesterday, totes doesn't count

PR #596, opened 2017-07-06
Ensure scala_service can ADD artefact from sub-tree

kenoir:

Fixing the currently broken build

### Who is this change for?

Anyone who wants to deploy our services.


kenoir:
There seems to a problem with the alpine dependency management system right now - hopefully this'll clear up soon.

PR #601, opened 2017-07-07
Fix date serialization issue with sns_utils

kenoir:

Working date serialisation in `sns_utils` in order that the notify_old_deploy lambda functions as expected, giving us the full details of old deploys.

### Who is this change for?

💻 

- [ ] Run `terraform apply`.


terraform/lambdas.tf
```diff
@@ -241,7 +241,7 @@ module "trigger_post_to_slack_dlqs_not_empty" {
 module "trigger_post_to_slack_esg_not_terminating" {
   source               = "./lambda/trigger_sns"
   lambda_function_name = "${module.lambda_post_to_slack.function_name}"
-  lambda_function_arn  = "${module.lambda_post_to_slack.arn}"
+  lambda_function_arn  = "${mo[]dule.lambda_post_to_slack.arn}"
```
alicefuzier: []?

PR #602, opened 2017-07-07
Put Grafana behind an nginx proxy, use make-based build system

alicefuzier:
Expose grafana through nginx and make it use the same template task definition as other services
### Who is this change for?
Devs

- [ ] Run `terraform apply`.


terraform/services/ecs_tasks/templates/default.json.template
```diff
@@ -19,13 +19,10 @@
     "image": "${app_uri}",
     "memory": 2048,
     "name": "app",
-    "environment": [
-      { "name" : "INFRA_BUCKET", "value" : "${infra_bucket}" },
-      { "name" : "CONFIG_KEY", "value" : "${config_key}" }
-    ],
+    "environment": [${environment_vars.*}],
```
alicefuzier: this does not work 😢

Makefile
```diff
@@ -27,6 +27,9 @@ nginx-build-loris: install-docker-build-deps
 nginx-build-services: install-docker-build-deps
 	./scripts/build_nginx_image.py --variant=services
 
+nginx-build-grafana: install-docker-build-deps
```
alexwlchan: You need to add the new tasks to the catchall `nginx-build` and `nginx-deploy` tasks. Then new images will be built and pushed by Travis. \o/

scripts/deploy_docker_to_aws.py
```diff
@@ -63,7 +63,8 @@
     try:
         image = docker_client.images.get(docker_image)
         image.tag(repository=repo_uri, tag=tag)
-        docker_client.images.push(repository=repo_uri, tag=tag)
+        resp = docker_client.images.push(repository=repo_uri, tag=tag)
+        print(resp)
```
alexwlchan: Do we want an extra string here so we know which response we’re looking at?

(Also, if you do `print(type(resp))`, you can see what the type of `resp` is – I think it’s `requests.Response`, in which case you could use `resp.raise_for_status()` again.)

Makefile
alicefuzier: oh right thanks

PR #608, opened 2017-07-10
Make healthchecks check on https port

alicefuzier:
Make health check fail if the service is responding. Currently they check that nginx replies with 301 which might still be true even if the actual application si broken
### Who is this change for?
Everyone

- [ ] Run `terraform apply`.

### DIFF
--- a/terraform/services/ecs_service/alb.tf
+++ b/terraform/services/ecs_service/alb.tf
@@ -8,10 +8,10 @@ resource "aws_alb_target_group" "ecs_service" {
   vpc_id   = "${var.vpc_id}"
 
   health_check {
-    port = 443
+    port     = 443
     protocol = "HTTPS"
-    path    = "${var.healthcheck_path}"
-    matcher = "200"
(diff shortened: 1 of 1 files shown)


alexwlchan:
Hmm. I kinda like having the HTTP check that it 301s.

Could we have two health check blocks, one for HTTP and one for HTTPS? Or a single block that checks both protocols, so if the service is down, it sees enough failures to raise an alarm?

If not, I’d be upset with AWS, but okay to merge this.

alicefuzier:
Nope, one healthcheck per target group and one target group per ecs service... annoying

terraform/services/ecs_service/alb.tf
```diff
@@ -8,8 +8,10 @@ resource "aws_alb_target_group" "ecs_service" {
   vpc_id   = "${var.vpc_id}"
 
   health_check {
+    port = 443
```
alexwlchan: oh, do remember to run `terraform fmt`

(Can you get IntelliJ to do this on save?)

PR #609, opened 2017-07-10
Forward healthcheck requests to our apps and not just nginx

alicefuzier:
Nginx http redirect mean that requests coming from ALB are always redirected meanung that the healthcheck never actually checks if the application is running. Make it forward the request if if comes from the ALB
### Who is this change for?
Everyone

- [x] Run `terraform apply`.

### DIFF
--- a/docker/nginx/api.nginx.conf
+++ b/docker/nginx/api.nginx.conf
@@ -15,6 +15,14 @@ http {
     proxy_next_upstream error;
 
     if ($http_x_forwarded_proto != "https") {
+      set $https_redirect "1";
+    }
+
+    if ($http_user_agent != "ELB-HealthChecker/2.0") {
+      set $https_redirect "${https_redirect}1";
+    }
+
+    if ($https_redirect = "11") {
       rewrite ^(.*)$ https://api.wellcomecollection.org$1 permanent;
     }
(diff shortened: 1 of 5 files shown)


docker/nginx/api.nginx.conf
```diff
@@ -15,6 +15,14 @@ http {
...
 
     if ($http_x_forwarded_proto != "https") {
+      set $https_redirect "1";
+    }
+
+    if ($http_user_agent != "ELB-HealthChecker/2.0") {
+      set $https_redirect "${https_redirect}1";
+    }
+
+    if ($https_redirect = "11") {
```
kenoir: LOGIC!

kenoir: I am guessing you wouldn't do this unless there is no other way.

alicefuzier: I basically copied it from here https://www.digitalocean.com/community/questions/conditional-within-an-if-in-nginx
Apparently nginx does not support multiple conditions in an if statement.
I hate nginx BTW...

jamesgorrie: > I hate nginx BTW...

But it loves you.

PR #610, opened 2017-07-10
Adds better logging to update service list

kenoir:

Better visibility of logs in the update-service-list lambda as it is throwing errors without logs.

### Who is this change for?

💻 

- [ ] Run `terraform apply`.

### DIFF
--- a/lambdas/update_service_list/update_service_list.py
+++ b/lambdas/update_service_list/update_service_list.py
@@ -9,8 +9,10 @@ It does not use the event data from the event.
 
 import datetime
 import json
+import logging
 import os
 import pprint
+import traceback
 
 import boto3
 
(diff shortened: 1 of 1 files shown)


lambdas/update_service_list/update_service_list.py
```diff
@@ -102,12 +104,19 @@ def send_ecs_status_to_s3(ecs_client, s3_client, bucket_name, object_key):
...
-    object_key = os.environ["OBJECT_KEY"]
+        ecs_client = boto3.client('ecs')
+        s3_client = boto3.client('s3')
+
+        bucket_name = os.environ["BUCKET_NAME"]
+        object_key = os.environ["OBJECT_KEY"]
+
+        send_ecs_status_to_s3(ecs_client, s3_client, bucket_name, object_key)
+    except Exception as e:
+        logging.error(traceback.format_exc())
```
alexwlchan: What do you anticipate this giving us that we don’t already get? AFAICT, stdout and stderr are logged to CloudWatch already.

kenoir: I don't really, I want to eliminate doubt. An error is being thrown somewhere and there is nothing in the logs.

alexwlchan: Let’s talk about it tomorrow, because the logging module probably isn’t what you want.

kenoir: Will do.

kenoir:
Closing as error invisibility was caused by the error being a timeout!

PR #612, opened 2017-07-11
Use SQSWorker module in id_minter

kenoir:

A uniform interfaces for services using event streams.

### Who is this change for?

💻 

- [ ] Deployed new versions


kenoir:
Part of https://github.com/wellcometrust/platform-api/issues/478

PR #614, opened 2017-07-11
Add timestamp of when the alarm was trigger to slack notifications

alicefuzier:
Give us more information when troubleshooting slack errors
### Who is this change for?
Devs

- [ ] Run `terraform apply`.

### DIFF
--- a/lambdas/post_to_slack/test_post_to_slack.py
+++ b/lambdas/post_to_slack/test_post_to_slack.py
@@ -32,6 +32,7 @@ def test_post_to_slack(mock_post):
     ]
     reason = "Threshold Crossed: 1 datapoint (4.0) was \
         greater than or equal to the threshold (1.0)."
+    timestamp = "2017-07-10T15:42:24.243+0000"
 
     alarm_info = {
         "AlarmName": alarm_name,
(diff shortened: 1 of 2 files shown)


lambdas/post_to_slack/test_post_to_slack.py
```diff
@@ -0,0 +1,93 @@
...
+                                  'MessageAttributes': {}}}]}
+
+    post_to_slack.main(event, None)
+
+    calls = mock_post.call_args_list
+    print(calls)
+
+    assert len(calls) == 1
+
+    assert calls[0][0][0] == url
```
alicefuzier: WTF??

kenoir: 😂

kenoir:
`flake8` is probably upset about whitespace ...

lambdas/post_to_slack/test_post_to_slack.py
```diff
@@ -0,0 +1,118 @@
...
+        }]
+    }
+
+    post_to_slack.main(event, None)
+
+    calls = mock_post.call_args_list
+
+    assert len(calls) == 1
+
+    assert calls[0][0][0] == url
```
alicefuzier: @alexwlchan is there a better way of doing this?

PR #615, opened 2017-07-11
Adds neo-sbt-scalafmt plugin

kenoir:

Adds https://github.com/lucidsoftware/neo-sbt-scalafmt plugin (Thanks Lucidchart!) in order that scalafmt is easy to run from within sbt.

### Who is this change for?

💻


alicefuzier:
Will this make the build fail if the code is not formatted? Or will it automatically format it when running a build? or neither of those?

PR #629, opened 2017-07-11
Run Scala tests, lint, and build in Docker

kenoir:

Run our Scala tests & build in a Docker container, in order to make tests & builds more portable and uniform with the other build steps.

This is complicated by the fact our tests use docker-compose to set up the correct environment for end-to-end tests.

We work our way around this by expecting our host to be running the docker daemon and making its socket available in the container.

### Who is this change for?

Developers!


kenoir:
I've seen intermittent issues with the `docker-compose` set up, in addition i've not yet ported our build script to the container.

PR #630, opened 2017-07-12
Should rebuild if files in the docker folder change

kenoir:

More sensible rebuilds in CI in order that we are notified of breaking changes, while reducing build time in CI as far as possible.

### Who is this change for?

Developers who want good test coverage and speedy build.

### DIFF
--- a/scripts/should_rebuild_master.py
+++ b/scripts/should_rebuild_master.py
@@ -21,12 +21,8 @@ def should_rebuild_project(changed_files, project):
     if should_retest_project(changed_files, project):
         return True
 
-    elif ('Dockerfile' in changed_files) or ('.dockerignore' in changed_files):
-        print("*** Changes to Dockerfile mean we should rebuild")
-        return True
-
-    elif 'run.sh' in changed_files:
-        print("*** Changes to run.sh mean we should rebuild")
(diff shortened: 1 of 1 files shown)


kenoir:
This will catch some unnecessary changes (changes to the loris Dockerfile for example) but now as much of the build relies on the images generated by the `Dockerfile`s in that folder we should rebuild on changes.

scripts/should_rebuild_master.py
```diff
@@ -21,12 +21,8 @@ def should_rebuild_project(changed_files, project):
     if should_retest_project(changed_files, project):
         return True
 
-    elif ('Dockerfile' in changed_files) or ('.dockerignore' in changed_files):
```
alexwlchan: 1. Did you mean to throw out the `.dockerignore`?
2. This isn't valid Python syntax (unmatched left parens), so this script will always exit with code 1 and a SyntaxError – which defaults to always rebuilding.

kenoir: 👍 

1. Yes
2. Obviously (shouldn't this fail something in CI).

alexwlchan: We only run flake8 over lambdas, could do it for scripts as well

PR #631, opened 2017-07-12
Add a script for installing Lambda dependencies

alexwlchan:

Allow us to install third-party Python dependencies in our Lambdas. Currently we get boto3 installed for free, but that’s about it.

### Who is this change for?

Developers who want access to a richer Python ecosystem.

structlog is there to allow us to start doing better logs in Lambdas (which will be a separate patch) – which in turn will hopefully reduce log spam.

- [ ] Run `terraform apply`.


alexwlchan:
@kenoir I think this is ready for review.

Makefile
```diff
@@ -163,6 +163,7 @@ sbt-deploy: \
 
 ## Run a plan
 terraform-plan: docker-build-terraform
+	./lambdas/install_lambda_deps.sh
```
kenoir: Can we do this as an operation inside the python3.6_ci container?

PR #634, opened 2017-07-12
Use the description for the Miro label, not a truncated version

### LINKED ISSUE #616: Map description to label for Miro V images
For Miro V images only, we should map the description to label, not the title to label.

This is because for the V images, the title is just a chopped copy of the description.

The existing mapping should remain in place for all other Miro collections.

alexwlchan:

Metadata for Miro records won’t have weird amounts of duplication for V images.

Resolves #616.

### Who is this change for?

End-users who want a less repetitive API. End-users who want a less repetitive API.

- [ ] Deployed new versions

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/models/MiroTransformable.scala
+++ b/common/src/main/scala/uk/ac/wellcome/models/MiroTransformable.scala
@@ -54,11 +54,20 @@ case class MiroTransformable(MiroID: String,
 
       // <image_title>: the Short Description.  This maps to our property
       // "label".
-      val label = miroData.title.get
+      //
+      // In the V collection, this is just a truncated form of the desc field,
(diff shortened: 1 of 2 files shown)


common/src/main/scala/uk/ac/wellcome/models/MiroTransformable.scala
```diff
@@ -54,11 +54,20 @@ case class MiroTransformable(MiroID: String,
 
       // <image_title>: the Short Description.  This maps to our property
       // "label".
-      val label = miroData.title.get
+      //
+      // In the V collection, this is just a truncated form of the desc field,
+      // so we pass the full string as the label and omit a description.
```
kenoir: Worth mentioning that `.get` on an option her will lead to an exception if empty - this is expected here but might be useful to add a comment describing why.

PR #637, opened 2017-07-12
Makes the dashboard work accross accounts

kenoir:

Displaying all infrastructure on the same dashboard!

### Who is this change for?

The experience and platform teams

- [X] Run `terraform apply`.

untitled


kenoir:
We currently do not trigger on ECS service changes from the experience team so this PR adds a trigger to refresh the dashboard periodically (every 5 minutes) in order to keep experience deployments up to date. @jamesgorrie it might be worth doing the work to integrate the event stream if you want "right now" updates.

PR #638, opened 2017-07-12
Fix lint task failure because of third party libs

kenoir:

Linting Python that doesn't include 3rd party libraries and consequently error incorrectly.

### Who is this change for?

💻


kenoir:
Worth noting we didn't see any failures in CI because of this - we should investigate.

PR #652, opened 2017-07-17
Add terminalFailureHook to TryBackoff

kenoir:

Extending objects may want a means to alert on terminal failure of a TryBackoff - this PR adds an overridable hook that can be used to execute some arbitrary side-effect on terminal failure.

### Who is this change for?

💻 

- [ ] Deployed new versions

### DIFF
--- a/common/src/test/scala/uk/ac/wellcome/utils/TryBackoffTest.scala
+++ b/common/src/test/scala/uk/ac/wellcome/utils/TryBackoffTest.scala
@@ -16,7 +16,13 @@ class TryBackoffTest
   val system: ActorSystem = ActorSystem.create("TryBackoffTestActorSystem")
 
   val tryBackoff = new TryBackoff {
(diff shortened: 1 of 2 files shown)


common/src/test/scala/uk/ac/wellcome/utils/TryBackoffTest.scala
```diff
@@ -58,6 +64,23 @@ class TryBackoffTest
...
+    val f = new TryBackoffHelper()
+
+    tryBackoff.wasCalled = false
+    tryBackoff.run(f.alwaysFails, system)
+
+    Thread.sleep(10000)
+    val finalLength = f.calls.length
+    Thread.sleep(5000)
+
+    f.calls.length shouldBe finalLength
```
alexwlchan: As discussed, let’s cut this middle bit – so this test only fails if the terminal failure hook is never called.

PR #664, opened 2017-07-18
Trigger lambdas that access ECS API but don't use event every minute

kenoir:
In order to reduce ECS API requests and avoid being rate limited.

- [ ] Run `terraform apply`.


kenoir:
@alexwlchan indeed - we'll clean up if this works.

PR #666, opened 2017-07-18
Prevent EC2 tagging rate limit in tagger lambda

kenoir:

Currently the EC2 instance tagger tends to hit the rate limit - this tries to tag each instance only once by recording the instance id in S3.

### Who is this change for?

💻 

- [ ] Run `terraform apply`.

### DIFF
--- a/lambdas/ecs_ec2_instance_tagger/ecs_ec2_instance_tagger.py
+++ b/lambdas/ecs_ec2_instance_tagger/ecs_ec2_instance_tagger.py
@@ -20,6 +20,7 @@ Requires a Cloudwatch Event with pattern:
 """
 
 import json
+import os
 
 import boto3
 from botocore.exceptions import ClientError
@@ -37,8 +38,10 @@ def create_tags(ec2_client, ec2_instance_id, event_detail):
     cluster_arn = event_detail['clusterArn']
     container_instance_arn = event_detail['containerInstanceArn']
 
-    print(f'Tagging {ec2_instance_id} clusterArn: {cluster_arn}')
-    print(f'Tagging {ec2_instance_id} containerInstanceArn: {container_instance_arn}')
+    print(f'Tag {ec2_instance_id} clusterArn: {cluster_arn}')
+    print(
+        f'Tag {ec2_instance_id} containerInstanceArn: {container_instance_arn}'
+    )
 
     return ec2_client.create_tags(
         Resources=[ec2_instance_id],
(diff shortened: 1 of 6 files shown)


kenoir:
Needs tests!

terraform/iam_policy_document.tf
```diff
@@ -236,6 +236,20 @@ data "aws_iam_policy_document" "s3_put_dashboard_status" {
...
+data "aws_iam_policy_document" "s3_put_infra_tmp" {
+  statement {
+    actions = [
+      "s3:PutObject",
+      "s3:GetObjectACL",
+      "s3:PutObjectACL",
+    ]
+
+    resources = [
+      "${aws_s3_bucket.infra.arn}/tmp/*",
```
alexwlchan: If this is temporary, when (if ever) will it be purged?

lambdas/ecs_ec2_instance_tagger/ecs_ec2_instance_tagger.py
```diff
@@ -46,8 +59,35 @@ def main(event, _):
...
+            print(f'{ec2_instance_id} not seen yet, tagging!')
+
+            response = create_tags(
+                ec2_client,
+                ec2_instance_id,
+                event['detail'])
+
+            print(f'response = {response!r}')
+        else:
+            raise ex
```
alexwlchan: minor: if you’re in the body of an `except:` in Python, you can just call `raise` and it automatically re-raises the exception.

lambdas/ecs_ec2_instance_tagger/ecs_ec2_instance_tagger.py
```diff
@@ -46,8 +59,35 @@ def main(event, _):
...
+
+    bucket_name = os.environ["BUCKET_NAME"]
+    object_path = os.environ["OBJECT_PATH"]
 
     ec2_instance_id = event['detail']['ec2InstanceId']
 
-    response = create_tags(ec2_client, ec2_instance_id, event['detail'])
-    print(f'response = {response!r}')
+    try:
+        s3_client.get_object(
```
alexwlchan: We could use `head_object()` here to be slightly more efficient – we’re only interested in checking whether it exists, not what the file contains.

```pycon
>>> import boto3
>>> s3 = boto3.client('s3')
>>> s3.head_object(Bucket='platform-infra', Key='tmp/doesnotexist')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/chana/.virtualenvs/api/lib/python3.6/site-packages/botocore/client.py", line 310, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/chana/.virtualenvs/api/lib/python3.6/site-packages/botocore/client.py", line 599, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (404) when calling the HeadObject operation: Not Found
```

Although note you’ll need to tweak the exception handling – specifically, I think you’d want

```python
if ex.response['Error']['Code'] == '404':
    ...
```

kenoir: I used get_object because I couldn't get the exception handling to work for head object :) I'll try this way.

terraform/iam_policy_document.tf
kenoir: I've added a lifecycle rule to the bucket to expire objects after 30 days.

PR #672, opened 2017-07-19
Transformer config to use sqs

alicefuzier:
Make the transformer start
### Who is this change for?
Everyone

- [ ] Run `terraform apply`.

### DIFF
--- a/terraform/iam_role_policy.tf
+++ b/terraform/iam_role_policy.tf
@@ -28,22 +28,16 @@ resource "aws_iam_role_policy" "ecs_calm_adapter_service_scheduler_sns" {
 
 # Role policies for the transformer
 
-resource "aws_iam_role_policy" "ecs_transformer_task_dynamo" {
-  name   = "ecs_task_transformer_task_dynamo_policy"
+resource "aws_iam_role_policy" "ecs_transformer_task_read_transformer_q" {
+  name   = "ecs_transformer_task_read_transformer_q"
   role   = "${module.ecs_transformer_iam.task_role_name}"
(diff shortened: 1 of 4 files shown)


terraform/iam_role_policy.tf
```diff
@@ -40,12 +40,6 @@ resource "aws_iam_role_policy" "ecs_transformer_task_sns" {
   policy = "${module.id_minter_topic.publish_policy}"
 }
 
-resource "aws_iam_role_policy" "ecs_transformer_task_kinesis_stream" {
-  name   = "ecs_task_kinesis_stream_policy"
-  role   = "${module.ecs_transformer_iam.task_role_name}"
-  policy = "${data.aws_iam_policy_document.read_calm_kinesis_stream.json}"
```
alexwlchan: Do we still need this policy document, or can we delete it now?

terraform/iam_role_policy.tf
```diff
@@ -28,10 +28,10 @@ resource "aws_iam_role_policy" "ecs_calm_adapter_service_scheduler_sns" {
 
 # Role policies for the transformer
 
-resource "aws_iam_role_policy" "ecs_transformer_task_dynamo" {
-  name   = "ecs_task_transformer_task_dynamo_policy"
+resource "aws_iam_role_policy" "ecs_transformer_task_read_transformer_q" {
+  name   = "ecs_transformer_task_read_transformer_q"
   role   = "${module.ecs_transformer_iam.task_role_name}"
-  policy = "${data.aws_iam_policy_document.allow_dynamodb_all.json}"
```
alexwlchan: And this one.

PR #676, opened 2017-07-19
Add aws_cloudfront_distribution for Loris

kenoir:

Serving images via CloudFront will reduce load on the image server and speed up serving images.

### Who is this change for?

Folk who want to see images, quick!

- [ ] Run `terraform apply`.

### DIFF
--- a/terraform/cloudfront.tf
+++ b/terraform/cloudfront.tf
@@ -4,15 +4,15 @@ resource "aws_cloudfront_distribution" "loris" {
     origin_id   = "loris"
 
     custom_origin_config {
-      https_port = 443
-      http_port = 80
+      https_port             = 443
+      http_port              = 80
       origin_protocol_policy = "match-viewer"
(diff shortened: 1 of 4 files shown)


kenoir:
Part of https://github.com/wellcometrust/platform-api/issues/653

terraform/cloudfront.tf
```diff
@@ -0,0 +1,52 @@
+resource "aws_cloudfront_distribution" "loris" {
+  origin {
+    domain_name = "${module.api_alb.dns_name}"
+    origin_id   = "loris"
+
+    custom_origin_config {
+      https_port = 443
+      http_port = 80
+      origin_protocol_policy = "match-viewer"
+      origin_ssl_protocols = ["TLSv1.2"]
```
alexwlchan: The format job should be whinging about this, but it doesn't notice changes in the `terraform` directory. Oops.

kenoir: formatted :)

PR #688, opened 2017-07-20
Removes calm_adapter

kenoir:
We're not using it right now, and will likely rewrite when we do. Cleaning up for now.


alexwlchan:
Remove the Travis job?

PR #689, opened 2017-07-20
Transformer should not fail for transform errors

kenoir:

At present errors in transforming data result in exceptions bubbling up to the trybackoff and resulting in a terminal failure.

### Who is this change for?

💻 

- [ ] Run `terraform apply`.


kenoir:
Fixes https://github.com/wellcometrust/platform-api/issues/675

transformer/src/main/scala/uk/ac/wellcome/transformer/receive/SQSMessageReceiver.scala
```diff
@@ -31,8 +32,11 @@ class SQSMessageReceiver(
         triedWork match {
           case Success(work) =>
             publishMessage(work)
+          case Failure(SQSReaderGracefulException(e)) =>
+            info("Recoverable failure extracting workfrom record", e)
```
alexwlchan: nit: /s/workfrom/work from

PR #691, opened 2017-07-21
Add a scale up/down schedule for our backend ASGs

alexwlchan:

Turn off our monitoring and services cluster outside working hours, when they’re not in use.

### Who is this change for?

Our bank account – this saves us a little over 1½ months of t2.large running time a month, so I reckon these forty lines save ~$120/month. 🤑 

- [x] Run `terraform apply`.


terraform/asg_schedule.tf
```diff
@@ -0,0 +1,43 @@
...
+  scheduled_action_name  = "Turn off the services cluster outside working hours"
+  desired_capacity       = 0
+  max_size               = 0
+  recurrence             = "0 19 * * 1-5"
+  autoscaling_group_name = "${module.services_cluster_asg.asg_name}"
+}
+
+resource "aws_autoscaling_schedule" "increase_services" {
+  scheduled_action_name  = "Turn on the services cluster during working hours"
+  desired_capacity       = 2
```
kenoir: Should these be a var to match that in `asg.tf`?

alexwlchan: Nice suggestion, done.

Issue #694, opened 2017-07-21
Only use the first line of the description as a label (if at all)


alexwlchan:
In this title:

```
b'An angel leading a soul into hell. Oil painting by a followe'
b'An angel leading a soul into hell. Oil painting by a follower of Hieronymus Bosch.\r\n\r\nIn this painting, a naked man arrives in hell escorted by a winged angel clad in white. The naked man is presumed to be a soul of the dead. Monsters inflict horrendous tortures on the damned who are already there. The angel who accompanies the soul of the deceased to the underworld is called the \'psychopomp\'. In ancient Graeco-Roman religion the psychopomp is Hermes (Mercury), but Christian doctrines are more vague on how the soul actually reaches its post-mortem destination: in some versions the event itself happens immediately after the death of an individual, while in others it happens after Judgment Day. \r\r\n\r\r\nPaintings, poems and songs have contributed as much to the idea of hell as official doctrines. Hell is depicted in mediaeval and early modern churches and cemeteries (as at Pisa) as well as on Buddhist paintings of the realm of Yama. The torments of the damned reflect the tortures inflicted by torturers acting for political masters and judicial authorities in the secular dungeons of this world. \r\r\n\r\r\nOne master painter who specialized in the genre of hell-scenes was Hieronymus Bosch (died 1516). His enormously popular pictures were emulated by many artists throughout the 16th century, in Italy as well as in the Low Countries. The present painting was at one time in Rome and may have been painted by an Italian painter. The rotunda in the right middle-ground is similar to Virgil\'s tomb at Piedigrotta, near Naples - appropriately, as Virgil described Aeneas\' descent to the underworld in the "Aeneid", and acted as Dante\'s guide to the underworld in Dante\'s "Inferno".'
```

We can probably do a better job if we guess everything up to the first newline as a candidate title.

alexwlchan:
We need to run a reindex before we can close this.

alexwlchan:
Hmm, this doesn’t seem to have worked: https://api.wellcomecollection.org/catalogue/v0/works/c3c7b5pa

I wonder – did something go wrong with the reindex?

alexwlchan:
So that item arrived at the ingestor, but the transformer data it received was wrong:

```
[...]

PR #697, opened 2017-07-24
Only use the first line of the description as a label (if at all)

alexwlchan:

Make our labels be a bit less ridiculously long. New heuristic: only use the first line of the description (if any) for the new label, keep the rest in the description field.

Related: #694.

### Who is this change for?

UX team, who want not-incredibly-long titles. API users who want better organisation of information.

- [ ] Deployed new versions
- [ ] Run `terraform apply`


alexwlchan:
Moving this back to WIP for now – I want to rethink how I do this, it’s going to get more complicated with #699.

PR #707, opened 2017-07-25
Tweak cluster compute resource

kenoir:

While keeping costs the same, trim some unused compute resource from the clusters and add it to Loris.

### Who is this change for?

💻 

- [x] Run `terraform apply`.


alexwlchan:
✨💰✨

kenoir:
Parwt of https://github.com/wellcometrust/platform-api/issues/537

PR #709, opened 2017-07-26
[WIP] Adds a container to watch for spot instance termination messages

kenoir:

In order to use spot instances it will be useful to have a way of detecting spot instance termination notices and reacting to them elsewhere.

### Who is this change for?

💻


kenoir:
This needs to run on every instance, so this needs to be rethought. It might be possible to add it to the coreos cloud-config - in which case the host instance will need permissions to push to the topic.

kenoir:
Closing for now, as spot instances aren't really a concern.

PR #710, opened 2017-07-26
Adds optional EFS performance mode to EFS module

kenoir:

Adds optional EFS performance mode to EFS module in order to utilise faster performance on EFS where necessary.

### Who is this change for?

💻


kenoir:

PR #713, opened 2017-07-27
Upload results from a Gatling test to S3

alexwlchan:
Put our Gatling tests in a central location. Easier sharing and viewing. Longer term, this is a useful step for running Gatling in ECS.


kenoir:

PR #714, opened 2017-07-27
Gatling as ECS Script Task

kenoir:

This PR adds the necessary changes to run Gatling in ECS as a script task in order to run load tests on a regular basis.


kenoir:

PR #715, opened 2017-07-27
Deploy gatling to ECR

kenoir:

Allows gatling container images to be deployed to ECR so they can later be run as tasks.


kenoir:

PR #716, opened 2017-07-27
Adds Gatling script task

kenoir:

We want to be able to run load tests on a regular basis inside ECS. This adds a script task to do that.


kenoir:

PR #720, opened 2017-07-27
Scale down API cluster as capacity not required

kenoir:

A reasonable provision for the API cluster.

- [ ] Run `terraform apply`.

### DIFF
--- a/terraform/asg.tf
+++ b/terraform/asg.tf
@@ -44,10 +44,10 @@ module "api_cluster_asg" {
   user_data             = "${module.api_userdata.rendered}"
   vpc_id                = "${module.vpc_api.vpc_id}"
 
-  asg_desired = "4"
-  asg_max     = "8"
+  asg_desired = "3"
+  asg_max     = "6"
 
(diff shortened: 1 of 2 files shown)


alexwlchan:
Better – but now we’re wasting a lot of capacity.

CPU:

```
a pair of c4.xlarge     2 * (1024 * 4)   = 8192

API tasks               2 * (1024 + 128) = 2304
Loris tasks             2 * (1024 + 128) = 2304
-----------------------------------------------
Spare                  8192 - (2304 * 2) = 3584
```

Memory:

```
a pair of c4.xlarge     2 * (1024 * 8)   = 16384

API tasks               2 * (2048 + 128) =  4352
Loris tasks             2 * (2048 + 128) =  4352
------------------------------------------------
Spare                 16384 - (4352 * 2) =  7680
```

tl;dr: If we’re paying for c4.xlarges, we can afford to run another API and another Loris task without paying an extra cent – deploys might be a bit slower, but the majority of the time is spent not-deploying.

PR #722, opened 2017-07-28
Tweak the way we allocate our EC2 provision

alexwlchan:

Rearrange the way we allocate tasks for the API and Loris in a way that saves money and gets better performance. Specifically:

* Reduce the CPU/memory allocation of the tasks so they take up a discrete amount of resource (1&nbsp;vCPU, 2&nbsp;GB of memory)
* Switch to using t2.mediums instead of c4.xlarges in the API cluster – better value for money, and allows us to fully pack tasks in the cluster
* Increase from two to three instances of the Loris and API tasks, so overall we should get better performance

This saves us ~$50/month on our long-term EC2 costs (when we were using a pair of t2.larges), and ~$180/month on the experimental plan of putting everything on a pair of c4.xlarges.

### Who is this change for?

💰 🚤 💰

- [ ] Run `terraform apply`.


kenoir:

PR #723, opened 2017-07-28
JSON example: CALM object from hierarchy

ChristopherHilton:
A low-level object from a CALM hierarchy - includes "isPartOf" and a link to digitised facsimile
### Who is this change for?
Silver Oliver and general Platform team

### DIFF
--- a/ontologies/CALM_from_hierarchy.json
+++ b/ontologies/CALM_from_hierarchy.json
@@ -1,0 +1,85 @@
+{
+  "@context": "https://api.wellcomecollection.org/catalogue/v0/context.json",
+  "id": "n9gfw78",
+  "type": "Work",
+  "identifiers": 
+    [
+      {
+        "value": "PPCRI/D/1/1/8",
+        "type": "Identifier",
+        "identifierScheme": "CALM-RefNo"
+      },
+      {
+        "value": "PP/CRI/D/1/1/8",
+        "type": "Identifier",
+        "identifierScheme": "CALM-AltRefNo"
+      }
+    ],
+	"isPartOf":
+      {
+        "value": "PPCRI/D/1/1",
+        "type": "Identifier",
+        "identifierScheme": "CALM-RefNo"
+      },
+  "title": "Correspondence: H",
+  "creationDate": 
+    {
+      "label": "1953-1970",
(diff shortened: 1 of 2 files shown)


ontologies/CALM_from_hierarchy.json
```diff
@@ -0,0 +1,85 @@
...
+  "creationDate": 
+    {
+      "label": "1953-1970",
+      "type": "Period"
+    },
+  "description": "<p>The correspondents are: <p>Dr David Harker (1953), Dr R A Hickling, Sir Harold P Himsworth (permission for Olby \"to see the report prepared for the Biophysics Committee to assist him with the historical account he is preparing...\"), Professor Mahlon Hoagland, Dorothy Hodgkin, Dr Robert W Holley, Dr R Holliday, Christopher Holme (BBC, \"The Life Code\"), Dr K Holmes (Paradies), Dr Norman J Holter, T A Hood (Central Office of Information, the counter current distribution machine), Dr David H Hubel, J H Humphrey, Dr John A Hunt, Dr Ing. Fritz Haas, Dr Lowell P Hager, Dr J B S Haldane, Dr James Hall (Polanyi), Dr L D Hamilton (<i>The Double Helix</i>), Dr Edda Hanington, Dr C Hansch (Pomona monograph), Miss Barabara Hanson, Miss Donna Haraway, [   ] Harris, Professor Henry Harris (Ken Burton), C M H Harrisson, W. Harrison (Leeds University), Dr Philip E Hartman, Dr Robert Haselkorn (<i>The Bulletin of the Atomic Scientists</i>, ochre suppressors), Dr K A Hassall, Professor Felix Haurowitz, R S Haworth (Crick family tree enquiry), Professor W Hayes, and [   ] Huxley.   ",
```
pollecuttn: I think I've said this on another json example, but do we want html tags in the json?

ChristopherHilton: Good point.  Leaving them there pro tem just to illustrate how it comes out of CALM, but yes....  In some cases we will be dealing with huge slabs of text that only really make sense to the reader if laid out and formatted in some way, so if the HTML were to be stripped out what would our other options be?

PR #726, opened 2017-07-28
Adds notification topic for load test results

kenoir:

We wish to pass on metrics from a load test, so do that!

- [ ] Run `terraform apply`.


kenoir:

PR #727, opened 2017-07-28
Post Gatling results to CloudWatch

kenoir:

Adds a lambda that accepts "assertions.json" as posted to an SNS topic,
and extracts the assertions and puts their values as CloudWatch metrics.

This is in order that we can effectively track load test results over
time.

### Who is this change for?

API consumers who want results fast!

- [ ] Run `terraform apply`.


kenoir:

PR #729, opened 2017-07-28
Fix matrix configuration

BanzaiMan:
This creates 13 jobs in the matrix

Fix Travis CI configuration

### Who is this change for?

You.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,22 +71,6 @@ cache:
     - "transformer/target"
     - "/home/travis/docker/"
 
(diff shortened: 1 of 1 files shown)


alexwlchan:
Ah, thanks! Clearly I wasn’t paying attention when I rearranged the config, oops.

Other Wellcome folks: one hint we might have spotted that the config was wrong was in the Requests tab, here: https://travis-ci.org/wellcometrust/platform-api/requests

PR #730, opened 2017-07-31
Various Loris improvements

alexwlchan:
This PR is stuff I’ve tried to improve our Loris performance that’s had a noticeable improvement in Gatling tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -111,3 +111,4 @@ env:
     - TASK=nginx-build
     - TASK=cache_cleaner-build
     - TASK=gatling-build
+    - TASK=loris-build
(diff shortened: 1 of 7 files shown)


alexwlchan:
Briefly, the difference with and without threading:

napkin 31-07-17 2 50 45 pm

alexwlchan:
Here’s what changing the process count did to my test:

napkin 31-07-17 3 34 53 pm

alexwlchan:
This is the difference between Alpine and Ubuntu:

napkin 31-07-17 4 02 57 pm

Note that this isn’t just the effect of a warmer cache – I did Alpine|Ubuntu|Alpine runs, and Ubuntu beat both instances of Alpine. Not sure why.

alexwlchan:
Here’s a zany idea: what if we cut out Loris and just proxy straight from S3?

PR #735, opened 2017-08-01
Adds simple load test for digital experience site

kenoir:

Simple load test for the digital experience image search, in order to benchmark the image API.

### Who is this change for?

💻 

- [ ] Run `terraform apply`.


kenoir:
Worth noting that Gatling is not pretending to be a browser very effectively so will not run any JS - though it should download image resources specified in the page (though notably not background images). 

I think that'll probably suit our purposes for now.

PR #739, opened 2017-08-01
Use ARN rather than lookup from name in Gatling container

kenoir:
In order that we don't need to give listTopics permission to tasks.

### Who is this change for?
💻 

- [ ] Run `terraform apply`.


docker/gatling/notify.sh
```diff
@@ -4,16 +4,14 @@ set -o errexit
 set -o nounset
 set -o verbose
 
-TOPIC_NAME=$1
+TOPIC_ARN=$1
```
kenoir: This does diverge this file from the `notify.sh` but I think it's ok so far (if we reuse this again then let's rethink).

Issue #505, opened 2017-07-03
Embed image-level metadata into converted C images


jtweed:

alexwlchan:
This we do need to use AWS Batch for.

PR #764, opened 2017-08-08
Start writing the batch job definition

alexwlchan:
Write a script that downloads a TIF from S3, embeds some metadata, and sticks it back in S3. Optionally deletes the original.

Related: #505.


docker/tif-metadata/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM alpine
+
+RUN apk update
+RUN apk add bash python3
+RUN pip3 install boto3 docopt awscli
```
kenoir: Suggest you move this into a requirements.txt where we can pin the version, then:

`pip3 install -r requirements.txt`

docker/tif-metadata/run.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python3
```
kenoir: Perhaps move this whole thing into an `app` directory alongside `requirements.txt`.

docker/tif-metadata/run.py
```diff
@@ -0,0 +1,98 @@
...
+                     Filename=job_filename)
+
+with open(job_filename, "r") as job_file:
+    job = json.load(job_file)
+
+tasks = job['task_list']
+print(f"Starting download of {len(tasks)} images from s3")
+processes = []
+for task in tasks:
+    _, tmp_fp = tempfile.mkstemp()
```
kenoir: You could move this into a function and then use a list comprehension to make it clearer what's going on.

i.e. 

```py
def download_source(source, processes = []):
   _, tmp_fp = tempfile.mkstemp()
  s3_source = task["source"]
  p = subprocess.Popen(["aws", "s3", "cp", s3_source, tmp_fp])
  processes.append((p, task, tmp_fp))

[download_source(task["source"], processes) for task in tasks]
```

docker/tif-metadata/run.py
```diff
@@ -0,0 +1,98 @@
...
+    s3_source = task["source"]
+    p = subprocess.Popen(["aws", "s3", "cp", s3_source, tmp_fp])
+    processes.append((p, task, tmp_fp))
+
+for process, task, _ in processes:
```
kenoir: Same suggestion as above about moving out logic and using a function name to document what's going on.

.travis.yml
```diff
@@ -112,6 +112,6 @@ env:
     - TASK=sbt-test-reindexer
     - TASK=test-lambdas
     - TASK=nginx-build
-    - TASK=cache_cleaner-build
+    - TASK=tif-metadata-build
```
alicefuzier: why does this seem to be replacing the cache cleaner?

kenoir: It is. We no longer use the cache cleaner AFAIK.

PR #776, opened 2017-08-09
Switch to using identifierScheme in the identifiers

### LINKED ISSUE #334: Normalise identifiers used for sources
Names of fields containing primary identifiers should mirror the domain from which they come (e.g. MiroId -> image_number). 

- Should this normalisation happen in the source tables or the transformation step? 
- Should source tables contain common names for hash/range keys to help internal modelling later on?

### LINKED ISSUE #338: Decide the format for identifiers in the API responses
Related:

* #306, which tracked exposing them at all
* #334, which tracks how we store them internally

alexwlchan:

Move our identifiers to use the model defined in the latest examples:

```json
{
    "type": "Identifier",
    "identifierScheme": "miro-image-number",
    "value": "V0000001"
}
```

I went with `miro-image-number` as the string, because I couldn’t think of anything better.

We need to be careful about promoting this to prod – it might break the Collection site’s image lookup (although I think they just use the `value` field, so maybe not?).

Fixes #334, fixes #338.

### Who is this change for?

A Platform team who want an up-to-date API.

- [x] Reviewed by @jtweed
- [x] Updated the JSON-LD context – not required
- [ ] Deployed new versions

### DIFF
--- a/api/src/main/scala/uk/ac/wellcome/platform/api/models/DisplayWork.scala
+++ b/api/src/main/scala/uk/ac/wellcome/platform/api/models/DisplayWork.scala
@@ -60,7 +60,7 @@ case class DisplayIdentifier(source: String, name: String, value: String) {
 
 object DisplayIdentifier {
   def apply(sourceIdentifier: SourceIdentifier): DisplayIdentifier =
-    DisplayIdentifier(source = sourceIdentifier.source,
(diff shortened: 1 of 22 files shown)


alexwlchan:
Once this is merged, I’m going to add a patch which puts all the identifier strings in one place – too much of this patch has been spent playing whack-a-mole with changed strings.

PR #781, opened 2017-08-10
API model docs

jtweed:

Put in place initial expanded API docs.

Note that another pull request will follow with the final text. This is just so we can see what's in the ontology docs as they stand.

You can see the output of this here:

https://wellcomecollection-catalogue.api-docs.io/v0/works/works

### Who is this change for?

Developers using our Swagger docs.


api/src/main/scala/uk/ac/wellcome/platform/api/models/DisplayWork.scala
```diff
@@ -53,9 +59,14 @@ case object DisplayWork {
   }
 }
 
-case class DisplayIdentifier(identifierScheme: String, value: String) {
-  @JsonProperty("type")
-  val ontologyType: String = "Identifier"
+@ApiModel(
```
kenoir: Love this. Documentation in code as documentation!

api/src/main/scala/uk/ac/wellcome/platform/api/models/DisplayResultList.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import com.sksamuel.elastic4s.http.search.SearchResponse
+import io.swagger.annotations.{ApiModel, ApiModelProperty}
+
+@ApiModel(
+  value="ResultList",
+  description="A list of things."
+)
+case class DisplayResultList(
```
kenoir: Nice change.

kenoir:
Travis is upset because you need to run `make format` to ensure the Scala code is formatted in a way that makes our build happy.

kenoir:
Also as this changes the API output slightly is causes some of the API tests to fail. The tests will need updating to match the expected output.

api/src/main/scala/uk/ac/wellcome/platform/api/models/DisplayResultList.scala
jtweed: A nice change that didn't entirely work! Fixing now.

PR #784, opened 2017-08-11
[WIP] Add a task for deleting unused Elasticsearch indices

### LINKED ISSUE #774: Write a script to delete old Elasticsearch indices
We can see which indices we're using by inspecting `terraform.tfvars`.

We can see which indices Elasticsearch has by inspecting the `_cat/indices` endpoint.

Old indices will eat storage and memory; we should have a way to trigger a manual cleanup.

alexwlchan:

Avoids wasting space on our Elasticsearch cluster. I’m trying a new directory layout, what do people think?

Resolves #774.

### Who is this change for?

A Platform team who want an efficient ES cluster.


alexwlchan:
@kenoir We were discussing rearranging the directory structure a bit – how do you feel about the approach in this PR?

tasks/elasticsearch_cleaner/Makefile
```diff
@@ -0,0 +1,7 @@
+build:
+	docker build --tag elasticsearch_cleaner .
+
+deploy: build
+	echo $$(git rev-parse HEAD) > $$(git rev-parse --show-toplevel)/.releases/elasticsearch_cleaner
+	docker tag elasticsearch_cleaner elasticsearch_cleaner:$$(git rev-parse HEAD)
+	$$(git rev-parse --show-toplevel)/scripts/deploy_docker_to_aws.py --project=elasticsearch_cleaner --infra-bucket=platform-infra
```
alexwlchan: This is a bit messy because our scripts assume you’re building from within the `docker` directory, but I think `tasks` is semantically neater. Thoughts?

kenoir:
I like this - but I want to be absolutely sure we're never going to delete indexes that are in use by accident. 

I'd like some further reassurance like the index hasn't been updated in two weeks, or it's on a whitelist. I'm not sure that just relying on what's in the TF is enough.

alexwlchan:
> I'd like some further reassurance like the index hasn't been updated in two weeks, or it's on a whitelist. I'm not sure that just relying on what's in the TF is enough.

I am 👍 on getting more reassurance, although I think last updated time isn’t a good metric to use. I expect the prod indexes to become more stable as time goes on – if anything, it’s the ones that are churning and being discarded quickly that are less useful.

The stats API in Elasticsearch 5.5 might contain something useful here, maybe? https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-stats.html

In the meantime, moving back to WIP until we decide what to do.

alexwlchan:
So the stats API doesn’t give us what we want, I’m just going to drop this for now.

PR #797, opened 2017-08-17
Add batch job definition for tif-metadata job

kenoir:

Adding the final batch infra to run the tif-metadata batch job

### Who is this change for?

💻 

- [x] Run `terraform apply`.


kenoir:
For https://github.com/wellcometrust/platform-api/issues/505

Issue #759, opened 2017-08-07
Filter out images from the API that don’t have technical metadata


alexwlchan:
So we don’t have broken images showing up in the API.

pollecuttn:
@alexwlchan Is this something the Experience team need to be aware of either because it helps them or in case they don't get the results they expect?

alexwlchan:
Practical effect of this is that the API should have less broken images – no change required for the Experience team, but the signal-to-noise ratio of results should increase.

We’re planning to use technical metadata (e.g. image dimensions) as a proxy for “does this image exist?”.

alexwlchan:
Resolved by #800, we’ve run a reindex and made this the prod API.

PR #798, opened 2017-08-17
[WIP] Reduce boilerplate in the MiroTransformable tests

alexwlchan:

When transforming a Miro record, we need a bunch of fields to be in place of the transform fails (e.g. image_copyright_cleared). Having to put those on every test is tedious and boring – so don’t. Instead define a wrapper method that adds those fields for us.

Extracted from work on #759.

### Who is this change for?

Developers who don’t want to write boilerplate.


alexwlchan:
y u lose my pr travis

alexwlchan:
Looks like GitHub hooks are super slow today (this only just appeared in Slack)

alexwlchan:
One of the changes in #759 is that `image_tech_file_size` will become a required non-empty field in Miro records, so we want to make sure it’s present on every record under test.

PR #800, opened 2017-08-17
Drop images from the API that don’t have technical metadata

### LINKED ISSUE #759: Filter out images from the API that don’t have technical metadata

alexwlchan:

Discard images from the API that don’t have technical metadata – i.e., no underlying file.

Related: #759.

### Who is this change for?

An Experience team who don’t want broken image links.

- [ ] Deployed new versions


alexwlchan:
Transformer tests are failing (no, I don’t know why) – but I think this patch is ready for review.

Issue #792, opened 2017-08-15
Add thumbnail on works


jtweed:
Add a `thumbnail` **include** to works, that returns the URL of a thumbnail image and its associated licence.

The license data for Miro images should be ingested from `image_use_restrictions`. This _should_ be a valid CC licence for all V images, but there may be some messy edge cases that we'll need to work through.

```
{
  "type": "Work",
  ...
  "thumbnail": {
    "type": "Location",
    "locationType": "thumbnail-image",
    "url": "https://iiif.wellcomecollection.org/image/V0042826ER.jpg/full/300,/0/default.jpg",
    "license": {
      "type": "License",
      "licenseType": "CC-BY",
      "label": "Attribution 4.0 International (CC BY 4.0)",
      "url" : "http://creativecommons.org/licenses/by/4.0/"
    }
  }
}
```

kenoir:
Add display type for license and location.

kenoir:
Swagger docs too.

alexwlchan:
Okay, I think all the work is done for this. We need to decide how we’re doing autogen and implement that, but after that the pieces should topple like dominos.

alexwlchan:
Ah wait, I’d missed this was an include!

alexwlchan:
These are now exposed in the prod API:

🎉 ✨

PR #811, opened 2017-08-18
Add thumbnails and licenses to our models and Swagger

alexwlchan:

This lays the groundwork for putting thumbnails and licenses in the API. In particular, our `Work` model now knows about `Location` and `License`, and likewise `DisplayWork` has enough information to show them in the Swagger docs. Actually populating these models is deferred to a separate patch.

The following parameter is added to the `Work` definition:

```json
"thumbnail": {
  "description": "Relates any thing to the location of a representative thumbnail image",
  "$ref": "#/definitions/Location"
}
```

And we add two new definitions:

```json
"License": {
  "type": "object",
  "properties": {
    "licenseType": {
      "type": "string",
      "description": "A type of license under which the work in question is released to the public.",
      "enum": [
        "CC-BY",
        "CC-BY-NC"
      ]
    },
    "label": {
      "type": "string",
      "description": "The title or other short name of a license"
    },
    "url": {
      "type": "string",
      "description": "URL to the full text of a license"
    },
    "type": {
      "type": "string",
      "description": "A type of thing",
      "readOnly": true
    }
  },
  "description": "The specific license under which the work in question is released to the public - for example, one of the forms of Creative Commons - if it is a precise license to which a link can be made."
},
"Location": {
  "type": "object",
    "properties": {
      "locationType": {
        "type": "string",
        "description": "The type of location that an item is accessible from.",
        "enum": [
          "thumbnail-image"
        ]
      },
    "url": {
      "type": "string",
      "description": "The title or other short name of a license"
    },
    "license": {
      "description": "The specific license under which the work in question is released to the public - for example, one of the forms of Creative Commons - if it is a precise license to which a link can be made.",
      "$ref": "#/definitions/License"
    },
    "type": {
      "type": "string",
      "description": "A type of thing",
      "readOnly": true
    }
  },
  "description": "A location that provides access to an item"
}
```

Plus a little bit of refactoring in the Swagger tests so we can add more if it gets more complicated.

Related: #792.

### Who is this change for?

Users of the Catalogue API who want to know what they’re receiving.

- [ ] Reviewed by @jtweed
- [x] Updated the Swagger documentation
- [ ] Deployed new versions

### DIFF
--- a/api/src/main/scala/uk/ac/wellcome/platform/api/models/DisplayLicense.scala
+++ b/api/src/main/scala/uk/ac/wellcome/platform/api/models/DisplayLicense.scala
@@ -1,0 +1,26 @@
+package uk.ac.wellcome.platform.api.models
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import io.swagger.annotations.{ApiModel, ApiModelProperty}
+
+@ApiModel(
+  value = "License",
+  description =
+    "The specific license under which the work in question is released to the public - for example, one of the forms of Creative Commons - if it is a precise license to which a link can be made."
+)
+case class DisplayLicense(
+  @ApiModelProperty(
+    value =
+      "A type of license under which the work in question is released to the public.",
+    allowableValues = "CC-BY, CC-BY-NC"
+  ) licenseType: String,
+  @ApiModelProperty(
+    value = "The title or other short name of a license"
+  ) label: String,
+  @ApiModelProperty(
+    value = "URL to the full text of a license"
+  ) url: String
+) {
+  @ApiModelProperty(readOnly = true, value = "A type of thing")
+  @JsonProperty("type") val ontologyType: String = "License"
+}
--- a/api/src/main/scala/uk/ac/wellcome/platform/api/models/DisplayLocation.scala
+++ b/api/src/main/scala/uk/ac/wellcome/platform/api/models/DisplayLocation.scala
@@ -1,0 +1,25 @@
+package uk.ac.wellcome.platform.api.models
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import io.swagger.annotations.{ApiModel, ApiModelProperty}
+
+@ApiModel(
+  value = "Location",
+  description = "A location that provides access to an item"
+)
+case class DisplayLocation(
+  @ApiModelProperty(
+    value = "The type of location that an item is accessible from.",
+    allowableValues = "thumbnail-image"
+  ) locationType: String,
+  @ApiModelProperty(
+    value = "The title or other short name of a license"
+  ) url: Option[String] = None,
+  @ApiModelProperty(
+    value =
+      "The specific license under which the work in question is released to the public - for example, one of the forms of Creative Commons - if it is a precise license to which a link can be made."
+  ) license: DisplayLicense
+) {
+  @ApiModelProperty(readOnly = true, value = "A type of thing")
+  @JsonProperty("type") val ontologyType: String = "Location"
+}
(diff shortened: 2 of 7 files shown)


api/src/main/scala/uk/ac/wellcome/platform/api/models/DisplayLicense.scala
```diff
@@ -0,0 +1,27 @@
+package uk.ac.wellcome.platform.api.models
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import io.swagger.annotations.{ApiModel, ApiModelProperty}
+
+@ApiModel(
+  value = "License",
+  description =
+    "The specific license under which the work in question is released to the public - for example, one of the forms of Creative Commons - if it is a precise license to which a link can be made."
```
alexwlchan: This text is copied from the .ttl files in the ontologies folder. Longer term, might be nice if we could populate these directly from the .ttl files.

jtweed: That would be great and something I'm sure @alicefuzier would love to help with 😄

api/src/main/scala/uk/ac/wellcome/platform/api/models/DisplayLicense.scala
```diff
@@ -0,0 +1,26 @@
...
+)
+case class DisplayLicense(
+  @ApiModelProperty(
+    value =
+      "A type of license under which the work in question is released to the public.",
+    allowableValues = "CC-BY, CC-BY-NC"
```
jtweed: Can this be an enum of some kind in the model and then set dynamically here?

api/src/main/scala/uk/ac/wellcome/platform/api/models/DisplayLocation.scala
```diff
@@ -0,0 +1,25 @@
...
+import io.swagger.annotations.{ApiModel, ApiModelProperty}
+
+@ApiModel(
+  value = "Location",
+  description = "A location that provides access to an item"
+)
+case class DisplayLocation(
+  @ApiModelProperty(
+    value = "The type of location that an item is accessible from.",
+    allowableValues = "thumbnail-image"
```
jtweed: Same as above.

common/src/main/scala/uk/ac/wellcome/models/License.scala
```diff
@@ -0,0 +1,11 @@
+package uk.ac.wellcome.models
+
+import com.fasterxml.jackson.annotation.JsonProperty
+
+case class License(
+  val licenseType: String,
+  val label: String,
+  val url: String
+) {
+  @JsonProperty("type") val ldType: String = "License"
```
jtweed: Are we using ontologyType or ldType here?

common/src/main/scala/uk/ac/wellcome/models/Work.scala
```diff
@@ -20,7 +20,8 @@ case class Work(
   createdDate: Option[Period] = None,
   subjects: List[Concept] = List(),
   creators: List[Agent] = List(),
-  genres: List[Concept] = List()
+  genres: List[Concept] = List(),
+  thumbnail: Option[License] = None
 ) {
   @JsonProperty("type") val ldType: String = "Work"
```
jtweed: As above.

common/src/main/scala/uk/ac/wellcome/models/License.scala
alexwlchan: I made up `ldType` (LD = Linked Data) when I started doing these models, but let’s standardise on `ontologyType` instead. I’ll fix our other models to match.

api/src/main/scala/uk/ac/wellcome/platform/api/models/DisplayLicense.scala
alexwlchan: I really want these to be dynamically populated from JSON files that live externally – that’s coming in a separate patch. Happy to leave as hard-coded for now?

alexwlchan:
Merging as-is, other changes will come separately.

Issue #806, opened 2017-08-17
Define scheduled tasks in Terraform config


alexwlchan:
- [x] Add a pin to our aws-provider module
- [x] Use the upgraded version to define scheduled tasks in Terraform

PR #812, opened 2017-08-18
Add ECS Scheduled tasks to Terraform

kenoir:

Describing all our infra in code in order that it can be audited and recreated.

### Who is this change for?

💻 

- [X] Run `terraform apply`.


alexwlchan:
Resolves #806.

terraform/cloudwatch.tf
```diff
@@ -47,3 +47,45 @@ resource "aws_cloudwatch_event_rule" "ecs_container_instance_state_change" {
 }
 PATTERN
 }
+
+# ECS Scheduled tasks
```
alexwlchan: I’d be tempted to put these in a separate file.

terraform/cloudwatch.tf
```diff
@@ -47,3 +47,45 @@ resource "aws_cloudwatch_event_rule" "ecs_container_instance_state_change" {
...
+    task_definition_arn = "${module.gatling_catalogue_api.task_arn}"
+  }
+}
+
+resource "aws_iam_role" "scheduled_tasks_role" {
+  name               = "scheduled_tasks_role"
+  assume_role_policy = "${data.aws_iam_policy_document.assume_scheduled_tasks_role.json}"
+}
+
+data "aws_iam_policy_document" "assume_scheduled_tasks_role" {
```
alexwlchan: These should be in a separate file, or maybe a `scheduled_task` module?

You pass in a rule, a cluster ARN and a task definition ARN, rest is filled in for you.

kenoir: I wasn't sure about the best way to abstract these. A `scheduled_task` module makes sense I think.

terraform/ecs_script_task/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "task_arn" {
+  value = "${aws_ecs_task_definition.task.arn}"
+}
```
alexwlchan: Do we need this output? I don’t see it used anywhere.

PR #817, opened 2017-08-18
Autogen of license-related code

alexwlchan:
Related: #792


Makefile
```diff
@@ -224,14 +224,20 @@ format-terraform:
...
-	format-scala
+format: format-terraform format-scala
 
 check-format: format
 	git diff --exit-code
 
 
+autogen:
+	pip3 install jinja2
+	./scripts/run_autogen.py
```
alexwlchan: This should really run inside a container.

common/src/main/scala/uk/ac/wellcome/models/License.scala
```diff
@@ -2,10 +2,46 @@ package uk.ac.wellcome.models
...
-  val label: String,
+trait License {
+  val licenseType: String
+  val label: String
   val url: String
-) {
   @JsonProperty("type") val ontologyType: String = "License"
 }
+
+// ======================= AUTOGENERATED CODE STARTS ======================= //
```
alexwlchan: Wrong! Everything in this file is autogenerated – this should go right at the top.

Also, this comment should really exclude this code from scalafmt.

alexwlchan:
A couple of other interesting things about this patch:

* autogen gets its own directory, README and Makefile – this is an attempt to make individual units a bit more self-contained, and start breaking up the monolithic Makefile in the root of the repo. What do people think of this approach?

* autogen felt like the least bad option here. We add a small amount of complexity to our build process in exchange for substantially easier management of data, and proper compile-time checking of our types. I had a crack at the reflection and runtime-lookup route – both were more complicated than this.

* It’s a biggish patch, but a lot of it is autogen infrastructure. When we add other autogen'd types (e.g. workType), I think their patches should be much smaller.

* The one thing I haven’t done run in this patch which I should is give some way to enumerate all the licenses – which we should use in the Swagger docs. I don’t have a working Scala environment to hand, so I’m going to leave that for now.

alexwlchan:
If you know a good way to do code autogeneration inside Scala/sbt, I’m all ears – I threw this together as a Friday afternoon idea.

I did go straight for Docker as a way to reduce the dependencies, and my thought would be that we’d check in the autogenerated code, so you only need to run autogen if you’ve changed the data sources.

But yeah, I think best thing to do is discuss it in-person tomorrow.

kenoir:
Closing for now to avoid confusion.

PR #822, opened 2017-08-21
Populate the thumbnail field on a MIRO work

alexwlchan:

Now we have models for Location and License, and some License objects (how they appear in the world tbc), let's start putting some data in them!

This patch only covers the transformer changes – changes to display logic in the API will be a separate patch. Merging is blocked on us deciding how to define Licenses, but stands as a standalone patch.

Related: #792.

### Who is this change for?

The Experience team, who are planning to use thumbnails as a proxy for licenses on items.

- [ ] Deployed new versions

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/models/MiroTransformable.scala
+++ b/common/src/main/scala/uk/ac/wellcome/models/MiroTransformable.scala
@@ -214,7 +214,7 @@ case class MiroTransformable(MiroID: String,
       }
       val thumbnail = Location(
         locationType = "thumbnail-image",
-        url = s"https://iiif.wellcomecollection.org/image/$MiroID.jpg/full/300,/0/default.jpg",
+        url = Some(s"https://iiif.wellcomecollection.org/image/$MiroID.jpg/full/300,/0/default.jpg"),
         license = chooseLicense(useRestrictions=useRestrictions)
       )
 
@@ -226,7 +226,7 @@ case class MiroTransformable(MiroID: String,
         creators = creators ++ secondaryCreators,
         subjects = subjects,
         genres = genres,
-        thumbnail = thumbnail
+        thumbnail = Some(thumbnail)
       )
     }
   }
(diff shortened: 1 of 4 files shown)


common/src/main/scala/uk/ac/wellcome/models/MiroTransformable.scala
```diff
@@ -23,9 +23,19 @@ case class MiroTransformableData(
...
   @JsonProperty("image_lc_genre") lcGenre: Option[String],
-  @JsonProperty("image_tech_file_size") techFileSize: Option[List[String]]
+  @JsonProperty("image_tech_file_size") techFileSize: Option[List[String]],
+  @JsonProperty("image_use_restrictions") useRestrictions: Option[String]
 )
 
+/** Thrown for errors in the data which need further investigation or indicate
+ *  a transformer bug.
+ */
+case class MiroDataException(message: String)
```
kenoir: Be aware this will trigger the try backoff failure state which may not be what you intended - a ShouldNotTransformException will not do so (see `SQSMessageReceiver`).

common/src/main/scala/uk/ac/wellcome/models/MiroTransformable.scala
```diff
@@ -194,15 +204,79 @@ case class MiroTransformable(MiroID: String,
...
+      // TODO: Less magic strings here.
+      val useRestrictions = miroData.useRestrictions match {
+        case Some(s) => s
+        case None => throw new MiroDataException(
+          "No value provided for image_use_restrictions?"
+        )
+      }
+      val thumbnail = Location(
+        locationType = "thumbnail-image",
+        url = Some(s"https://iiif.wellcomecollection.org/image/$MiroID.jpg/full/300,/0/default.jpg"),
```
kenoir: Definitely move this template into config.

alexwlchan:
Duplicate of #832.

PR #826, opened 2017-08-21
Add an Item model

kenoir:

Adds an item model to move towards representing items from the ontology in the API.

### Who is this change for?

💻  & 📚 

- [ ] Deployed new versions


kenoir:

PR #827, opened 2017-08-21
Adds Item to ES mapping

kenoir:

Moving towards ingesting items into the API this adds an ES mapping to make it possible to store the items as per the ontology.

This PR also refactors the mappings a little to reduce code duplication.

### Who is this change for?

💻  & 📚 

- [ ] Deployed new versions

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/elasticsearch/mappings/WorksIndex.scala
+++ b/common/src/main/scala/uk/ac/wellcome/elasticsearch/mappings/WorksIndex.scala
@@ -1,14 +1,13 @@
 package uk.ac.wellcome.elasticsearch.mappings
 
 import com.google.inject.Inject
+import com.sksamuel.elastic4s.analyzers._
 import com.sksamuel.elastic4s.http.ElasticDsl._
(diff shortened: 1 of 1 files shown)


kenoir:

common/src/main/scala/uk/ac/wellcome/elasticsearch/mappings/WorksIndex.scala
```diff
@@ -19,48 +18,60 @@ class WorksIndex @Inject()(client: HttpClient,
...
+      keywordField("type"),
+      keywordField("locationType"),
+      textField("url"),
+      license
+    )
+
+  def date(fieldName: String) = objectField(fieldName).fields(
+    textField("label"),
+    keywordField("type")
+  )
```
alexwlchan: Could this be simplified to

```scala
def date(fieldName: String) = labelledTextField(fieldName)
```

?

kenoir: I think we might modify this to actually be a date at some point so i'd like to keep it separate.

PR #828, opened 2017-08-21
Expose thumbnails in the API

alexwlchan:
This is the display logic corresponding to #822 – once the ingestor puts information about thumbnails in Elasticsearch, make sure it appears on the other side.

Related: #792.


api/src/main/scala/uk/ac/wellcome/platform/api/models/DisplayWork.scala
```diff
@@ -85,7 +85,13 @@ case object DisplayWork {
       identifiers =
         if (includes.identifiers)
           Some(identifiedWork.work.identifiers.map(DisplayIdentifier(_)))
-        else None
+        else None,
+      thumbnail =
+        if (includes.thumbnail)
+          identifiedWork.work.thumbnail match {
```
kenoir: You can map here.

PR #837, opened 2017-08-22
More fixes to the Miro adapter

alexwlchan:
* Don’t fail if we hit throttle limits, just back off and come back later.
* Fix a bunch of encoding nonsense that was causing issues in the staging API. Make sure we get the encoding right, and throw away only the minimal set of special characters that cause issues.

### DIFF
--- a/miro_adapter/miro_adapter.py
+++ b/miro_adapter/miro_adapter.py
@@ -17,8 +17,10 @@ Options:
 """
 
 import json
+import time
 
 import boto3
+from botocore.exceptions import ClientError
 import docopt
 
 from utils import generate_images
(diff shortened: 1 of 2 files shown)


miro_adapter/miro_adapter.py
```diff
@@ -32,18 +34,31 @@ def push_to_dynamodb(table_name, collection_name, image_data):
...
+                        'ReindexVersion': 0,
+                        'data': json.dumps(image, separators=(',', ':'))
+                    }
+                )
+
+            # We hit write limits on the DynamoDB table.  Wait a bit then
+            # carry on.  Wait up to five minutes between bursts if necessary.
+            except ClientError as err:
+                if err.response['Error']['Code'] == 'ProvisionedThroughputExceededException':
+                    wait_time = min(wait_time * 2, 5 * 50)
```
alexwlchan: 50????

miro_adapter/utils.py
```diff
@@ -96,7 +86,11 @@ def read_image_chunks_from_s3(bucket, key):
...
+    # Because this is a stream parser, lxml doesn't know about the encoding
+    # declaration at the top of the Miro XML exports.  We have to tell it.
+    # It's a bit magic, but this is the easiest way to do it.
+    iso_88591_parser = etree.XMLParser(encoding='iso-8859-1')
     for xml_chunk in read_image_chunks_from_s3(bucket, key):
         xml_string = fix_miro_xml_entities(xml_chunk)
-        lxml_elem = etree.fromstring(xml_string)
-        yield elem_to_dict(lxml_elem)
+        lxml_elem = etree.fromstring(xml_string, parser=iso_88591_parser)
+        yield lxml_elem
```
alexwlchan: oops

PR #840, opened 2017-08-22
Add an ontologyType column to the ID minter database

alexwlchan:
### What is this PR trying to achieve?

Allow us to record ontologyType in the ID minter database. This allows us to have multiple rows with the same MiroID – recording canonical IDs for the corresponding work and item.

This patch sets up the underlying database field to store the ontologyType, although nothing in the ID minter yet sets it to anything except "Work". That will come in a separate patch.

### Who is this change for?

Folks working on #839.

### Have the following been considered/are they needed?

- [ ] Deployed new versions

PR #841, opened 2017-08-22
Preliminary refactoring of IdentifiersDao in the ID minter

alexwlchan:
Separated from #840 – this is a bit of refactoring in the IdentifiersDao.

Underlying ideas:

* Reduce code repetition
* More whimsy/unique strings (yes I like whimsy, also makes it easier to see which tests are in-flight and debug broken things)
* Prepare for the fact that we’re going to get more complex uniqueness checks, and so making it easy to test “can I do X but not Y” will be helpful

### DIFF
--- a/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
+++ b/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
@@ -46,7 +46,7 @@ class IdentifiersDao @Inject()(db: DB, identifiers: IdentifiersTable)
         throw e
     }
 
-  def saveIdentifier(identifier: Identifier): Future[Unit] = {
+  def saveIdentifier(identifier: Identifier): Future[Int] = {
     val insertIntoDbFuture = Future {
       blocking {
         info(s"putting new identifier $identifier")
@@ -57,7 +57,6 @@ class IdentifiersDao @Inject()(db: DB, identifiers: IdentifiersTable)
               identifiers.column.CanonicalID -> identifier.CanonicalID,
               identifiers.column.MiroID -> identifier.MiroID)
         }.update().apply()
-        ()
       }
     }
     insertIntoDbFuture.onFailure {
--- a/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
+++ b/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
@@ -16,21 +16,6 @@ class IdentifiersDao @Inject()(db: DB, identifiers: IdentifiersTable)
 
   implicit val session = AutoSession(db.settingsProvider)
 
-  def lookupCanonicalID(canonicalID: String): Future[Option[Identifier]] =
-    Future {
-      blocking {
-        info(s"About to search for canonical ID $canonicalID in Identifiers")
-        val i = identifiers.i
-        withSQL {
-          select.from(identifiers as i).where.eq(i.CanonicalID, canonicalID)
-        }.map(Identifier(i)).single.apply()
-      }
-    } recover {
-      case e: Throwable =>
-        error(s"Failed getting canonicalID $canonicalID in Identifiers", e)
-        throw e
-    }
-
   def lookupMiroID(miroID: String): Future[Option[Identifier]] =
     Future {
       blocking {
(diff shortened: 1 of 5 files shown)


id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
```diff
@@ -16,22 +16,22 @@ class IdentifiersDao @Inject()(db: DB, identifiers: IdentifiersTable)
...
       }
     } recover {
       case e: Throwable =>
-        error(s"Failed getting MiroID $miroId in DynamoDB", e)
+        error(s"Failed getting MiroID $miroID in Identifiers", e)
         throw e
     }
 
-  def saveIdentifier(identifier: Identifier): Future[Unit] = {
+  def saveIdentifier(identifier: Identifier): Future[Int] = {
```
kenoir: Why's this an `int`?

id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
```diff
@@ -16,22 +16,22 @@ class IdentifiersDao @Inject()(db: DB, identifiers: IdentifiersTable)
 
   implicit val session = AutoSession(db.settingsProvider)
 
-  def findSourceIdInDb(miroId: String): Future[Option[Identifier]] =
+  def lookupMiroID(miroID: String): Future[Option[Identifier]] =
```
kenoir: Makes sense to have this be more specific as this only looks for a single thing, but worth thinking about (if not implementing) how this will work with multiple identifiers.

id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDaoTest.scala
```diff
@@ -16,37 +16,33 @@ class IdentifiersDaoTest
...
-          .into(identifiersTable)
-          .namedValues(identifiersTable.column.CanonicalID -> canonicalId,
-                       identifiersTable.column.MiroID -> miroId)
-      }.update().apply()
-
-      whenReady(identifiersDao.findSourceIdInDb(miroId)) { maybeIdentifier =>
+  describe("lookupMiroID") {
+    it("should return a future of Some[Identifier] if it can find a MiroID in the DB") {
+      val identifier = Identifier(
+        CanonicalID = "A sand snail",
```
kenoir: 🐌

id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDaoTest.scala
```diff
@@ -16,37 +16,33 @@ class IdentifiersDaoTest
...
     }
   }
 
   describe("saveIdentifier") {
-
     it("should insert the provided identifier into the database") {
-      val identifier = Identifier(CanonicalID = "5678", MiroID = "1234")
+      val identifier = Identifier(
+        CanonicalID = "A provision of porpoises",
+        MiroID = "A picture of pangolins"
```
kenoir: +1 Pangolins are one of my favourite animals.

id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
alexwlchan: That’s what the `apply()` method returns, I didn't have time to look up why. Felt better than just dropping it and returning an empty `Unit`.

id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
alexwlchan: My current thought (with half an eye to ontologyType as I wrote this code) is that, at least for now, you stick an extra parameter on this method:

```scala
def lookupMiroID(miroID: String, ontologyType: String)
```

Further down, doing this in a way that doesn’t require us to hard-code each lookup as a method we might want would be nice, but I think hand-rolling these is still simpler (for now).

alexwlchan: But really I just didn’t want to continue down a path where Miro ID is a “special” identifier – and exactly what we need when Sierra/CALM enter will determine what the final structure looks like.

id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDaoTest.scala
alicefuzier: I didn't even know that pangolins existed...I learned something today

Issue #839, opened 2017-08-22
Upgrade the ID minter to support minting IDs on items


alexwlchan:
Related: #572

alexwlchan:
See also: #886, #884, #879, #878, #874.

alexwlchan:
🎉 

https://api.wellcomecollection.org/catalogue/v0/works?includes=items

PR #847, opened 2017-08-23
[WIP] Works in the ID minter draw their ontologyType from their model

alexwlchan:

We no longer default to treating instances of Identifier as having ontologyType Work; we deduce it from the Identifiable trait.

### Who is this change for?

Folks working on #839.

- [ ] Deployed new versions


alexwlchan:
Closing, this has been superseded.

PR #851, opened 2017-08-23
Implement a JSON tree walker in the ID minter

alexwlchan:
This patch only covers implementing the tree walker itself – actually passing the `ontologyType` and `sourceIdentifiers` off the database is a separate patch. A pure in-memory operation, which hopefully means it’s really easy to test.

Still to do:

* [ ] Proper error handling if we encounter some identifier-looking blocks that don’t match our worldview
* [x] ~Check the type on those blocks as well?~
* [x] I’m not sure `ontologyType` is the string field I want, actually
* [x] Tests that it finds and updates Identifiable blocks accurately
* [x] ~Maybe even cast back to a string and re-parse as a list of `SourceIdentifier` blocks, idk~

### DIFF
--- a/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/utils/IdentifiableWalker.scala
+++ b/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/utils/IdentifiableWalker.scala
@@ -1,0 +1,23 @@
+package uk.ac.wellcome.platform.idminter.utils
+
+import com.fasterxml.jackson.databind.{JsonNode, ObjectMapper}
+
+/* This object takes a JSON string (which is assumed to be a map) and walks
+ * it, looking for objects that conform to the Identifiable trait.
+ * Rather than using first-class Scala types, this does a completely generic
+ * tree-walk -- we're trading some type safety for flexibility.
+ *
+ * In this case, the interesting features of the Identifiable trait are:
+ *
+ *   - a string field called "ontologyType"
+ *   - a list field called "identifiers" which has a non-empty list of
+ *     objects which deserialise as instances of SourceIdentifier
+ */
+object IdentifiableWalker {
+
+  def readTree(jsonString: String): JsonNode = {
+    val mapper = new ObjectMapper()
+    mapper.readTree(jsonString)
+  }
+
+}
--- a/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/utils/IdentifiableWalker.scala
+++ b/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/utils/IdentifiableWalker.scala
@@ -14,7 +14,7 @@ import com.fasterxml.jackson.databind.node.{
 import uk.ac.wellcome.models.SourceIdentifier
 import uk.ac.wellcome.utils.JsonUtil
 
-/* This object takes a JSON string (which is assumed to be a map) and walks
+/* This class takes a JSON string (which is assumed to be a map) and walks
  * it, looking for objects that conform to the Identifiable trait.
  * Rather than using first-class Scala types, this does a completely generic
  * tree-walk -- we're trading some type safety for flexibility.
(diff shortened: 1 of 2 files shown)


id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/utils/IdentifiableWalker.scala
```diff
@@ -0,0 +1,108 @@
...
+  def identifyDocument(jsonString: String): String = {
+    val node = readTree(jsonString)
+    JsonUtil.toJson(rebuildObjectNode(node)).get
+  }
+
+  // Dummy function for generating canonical IDs.  Eventually this should
+  // use the identifier proper, but for the purposes of testing the tree
+  // parsing code is just drops in a string that proves it extracted the
+  // identifiers correctly.
+  def generateCanonicalId(sourceIdentifiers: List[SourceIdentifier],
```
alexwlchan: Could you find a way to pass this as a callback parameter, so tests can do pure in-memory operations that are super fast, and the real thing gets the database-wired version?

(Answer: probably, but not now!)

alexwlchan: What if this was a case object?

kenoir: Not sure exactly what you mean.

alexwlchan: I’ve got it working, and implemented in code. ~Probably easiest to show you when I get in~

I’ll push the code I’ve got to show what I mean.

id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/utils/IdentifiableWalker.scala
```diff
@@ -29,8 +29,15 @@ import uk.ac.wellcome.utils.JsonUtil
...
  * passed through unmodified.
+ *
+ * This class takes a single parameter: a `generateCanonicalId` callback
+ * which receives a list of `SourceIdentifier` instances, and the ontology
+ * type of the object to be identifier.  This callback should return the
+ * canonical ID.
  */
-object IdentifiableWalker {
+case class IdentifiableWalker(
+  generateCanonicalId: (List[SourceIdentifier], String) => String
```
kenoir: You could also compose this functionality by requiring this function in a base class and applying it via a trait.

alexwlchan:
I could write more tests for this for hours, but I’m not sure it would add anything. The code is fairly simple, and I think these tests give us good coverage. I think this is ready for review.

alexwlchan:
(I am also punting the “what if error” question, and saying that it just bubbles up whatever exception it throws if/when that happens.)

id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/utils/IdentifiableWalker.scala
```diff
@@ -0,0 +1,100 @@
...
+    val newNode = new ObjectNode(new JsonNodeFactory(false))
+    for (field <- node.fields) {
+      newNode.set(field.getKey, processValue(field.getValue))
+    }
+
+    if (node.has("identifiers") && node.has("type")) {
+
+      // This code may throw an exception if these identifiers don't look
+      // correct, which may screw with the TryBackoff mechanism.
+      // TODO: Be less crappy with error handling here.
```
kenoir: You may need to reimplement a structure like that in `SQSMessageReceiver` in the transformer (or abstract it to common somehow).

alexwlchan: sorry, not following

reimplement a structure like what?

kenoir: Throw an exception indicating that you want to stop gracefully and handle it as such.

PR #852, opened 2017-08-23
Add Items to Works in MiroTransformable

kenoir:

Ingest Items on Works for Miro data.

### Who is this change for?

💻  & 📚 

- [ ] Deployed new versions


kenoir:

alexwlchan:
What you’ve got looks good. Any tests? And does it matter that we won’t be adding any canonical IDs to these items?

kenoir:
> And does it matter that we won’t be adding any canonical IDs to these items?
Yes it does matter.

> Any tests?
There are already tests which cover this functionality (but as I forgot to actually add items to works they haven't needed to be updated).

alexwlchan:
Also, having the different bits of transformer logic in different methods makes these patches much cleaner, I like.

PR #855, opened 2017-08-24
Use our own AMI for batch jobs

alicefuzier:
Amazon ECS optimized AMI does not have enough disk space for our tif-conversion jobs and they end up failing with CAnnotCreateContainerError when the container instance disk becomes full. This creates a bigger AMI with 100GB disk space and configured to clean up the containers 1 minute after they have stopped to save space.
### Who is this change for?
Everyone who cares about tif conversion (anyone?)

- [ ] Run `terraform apply`.


packer/batch_userdata.txt
```diff
@@ -0,0 +1,36 @@
+Content-Type: multipart/mixed; boundary="==BOUNDARY=="
+MIME-Version: 1.0
+
+--==BOUNDARY==
+Content-Type: text/x-shellscript; charset="us-ascii"
+
+#!/bin/bash
+# Set ECS agent configuration options
+cat <<'EOF' > /etc/ecs/ecs.config
+ECS_CLUSTER=default_Batch_2759339e-568d-39fa-944e-856ad95cc8be
```
alicefuzier: Find a way to not hard code this and get it out of terraform

terraform/batch.tf
```diff
@@ -1,10 +1,11 @@
 module "aws_batch_compute" {
   source = "./batch_compute"
-
+  name = "default"
   subnets            = "${join(",", formatlist("\"%s\"", module.vpc_batch.subnets))}"
   key_name           = "${var.key_name}"
   admin_cidr_ingress = "${var.admin_cidr_ingress}"
   vpc_id             = "${module.vpc_batch.vpc_id}"
+  image_id = "ami-0657a97f"
```
alicefuzier: Do not hard code, is it possible to get it from packer?

kenoir: Probably possible to generate a tfvars file, but you might be able to reference this with a data block too. AMI IDs are globally unique identifiers so I think it's acceptable to reference it somewhere.

kenoir:
Closing as we don't need to merge this to move forward. If we want to switch out our current COreOS container hosts for these in the future we can return to this changeset.

PR #857, opened 2017-08-25
Looking up IDs in the ID minter is now (more) source identifier agnostic

alexwlchan:

Rather than hard-coding logic for Miro, Calm, Sierra source identifiers – have a single `lookupID()` method that makes one big SQL query for us.

Another piece of #839.

### Who is this change for?

Devs who want generic lookups in the ID minter.

- [ ] Deployed new versions

### DIFF
--- a/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
+++ b/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
@@ -12,6 +12,9 @@ import uk.ac.wellcome.utils.GlobalExecutionContext.context
 
 import scala.concurrent.{Future, blocking}
 
+case class UnableToMintIdentifierException(message: String)
+    extends Exception(message)
+
 @Singleton
 class IdentifiersDao @Inject()(db: DB, identifiers: IdentifiersTable)
     extends Logging {
@@ -25,7 +28,10 @@ class IdentifiersDao @Inject()(db: DB, identifiers: IdentifiersTable)
    * source identifiers.
    */
   def lookupID(sourceIdentifiers: List[SourceIdentifier],
-               ontologyType: String): Future[Option[Identifier]] =
+               ontologyType: String): Future[Option[Identifier]] = {
+    if (sourceIdentifiers.isEmpty) {
+      throw new UnableToMintIdentifierException("No source identifiers supplied!")
+    }
     Future {
       blocking {
         info(s"About to search for existing ID matching $identifiers and $ontologyType")
@@ -61,6 +67,7 @@ class IdentifiersDao @Inject()(db: DB, identifiers: IdentifiersTable)
         }.map(Identifier(i)).single.apply()
       }
     }
+  }
 
   /* For a given source identifier scheme (e.g. "miro-image-number") and a
    * corresponding column in the SQL database (e.g. i.MiroID), add a condition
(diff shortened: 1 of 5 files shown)


alexwlchan:
Okay, I think this is a useful, self-contained patch.

The key part is the `lookupID()` method, the rest of the patch is a lot of tests and some small changes to get Calm AltRefNos working. I’ve flipped `lookupMiroID()` to use the new lookup under the hood, but I’ll defer replacing other calls to it for a separate patch.

id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
```diff
@@ -5,37 +5,117 @@ import javax.inject.Singleton
...
-            .eq(i.MiroID, miroID)
-        }.map(Identifier(i)).single.apply()
+
+            // Add conditions for matching on different source identifiers.
+            .map { sql: ConditionSQLBuilder[String] =>
+              addConditionForLookingUpID(
+                sql = sql,
+                sourceIdentifiers = sourceIdentifiers,
+                column = i.MiroID,
+                identifierScheme = "miro-image-number"
```
alexwlchan: reminder to self: you put in the `IdentifierSchemes` module for a reason, you should use it!

id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
```diff
@@ -5,37 +5,117 @@ import javax.inject.Singleton
...
+   * This method looks for existing IDs that have matching ontology type and
+   * source identifiers.
+   */
+  def lookupID(sourceIdentifiers: List[SourceIdentifier],
+               ontologyType: String): Future[Option[Identifier]] = {
+
+    // TODO: This exception should be handled gracefully, not sent around the
+    // TryBackoff ad infinitum.
+    if (sourceIdentifiers.isEmpty) {
+      throw new UnableToMintIdentifierException(
```
alicefuzier: I'm not sure I like a function that by signature returns a future throwing an exception. I think it should be a failed future. It's unexpected to have to deal with exceptiond with functions that return futures

id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
```diff
@@ -47,7 +130,8 @@ class IdentifiersDao @Inject()(db: DB, identifiers: IdentifiersTable)
             .namedValues(
               identifiers.column.CanonicalID -> identifier.CanonicalID,
               identifiers.column.ontologyType -> identifier.ontologyType,
-              identifiers.column.MiroID -> identifier.MiroID
+              identifiers.column.MiroID -> identifier.MiroID,
+              identifiers.column.CalmAltRefNo -> identifier.CalmAltRefNo
```
alexwlchan: This change is untested for now, and only required so I can get tests working – I’ll add proper tests for this function in a separate patch.

alicefuzier: What kind of extra tests would it need? I can see a bunch of Calm related tests below...

alexwlchan: It doesn’t, I realised they were already written…

PR #860, opened 2017-08-25
Remove lookupMiroID in favour of lookupID

alexwlchan:

The natural follow-up to #857. This removes some Miro-specific logic from the database lookups.

Yet another patch for #839.

### DIFF
--- a/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/IdentifierGenerator.scala
+++ b/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/IdentifierGenerator.scala
@@ -22,7 +22,10 @@ class IdentifierGenerator @Inject()(identifiersDao: IdentifiersDao,
       "generate-id",
       () =>
         findMiroID(work) match {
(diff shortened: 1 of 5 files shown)


id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/IdentifierGenerator.scala
```diff
@@ -32,8 +35,9 @@ class IdentifierGenerator @Inject()(identifiersDao: IdentifiersDao,
   }
 
   private def retrieveOrGenerateCanonicalId(
-    identifier: SourceIdentifier): Future[String] =
-    identifiersDao.lookupMiroID(identifier.value).flatMap {
+    identifier: SourceIdentifier,
+    ontologyType: String): Future[String] = {
+    identifiersDao.lookupID(List(identifier), ontologyType).flatMap {
```
alexwlchan: This function really needs another level of refactoring, so it takes the list of source identifiers rather than a single source identifier – but this patch is just about ripping out `lookupMiroID`.

PR #868, opened 2017-08-30
Add the ability to configure the ReindexShard in the reindexer

alexwlchan:

Adds a config option to the reindexer that allows us to reindex a specific shard.

### Who is this change for?

Being able to reindex only a subset of the database means that:

* We could (theoretically) run multiple reindexes in parallel, making it go faster
* We can reindex specific collections in Miro

- [ ] Deployed new versions


alexwlchan:
Duplicate of #870.

alexwlchan:
Wait, no. The adapter change is duplicated, the reindexer change isn’t.

PR #870, opened 2017-08-31
Build JSON intermediaries in the Miro adapter

alexwlchan:

The Miro adapter constructs a JSON document, then puts that document in DynamoDB.

This adds an intermediate step that sticks a copy of those JSON documents, one per line, in our S3 bucket.

### Who is this change for?

Platform devs – if you want to ask questions like “how many Miro records have license `Academics`”, doing it from the JSON dumps is easy, and faster than doing it from the XML dumps every time.

In particular, getting a list of Miro records in Python is now:

```python
from utils import read_json_lines_from_s3

for doc in read_json_lines_from_s3(bucket='miro-data', key='images-V.txt'):
    print(doc)
```

- [ ] Deployed new versions


kenoir:
Nice. Longer term do you think we need a solution like this for all adapters, and is S3 the correct place for that store?

miro_adapter/push_json_to_dynamodb.py
```diff
@@ -40,7 +40,7 @@ def push_to_dynamodb(table_name, collection_name, image_data):
                 Item={
                     'MiroID': image['image_no_calc'],
                     'MiroCollection': collection_name,
-                    'ReindexShard': 'default',
+                    'ReindexShard': collection_name,
```
alexwlchan: This is the one-line change required to put records in a ReindexShard by collection name.

alexwlchan:
> Longer term do you think we need a solution like this for all adapters, and is S3 the correct place for that store?

Ooh, good thought. I can imagine us wanting to do similar analysis on other data, and then this seems like a reasonably convenient format. And S3 seems as good a place as any.

That might be a nice future enhancement (and maybe we redesign the Calm adapter to work that way) – but this is really just formalising the pile of hacks I use to do analysis of the Miro data, e.g. how many images do we expect with this ruleset to appear in the API?

miro_adapter/run_miro_adapter.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env sh
+
+set -o errexit
+set -o nounset
+
+python3 miro_adapter/build_json_derivatives.py \
```
kenoir: We don't always want to do this step do we? In fact could we split this into two even smaller services?

alexwlchan: Extreme microsizing!

Not sure it’s worth the overhead of another service, but we could check if it’s up-to-date before running this step (writes are atomic, so last modified time is last completed time).

alexwlchan: Ah, but what if this sort of JSON manifest is how we do all the adapters?

An adapter could watch the files, and do the DynamoDB step every time they change!

alexwlchan:
Some of the code from this PR has been spun out into a separate package, with tests and everything!

https://pypi.org/project/lazyreader/
https://github.com/alexwlchan/lazyreader

pollecuttn:
Re the separate package, should the example in which a records is separated with commas have 
`delimiter=','`
instead of 
`delimiter=';'`

Obviously just ignore me if I've misunderstood!

alexwlchan:
> should the example in which a records is separated with commas have 
>
> `delimiter=','`
> instead of 
> `delimiter=';'`

😳 

Good spot, will fix. Thanks!

kenoir:
@alexwlchan can't see where you referencing those packages in this PR?

alexwlchan:
@kenoir I’m not, yet. I wrote and published the package at the weekend, actually using it comes today. :-)

miro_adapter/run_miro_adapter.sh
kenoir: Did this get addressed?

alexwlchan: No.

miro_adapter/build_json_derivatives.py
```diff
@@ -0,0 +1,53 @@
...
+
+if __name__ == '__main__':
+    args = docopt.docopt(__doc__)
+    image_data = generate_images(bucket=args['--bucket'], key=args['--src'])
+
+    tmp_json = tempfile.mktemp()
+    os.makedirs(os.path.dirname(tmp_json), exist_ok=True)
+    with open(tmp_json, 'w') as f:
+        for img in image_data:
+
```
alexwlchan: You need a step here that uploads an individual JSON file for each image

miro_adapter/utils.py
```diff
@@ -66,23 +68,55 @@ def fix_miro_xml_entities(xml_string):
...
             break
+
         running += new_data
-        if b'</image>' in running:
-            curr, running = running.split(b'</image>')
-            curr = curr.split(b'<image>')[-1]
-            yield (b'<image>' + curr + b'</image>')
+        if delimiter in running:
+            curr, running = running.split(delimiter)
+            yield curr + delimiter
```
alexwlchan: You should use lazyreader for this!

kenoir:
Superseded by https://github.com/wellcometrust/platform/pull/891

Issue #572, opened 2017-07-05
Add items to Catalogue API


jtweed:
Add a list of `items` to works. Items are identified objects in their own right and need to have IDs assigned by the ID minter.

`items` should be an include on works in the API.

Right now, they have no additional metadata for Miro data (they will for other sources), but we need to add them so we can properly link to the images.

See: https://github.com/wellcometrust/platform-api/blob/master/ontologies/Examples/MIRO_sample_image_of_artwork_v2.json

alexwlchan:
> `items` should be an include on works in the API.

They’re appearing in the staging API, but not as an include. I’ll fix that now.

alexwlchan:
The prod API now has (unidentified) items, so I’ve opened #876 to track putting IDs on there. We can then manage that separately.

PR #872, opened 2017-08-31
Make `items` an include on works in the API

alexwlchan:

As the description says. This is mentioned as a requirement on #572.

### Who is this change for?

People who like to stick to the requirements.

- [ ] Reviewed by @jtweed
- [ ] Deployed new versions

### DIFF
--- a/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorksIncludes.scala
+++ b/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorksIncludes.scala
@@ -29,7 +29,7 @@ case object WorksIncludes {
     val includesList = queryParam.split(",").toList
     val unrecognisedIncludes = includesList
       .filterNot(recognisedIncludes.contains)
-    if (unrecognisedIncludes.length == 0) {
+    if (unrecognisedIncludes.isEmpty) {
       WorksIncludes(
         identifiers = includesList.contains("identifiers"),
         thumbnail = includesList.contains("thumbnail"),
--- a/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorksIncludes.scala
+++ b/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorksIncludes.scala
@@ -10,7 +10,8 @@ import com.fasterxml.jackson.databind.{
 
 case class WorksIncludes(
   identifiers: Boolean = false,
-  thumbnail: Boolean = false
+  thumbnail: Boolean = false,
+  items: Boolean = false
 )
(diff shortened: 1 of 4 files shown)


api/src/main/scala/uk/ac/wellcome/platform/api/models/WorksIncludes.scala
```diff
@@ -31,7 +32,8 @@ case object WorksIncludes {
     if (unrecognisedIncludes.length == 0) {
```
alicefuzier: minor: can we use `.isEmpty` here?

api/src/test/scala/uk/ac/wellcome/platform/api/ApiWorksTest.scala
```diff
@@ -513,8 +500,7 @@ class ApiWorksTest
                           |     "title": "${work1.title}",
                           |     "creators": [],
                           |     "subjects": [ ],
-                          |     "genres": [ ],
-                          |     "items": [ ]
+                          |     "genres": [ ]
```
alicefuzier: This test is about rendering items with no ids, but items have been taken out? Or am I missing something?

api/src/main/scala/uk/ac/wellcome/platform/api/models/WorksIncludes.scala
alexwlchan: 👍

api/src/test/scala/uk/ac/wellcome/platform/api/ApiWorksTest.scala
alexwlchan: No, that’s on L194 and 210 – this test is

> should return matching results if doing a full-text search

alexwlchan:
> Looks ok, but I don't see any test where the include=items is passed in? I think at least one of the ApiTests should test this case

Agreed, but I think there’s already a test for this. See:

> should be able to render an item with no canonicalId if the items include is present

alicefuzier:
Uhm, I still think it should have its own test. The `should be able to render an item with no canonicalId if the items include is present` one was about the api not crashing if items have no id. It just now happens to require the include. If that is the only test for the include=items thing we achieve the code coverage, perhaps, but we kind of lose the documentation aspect of tests cause that is very specific to items with no id. The items include has nothing to do with that

alexwlchan:
Ah, gotcha. Will change.

alexwlchan:
Okay, latest couple of commits have a test that just covers the `?includes=items` identifier.

PR #874, opened 2017-08-31
[WIP] Use Lenses!

alicefuzier:
Use [Monocle](http://julien-truffaut.github.io/Monocle/) to reduce the size of patches needed to the idMinter to generate identifiers for things other than Work. As discussed with @alexwlchan, the id_minter will not be completely agnostic of the Json structure (as that turned out to be a difficult problem to solve) but this change should make it a one liner to tell the idminter where to add ids
### Who is this change for?
Devs

- [ ] Deployed new versions


alicefuzier:
Closing this as it's already too big. I'll be opening a few smaller PR to prepare for the switch to lenses before actually opening a lenses one

PR #879, opened 2017-09-01
Isolate logic to generate the id from the case class parsing logic

alicefuzier:
Currently the IdentifierGenerator does two things:
 - finds the source identifier in the case class structure and 
 - retrieves or generates an id
In order to switch to Lenses these two things need to be separate, so let's do that
### Who is this change for?
devs

- [ ] Deployed new versions


id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/services/IdMinterWorkerService.scala
```diff
@@ -5,23 +5,20 @@ import com.google.inject.Inject
...
 import scala.concurrent.Future
-import uk.ac.wellcome.utils.GlobalExecutionContext.context
 
 class IdMinterWorkerService @Inject()(
-  idGenerator: IdentifierGenerator,
-  writer: SNSWriter,
-  reader: SQSReader,
-  system: ActorSystem,
-  metrics: MetricsSender
+                                       something: IdEmbedder,
```
kenoir: something?

id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/IdEmbedder.scala
```diff
@@ -0,0 +1,28 @@
+package uk.ac.wellcome.platform.idminter.steps
+
+import com.google.inject.Inject
+import com.twitter.inject.Logging
+import uk.ac.wellcome.metrics.MetricsSender
+import uk.ac.wellcome.models.Work
+
+import scala.concurrent.Future
+
+class IdEmbedder @Inject()(metricsSender: MetricsSender,
```
kenoir: Would this be better off as a trait?

id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/IdentifierGenerator.scala
```diff
@@ -1,73 +1,76 @@
...
       .map { _ =>
         canonicalId
       }
   }
+
+  private def findIdentifierWith(identifiers: List[SourceIdentifier],
+                                 identifierScheme: String): String = {
+    identifiers
+      .find(identifier => identifier.identifierScheme == identifierScheme)
+      .fold[String](null)(identifier => identifier.value)
```
kenoir: Is `null` what we want here? Can we use `None`?

id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/IdEmbedder.scala
alicefuzier: It could be a trait but I'm not sure if it would be better. I think it would be kind of equivalent so I don't really have a preference. I did it this way 'cause it was what came more naturally. But I think that this class might eventually swallow the class that's using it, so I think we can leave it as is for now

id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/IdentifierGenerator.scala
alicefuzier: Yeah.. this is a slightly unfortunate thing about how the table is defined for scalalikejdbc..

alicefuzier:
Don't merge, I forgot to change the config files with the new flag

PR #883, opened 2017-09-04
Push even more of our build process inside Docker 🐳

alexwlchan:

Pushing an image to ECR and a release ID to S3 (“publishing a service”) now takes place entirely inside Docker.

### Who is this change for?

Who *isn’t* this for?


alexwlchan:
This patch is getting a bit unwieldy, I’ll probably carve chunks off and post them for individual review once I’ve got these sbt bits working.

PR #884, opened 2017-09-04
[WIP] Use Lenses in Id_minter

alicefuzier:
Reduce code that we need to write when adding stuff that needs to be identified in our data model
### Who is this change for?
Dev

- [ ] Deployed new versions


alicefuzier:
Closing this one in favor of #886

PR #890, opened 2017-09-06
Remove our third-party Lambda dependencies

alexwlchan:

Reduce the dependencies required to install our Lambdas.

### Who is this change for?

Developers who want a simpler build system.

- [ ] Deployed new versions

### DIFF
--- a/lambdas/common/setup.py
+++ b/lambdas/common/setup.py
@@ -13,9 +13,8 @@ setup(
     url='https://github.com/wellcometrust/platform-api/lambdas/common',
     python_requires='>=3.6',
     install_requires=[
-        'simplejson==3.11.1',
-        'six==1.10.0',
-        'structlog==17.2.0',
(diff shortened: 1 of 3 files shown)


lambdas/common/setup.py
```diff
@@ -13,9 +13,8 @@
     url='https://github.com/wellcometrust/platform-api/lambdas/common',
     python_requires='>=3.6',
     install_requires=[
-        'simplejson==3.11.1',
-        'six==1.10.0',
-        'structlog==17.2.0',
-        'boto3>=1.4.5'
+        # Strictly speaking, boto3 is a dependency, but it's pre-installed
+        # in AWS Lambda so we omit it here.
+        # 'boto3>=1.4.5',
```
alexwlchan: The Python 3.6 CI image we use installs boto3 as well.

PR #891, opened 2017-09-06
XML to JSON converter: Miro Preprocessor Pipeline Part 1

alexwlchan:
First part of adding the `Miro Preprocessor Pipeline`.  

These changes add a task which converts Miro XML from an S3 object to JSON and then places it back in S3. This should have parity with the first part of the existing Miro Adapter.


miro_preprocessor/Makefile
```diff
@@ -0,0 +1,53 @@
...
+xml_to_json_converter-test: $(ROOT)/.docker/python3.6_ci
+	docker run \
+		-v $(ROOT)/miro_preprocessor/xml_to_json_converter:/data \
+		-e OP=test -e FIND_MATCH_PATHS='/data' \
+		python3.6_ci:latest
+
+xml_to_json_converter-publish: xml_to_json_converter-build $(ROOT)/.docker/publish_service_to_aws
+	$(ROOT)/scripts/run_docker_with_aws_credentials.sh \
+	    -v /var/run/docker.sock:/var/run/docker.sock \
+		-v $(ROOT):/repo \
```
kenoir: Here we need docker in docker & AWS credentials - maybe we should thinking about changing this helper scripts so it's possible to compose requirements.

alexwlchan: For sure. I have some ideas about how we’d do this, but that’ll be a separate patch.

miro_preprocessor/Makefile
```diff
@@ -0,0 +1,53 @@
...
+	include $(ROOT)/shared.Makefile
+endif
+
+
+xml_to_json_converter-build: xml_to_json_converter-test $(ROOT)/.docker/image_builder
+	$(ROOT)/scripts/run_docker_in_docker.sh \
+		-v $(ROOT):/repo \
+		image_builder:latest \
+		--project=xml_to_json_converter \
+		--file=/repo/miro_preprocessor/xml_to_json_converter.Dockerfile
```
alexwlchan: Minor: I don’t use the `/repo/` from this argument, because the working directory is `/repo/`, and that makes the command continue to work if we change the mount point in the container.

Not a massive issue tho.

miro_preprocessor/Makefile
```diff
@@ -0,0 +1,53 @@
...
+miro_preprocessor-build: xml_to_json_converter-build
+
+miro_preprocessor-publish: xml_to_json_converter-publish
+
+miro_preprocessor-terraform-plan: $(ROOT)/.docker/terraform_ci
+	$(ROOT)/scripts/run_docker_with_aws_credentials.sh \
+		-v $(ROOT)/miro_preprocessor:/data \
+		-v $(ROOT)/terraform:/terraform \
+		-e OP=plan \
+		terraform_ci:latest
```
alexwlchan: While we’re thinking about changing the helper scripts, sorting out the `terraform` stuff might be useful as well.

miro_preprocessor/Makefile
```diff
@@ -0,0 +1,53 @@
...
+
+miro_preprocessor-terraform-apply: $(ROOT)/.docker/terraform_ci
+	$(ROOT)/scripts/run_docker_with_aws_credentials.sh \
+	-v $(ROOT)/miro_preprocessor:/data \
+	-v $(ROOT)/terraform:/terraform \
+	-e OP=apply \
+	terraform_ci:latest
+
+miro_preprocessor-run:
+	echo "Not implemented!"
```
alexwlchan: Can we get this task to fail as well?

miro_preprocessor/README.md
```diff
@@ -0,0 +1,7 @@
+# miro_preprocessor
+
+This stack is responsible for sorting Miro metadata and assets into one of three piles:
+
+*   Cold store / Amazon Glacier
+*   Tandem Vault
+*   The Catalogue API
```
alexwlchan: I think a brief explanation of the different pieces might be useful.

e.g. What’s the role of `xml_to_json_converter`?

miro_preprocessor/generate_tfvars.sh
```diff
@@ -0,0 +1,22 @@
...
+aws s3 cp s3://platform-infra/releases releases --recursive
+
+# Build a tfvars file containing the release ids
+echo "release_ids = {" >> "$TF_VARS"
+for f in releases/*;
+do
+  echo "Processing $f: $(cat $f)"
+  echo "  $(basename $f) = \"$(cat $f)\"" >> "$TF_VARS"
+done
+echo "}" >> "$TF_VARS"
```
alexwlchan: Ick. This script now appears three times in the repo – in `shared_infra`, and in `lambdas`. We should put a single, canonical copy in the `terraform` directory.

miro_preprocessor/main.tf
```diff
@@ -0,0 +1,44 @@
+module "xml_to_json_converter" {
+  source        = "../terraform/ecs_script_task"
+  task_name     = "xml_to_json_converter"
+  app_uri       = "${module.ecr_repository_xml_to_json_converter.repository_url}:${var.release_ids["xml_to_json_converter"]}"
+  task_role_arn = "${module.ecs_xml_to_json_converter_iam.task_role_arn}"
+
+  # This script has to load the XML files into memory, so make sure it
+  # has plenty of overhead.
+  memory = 2000
```
alexwlchan: If we’re doing stream parsing, is that true any more?

miro_preprocessor/main.tf
```diff
@@ -0,0 +1,44 @@
...
+  role   = "${module.ecs_xml_to_json_converter_iam.task_role_name}"
+  policy = "${data.aws_iam_policy_document.s3_read_miro_data.json}"
+}
+
+data "aws_iam_policy_document" "s3_read_miro_data" {
+  statement {
+    actions = [
+      "s3:Get*",
+      "s3:List*",
+      "s3:Put*",
```
alexwlchan: Minor: the XML-to-JSON converter reads files from one location, but writes to another.

Is it worth giving it permissions for just those locations, not the whole bucket?

miro_preprocessor/provider.tf
```diff
@@ -0,0 +1,8 @@
+# Specify the provider and access details
+provider "aws" {
+  region = "${var.aws_region}"
+
+  version = "0.1.4"
+}
+
+data "aws_caller_identity" "current" {}
```
alexwlchan: Do we need that? I think that was just needed for SNS topics in the shared infra pot.

miro_preprocessor/xml_to_json_converter/install_deps.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+pip install -r requirements.txt
```
alexwlchan: This script is never used.

miro_preprocessor/xml_to_json_converter/requirements.txt
```diff
@@ -0,0 +1,11 @@
+boto3==1.4.4
+botocore==1.5.46          # via boto3, s3transfer
+docopt==0.6.2
+docutils==0.13.1          # via botocore
+futures==3.1.1            # via s3transfer
+jmespath==0.9.2           # via boto3, botocore
+lazyreader==1.0.1
```
alexwlchan: 🎉

miro_preprocessor/xml_to_json_converter/run.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python
+# -*- encoding: utf-8 -*-
+"""
+Parse image records from a Miro export and push JSON derivatives to S3.
+
+The output is a file stored in S3 which encodes Miro records as JSON objects,
+one object per line.  This makes it easier for us to run one-off scripts
```
alexwlchan: I thought we were writing individual JSON files as well?

miro_preprocessor/xml_to_json_converter.Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+# Required for lxml
+RUN apk add build-base libxml2-dev libxslt-dev python3 python3-dev
+
+COPY xml_to_json_converter/requirements.txt /requirements.txt
+RUN pip3 install -r requirements.txt
+
+COPY xml_to_json_converter /app
+
+ENTRYPOINT ["/app/run.py"]
```
alexwlchan: Minor: I’d love us to get out of the habit of giving these scripts such generic names.

`run_xml_to_json_converter.py`?

miro_preprocessor/xml_to_json_converter.Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM python:3-alpine3.6
```
alexwlchan: Minor: either `FROM alpine` and install python3 manually, or `FROM python:3-alpine3.6` and you can drop the explicit `apk add python3` – but not both.

miro_preprocessor/xml_to_json_converter/run.py
```diff
@@ -0,0 +1,60 @@
...
+    )
+
+    os.unlink(tmp_json)
+
+
+if __name__ == '__main__':
+    args = docopt.docopt(__doc__)
+    bucket = args['--bucket']
+    source_key = args['--src']
+    destination_key = args['--dst']
```
alexwlchan: nit: I often use `src_key` and `dst_key`, so alignment works and things match

miro_preprocessor/xml_to_json_converter/test_run.py
```diff
@@ -0,0 +1,71 @@
+# -*- encoding: utf-8 -*-
+
+import pytest
```
alexwlchan: nit: pytest is a third-party import

miro_preprocessor/xml_to_json_converter/test_utils.py
```diff
@@ -0,0 +1,140 @@
...
+    # Dict-like structure with non-unique keys
+    '''
+    <apple>
+        <variety>braeburn</variety>
+        <variety>cox</variety>
+    </apple>
+    ''',
+])
+def test_bad_elem_throws_assert_error(xml_string):
+    """Parsing malformed XML strings throws an AssertionError."""
```
alexwlchan: Now I see this written out, I realise that we should really be throwing a more specific `InvalidMiroXMLError` exception, not just asserting. Mea culpa.

shared.Makefile
```diff
@@ -44,4 +43,5 @@ $(ROOT)/.docker/miro_adapter_tests:
 		--file=miro_adapter/miro_adapter_tests.Dockerfile
 
 clean:
-	rm -rf $(ROOT)/.docker
\ No newline at end of file
+	rm -rf $(ROOT)/.docker
+
```
alexwlchan: nit: lose the newline

miro_preprocessor/xml_to_json_converter/test_utils.py
```diff
@@ -0,0 +1,140 @@
...
+    elem = etree.fromstring(xml_string)
+    with pytest.raises(AssertionError):
+        elem_to_dict(elem)
```
alexwlchan: I’d expect linting to whinge about this – Python files should end on a newline. Have I got my lints wrong, or is the linter broken? 😞

miro_preprocessor/xml_to_json_converter/utils.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def read_json_lines_from_s3(bucket, key):
+    """
+    Read a document which contains one JSON document per line, and parse the
+    JSON before passing to the caller.
+    """
+    for doc in chunked_s3_reader(bucket=bucket, key=key, delimiter=b'\n'):
+        if doc:
+            yield json.loads(doc)
```
alexwlchan: 👍 nice use of my library

miro_preprocessor/xml_to_json_converter/run.py
```diff
@@ -0,0 +1,60 @@
...
+    image_data = generate_images(bucket=bucket, key=source_key)
+
+    tmp_json = tempfile.mktemp()
+    os.makedirs(os.path.dirname(tmp_json), exist_ok=True)
+    with open(tmp_json, 'w') as f:
+        for img in image_data:
+
+            # Adding the separators omits unneeded whitespace in the JSON,
+            # giving us smaller files.
+            f.write(json.dumps(img, separators=(',', ':')) + '\n')
```
alexwlchan: If you extract the JSON body as a variable, you can push that straight to S3 without waiting for this file to finish writing.

miro_preprocessor/generate_tfvars.sh
kenoir: Each project _might_ do this differently (and we do actually). We could extract some of the functionality elsewhere but the scripts are acting as "hooks" to allow each service to implement this however necessary.

miro_preprocessor/xml_to_json_converter/install_deps.sh
kenoir: It's used by the test operation in the python container - this is again optionally present to allow the project to trigger installing dependencies for test and build in whatever way is necessary.

This and the `generate_tf_vars` hook should probably be documented somewhere.

miro_preprocessor/main.tf
kenoir: In order to write the whole JSON file at some point _that_ needs to be accumulated in memory - so while the stream parsing will help I think the memory requirements will be similar.

miro_preprocessor/xml_to_json_converter/run.py
kenoir: Yes, but not in this PR.

miro_preprocessor/xml_to_json_converter/run.py
kenoir: This is a straight copy of (half) of what happens right now, just to get it into another project. Further PRs will implement changes.

miro_preprocessor/xml_to_json_converter.Dockerfile
kenoir: This is part of a larger app contained in this folder (and is not the only part).

Its name denotes its part of the app.

miro_preprocessor/xml_to_json_converter/test_utils.py
kenoir: This seems to be testing an external lib? Not sure how to resolve that issue in this context.

miro_preprocessor/xml_to_json_converter/test_utils.py
kenoir: It was - I fixed it.

kenoir:
@alexwlchan ready for another look now I think.

docker/python3.6_ci/run.sh
```diff
@@ -18,7 +18,7 @@ function install_dependencies {
 function lint_python {
   echo "Linting ..."
 
-  flake8 --exclude six.py,six-*,structlog*,simplejson* **/*.py --ignore=E501
+  flake8 --exclude target  --ignore=E501
```
alexwlchan: 👍

miro_preprocessor/Makefile
alexwlchan: TODO: Open an issue to track fixing this

miro_preprocessor/xml_to_json_converter/test_utils.py
alexwlchan: TODO: Fix this (Alex)

PR #895, opened 2017-09-06
Promote Loris to a top level project

alexwlchan:

Start rearranging the repository. In particular:

* `loris` moves from inside the `docker` directory into a toplevel project, and gains its own Makefile and README
* Starting to pull out some core builder tasks from the main Makefile, so they’re accessible to Loris
* Rename `loris-deploy` to `loris-publish`.
* Firm up a bunch of assumptions about what we’re sharing into Docker containers: if we want to share the whole repo, really share the whole repo – not just the current working directory

Still to do:

* Move in a Terraform stack (large enough to merit a separate patch)
* ~Get a `loris-run` command working for local dev~

Questions for @alicefuzier and @kenoir:

* Is the new Make layout clear? Potentially a bit more magical than before.
* What, if anything, do we want in the README beyond the current stub?
* Does this patch make sense?

### Who is this change for?

A Platform team who want a more sensible repository.


Makefile
```diff
@@ -1,4 +1,5 @@
-export INFRA_BUCKET = platform-infra
+include builds.Makefile
```
alexwlchan: shared? root?

loris/Makefile
```diff
@@ -0,0 +1,22 @@
...
+
+ifneq ($(ROOT), $(shell pwd))
+	include $(ROOT)/builds.Makefile
+endif
+
+
+loris-build: $(ROOT)/.docker/image_builder
+	PROJECT=loris FILE=loris/Dockerfile $(ROOT)/builds/build_image.sh
+
+loris-serve: loris-build
```
kenoir: Maybe `loris-run`

alexwlchan:
I’ll merge once this passes tests.

PR #897, opened 2017-09-06
Remove BaseLicense and make License a real enum

alicefuzier:
Licenses should be enums, so we can only create valid license instances
### Who is this change for?
Devs

- [ ] Deployed new versions

### DIFF
--- a/api/src/test/scala/uk/ac/wellcome/platform/api/models/DisplayItemTest.scala
+++ b/api/src/test/scala/uk/ac/wellcome/platform/api/models/DisplayItemTest.scala
@@ -10,10 +10,6 @@ class DisplayItemTest extends FunSpec with Matchers {
     val thumbnailUrl = Some("https://iiif.example.org/V0000001/default.jpg")
     val locationType = "thumbnail-image"
 
-    val licenseType = "CC-Test"
-    val licenseLabel = "A fictional license for testing"
-    val licenseUrl = "http://creativecommons.org/licenses/test/-1.0/"
-
     Location(
       locationType = locationType,
       url = thumbnailUrl,
--- a/api/src/test/scala/uk/ac/wellcome/platform/api/models/DisplayItemTest.scala
+++ b/api/src/test/scala/uk/ac/wellcome/platform/api/models/DisplayItemTest.scala
@@ -6,18 +6,6 @@ import uk.ac.wellcome.models._
 
 class DisplayItemTest extends FunSpec with Matchers {
 
-  val license: License = {
-    val licenseType = "CC-Test"
-    val label = "A fictional license for testing"
-    val url = "http://creativecommons.org/licenses/test/-1.0/"
-
-    License(
-      licenseType = licenseType,
-      label = label,
-      url = url
-    )
-  }
-
   val location: Location = {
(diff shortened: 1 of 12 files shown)


common/src/main/scala/uk/ac/wellcome/models/License.scala
```diff
@@ -1,48 +1,68 @@
...
+    createLicense(licenseType)
+  }
+
+  private def createLicense(licenseType: String): License = {
+    licenseType match {
+      case s: String if s == License_CCBY.licenseType => License_CCBY
+      case s: String if s == License_CCBYNC.licenseType => License_CCBYNC
+      case s: String if s == License_CCBYNCND.licenseType => License_CCBYNCND
+      case s: String if s == License_CC0.licenseType => License_CC0
+      case s: String if s == License_PDM.licenseType => License_PDM
```
alexwlchan: Would it be helpful to have a fallthrough case that logs out the `licenseType`?

api/src/test/scala/uk/ac/wellcome/platform/api/models/DisplayLocationTest.scala
```diff
@@ -8,27 +8,20 @@ class DisplayLocationTest extends FunSpec with Matchers {
...
 
     displayLocation.locationType shouldBe locationType
     displayLocation.url shouldBe thumbnailUrl
     displayLocation.license shouldBe DisplayLicense(
-      licenseType = licenseType,
-      label = licenseLabel,
-      url = licenseUrl
+      licenseType = License_CCBY.licenseType,
+      label = License_CCBY.label,
+      url = License_CCBY.url
```
alexwlchan: I think this could be even simpler:

```scala
displayLocation.license shouldBe DisplayLicense(internalLocation.license)
```

api/src/test/scala/uk/ac/wellcome/platform/api/models/DisplayItemTest.scala
```diff
@@ -29,7 +17,7 @@ class DisplayItemTest extends FunSpec with Matchers {
     Location(
       locationType = locationType,
       url = thumbnailUrl,
-      license = license
+      license = License_CCBY
```
alexwlchan: I think this is my bad – but while we’re here, can we scrub the `license{Type, Label, Url}` definitions on the lines above?

common/src/test/scala/uk/ac/wellcome/models/WorkTest.scala
```diff
@@ -220,7 +203,7 @@ class WorkTest extends FunSpec with Matchers {
     print(result)
 
     result.isSuccess shouldBe true
-    result.get shouldBe identifiedWorkJson
+    assertJsonStringsAreEqual(result.get, identifiedWorkJson)
```
alexwlchan: 👍

api/src/test/scala/uk/ac/wellcome/platform/api/models/DisplayItemTest.scala
alicefuzier: Not sure what you mean... Do you want to move the license json string building out?

PR #902, opened 2017-09-07
Move main stack into a separate own folder

kenoir:
🏭 🐌


shared_infra/releases/api
```diff
@@ -0,0 +1 @@
+0.0.1-9556f9000c69b435a320feea6f045a27057998fc_prod
```
kenoir: ruh roh

shared_infra/Makefile
```diff
@@ -0,0 +1,17 @@
+ROOT = $(shell git rev-parse --show-toplevel)
+
+ifneq ($(ROOT), $(shell pwd))
+include $(ROOT)/shared.Makefile
```
alexwlchan: nit: tab indent for readability

shared_infra/Makefile
```diff
@@ -0,0 +1,17 @@
...
+endif
+
+
+## Run a plan on main stack
+terraform-shared-plan: uptodate-git .docker/terraform_ci
+	docker run -v $$(pwd)/terraform:/terraform -v $$(pwd)/shared_infra:/data -v $$HOME/.aws:/root/.aws -v $$HOME/.ssh:/root/.ssh -e OP=plan terraform_ci:latest
```
alexwlchan: nit: some linebreaks for readability might be nice

shared_infra/Makefile
```diff
@@ -0,0 +1,17 @@
+ROOT = $(shell git rev-parse --show-toplevel)
+
+ifneq ($(ROOT), $(shell pwd))
+include $(ROOT)/shared.Makefile
+endif
+
+
+## Run a plan on main stack
+terraform-shared-plan: uptodate-git .docker/terraform_ci
```
alexwlchan: Because `uptodate-git` is still defined in the top-level Makefile, I don't think this will work if you're running inside the `shared_infra` directory. Is that intentional?

Ditto `.docker/terraform_ci`.

shared_infra/README.md
```diff
@@ -0,0 +1,7 @@
+# shared_infra
+
+The currently monolithic terraform stack for _most_ of the services we provide.
+
+## Overview
+
+Contains all the ingest pipeline services, the Catalogue API, the IIIF API (Loris) and attendant ECS Script Tasks.
```
alexwlchan: 👍 for "attendant"

terraform/cloudformation/asg.json.template
```diff
@@ -1,45 +0,0 @@
-{
-  "Resources": {
```
alexwlchan: Are we just not using this any more?

.gitignore
```diff
@@ -5,7 +5,7 @@
 .lambda_hashes.json
 terraform.plan
 .terraform
-terraform/releases
+shared_infra/releases
```
alexwlchan: Minor: maybe we could rename this directory to `.releases`?

Then it’s hidden from view, and we get the same effect if we use this pattern in other stacks.

terraform/cloudformation/asg.json.template
kenoir: Yup. It got moved inside another module and we didn't delete this one - should probably be another PR, but low hanging fruit.

Issue #734, opened 2017-08-01
Loris is screwing with colour profiles in images


alexwlchan:
https://iiif.wellcomecollection.org/image/V0000003.jpg/full/full/0/default.jpg

Left is the JPEG from miro-images-public; right is the image served by Loris. They don’t look the same. 😢 

![screen shot 2017-08-01 at 11 36 14](https://user-images.githubusercontent.com/301220/28821537-b351062e-76ad-11e7-9c80-2854ba6d8ee7.png)

jtweed:
Loris should be transforming to sRGB if necessary. There is some support colour management support in Loris already, we may need extend on that if it doesn't do what we need.

alexwlchan:
```
emb_profile = io.BytesIO(im2.info['icc_profile'])
im3 = profileToProfile(im2alt, emb_profile, '/System/Library/ColorSync/Profiles/sRGB Profile.icc')
im3.save('V0000003_srgb_sys.jpg')
```

alexwlchan:
I managed to get a WIP PR against Loris tonight. Still missing the necessary tests, and a bit messy, but basically works. \o/

alexwlchan:
I’m waiting for green CI and a review on https://github.com/loris-imageserver/loris/pull/351, then I should be able to finish the Loris PR and get it merged.

alexwlchan:
Okay, Loris patch is merged. We’ll need a config change and new Docker image on our side, then this should be good to go.

alexwlchan:
Patch is done and merged, we just need to deploy it. The Travis build failed, which I’m fixing in #907, meanwhile I’m just going to deploy from local, as Docker means it doesn’t make a difference!

alexwlchan:
Doing this uncovered _another_ Loris patch we needed to get this working with rotation, but I’ve deployed from my fork to get this done. We’ll need to flip back to Loris proper when it gets merged, but it should be the same code.

I’ve also flushed the CloudFront cache, so all (library) images now appear with their correct colour profiles!

Example URLs:

https://iiif-origin.wellcomecollection.org/image/V0000003.jpg/full/full/0/default.jpg
https://iiif-origin.wellcomecollection.org/image/V0000003-broken.jpg/full/full/0/default.jpg
https://iiif-origin.wellcomecollection.org/image/V0000003.jpg/full/full/45/default.jpg

PR #906, opened 2017-09-11
Update our version of Loris to do color profile conversion

alexwlchan:

This updates our version of Loris to include https://github.com/loris-imageserver/loris/pull/350, which adds support for sRGB colour conversion in JPEGs. It also includes an sRGB colour profile and the required config.

Related: #734.

### Who is this change for?

People who want good-looking images! 🖼 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


loris/Dockerfile
```diff
@@ -3,16 +3,18 @@ FROM ubuntu:14.04
...
 RUN pip install awscli
 
-ENV LORIS_COMMIT 400bc3979a7dc3ff938abff84665c2731f750a69
+ENV LORIS_COMMIT 1e5de1f8b4dedcd80506c1f531fc76caec8ad504
 
 COPY install_loris.sh /install_loris.sh
 RUN /install_loris.sh
 
-RUN apt-get install -y uwsgi
+# This is an sRGB color profile downloaded from
```
kenoir: 👍

loris/Dockerfile
```diff
@@ -3,16 +3,18 @@ FROM ubuntu:14.04
 RUN apt-get update
 
 RUN apt-get install -y python python-pip python-setuptools python-dev \
-    uwsgi-plugin-python
+    uwsgi uwsgi-plugin-python
```
kenoir: Why do we know include `uwsgi` by itself?

alexwlchan: We already did – this just puts our `apt-get install` commands on one line. Previously we were installing it after we’d installed Loris.

kenoir: Ah, I see it.

Issue #882, opened 2017-09-04
💥 Let’s break up the Miro adapter, and decide where images should live 💥


alexwlchan:
This is a tracking ticket for our conversations this morning, and all the bits that need to happen for this to be done:

- [x] Write an XML-to-JSON converter service, which takes XML Miro exports and breaks them up into individual JSON files
- [x] Write a sorter that spots changes to those files, and decides the final category for each Miro record
- [x] Write a Catalogue adapter that spots the category assignment, and copies metadata and assets accordingly

alexwlchan:
Ah, I remembered why it’s useful to have a single file with all the JSON blobs!

When you want to do analysis of the entire set (e.g. “if I make this change, how many records will be dropped”?) that’s _much_ easier if you can download a single file and work locally than iterate over thousands of files that you’ve had to fetch from S3.

alexwlchan:
The WI selection rules document is on SharePoint: https://wellcomecloud-my.sharepoint.com/personal/c_henshaw_wellcome_ac_uk/_layouts/15/onedrive.aspx?id=%2fpersonal%2fc_henshaw_wellcome_ac_uk%2fDocuments%2fShared+stuff+-+internal%2fWI+migration&FolderCTID=0x012000C157F1E960030E429A435A8C372025FF

Also, feedback from Christy:

> In the same folder as the WI selection rules document is a text file with a list of image numbers. These are images with the phrase “title page” in the description, slightly edited to remove a few images that we might want to keep for whatever reason.
>
> This list is for total obliteration – don’t need the images, don’t need the MIRO records, even.

alexwlchan:
Relevant: #711

kenoir:
Done

kenoir:
Closing as done.

PR #908, opened 2017-09-12
Add multiple json files from miro XML

kenoir:

Create a single JSON file for each Miro "image" processed in order to enable processing on that image via lambda in a distributed fashion.

### Who is this change for?

💻 

- [ ] Deployed new versions

### DIFF
--- a/miro_preprocessor/xml_to_json_converter/run.py
+++ b/miro_preprocessor/xml_to_json_converter/run.py
@@ -30,19 +30,32 @@ import docopt
 from utils import generate_images
 
 
-def main(bucket, src_key, dst_key):
+def main(bucket, src_key, dst_key, js_path="json"):
     image_data = generate_images(bucket=bucket, key=src_key)
 
     tmp_json = tempfile.mktemp()
     os.makedirs(os.path.dirname(tmp_json), exist_ok=True)
+
+    s3 = boto3.client('s3')
+
     with open(tmp_json, 'w') as f:
         for img in image_data:
+            img_json_dump = json.dumps(img, separators=(',', ':'))
+
+            image_id = img["image_no_calc"]
+            json_object_key = f'{js_path}/{image_id}.json'
+            byte_encoded_json_dump = img_json_dump.encode()
+
+            s3.put_object(
+                Bucket=bucket,
+                Key=json_object_key,
+                Body=byte_encoded_json_dump
+            )
 
             # Adding the separators omits unneeded whitespace in the JSON,
             # giving us smaller files.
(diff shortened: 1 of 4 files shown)


kenoir:
Part of: https://github.com/wellcometrust/platform/issues/882

miro_preprocessor/xml_to_json_converter/test_run.py
```diff
@@ -55,15 +67,65 @@ def test_creates_txt_with_all_images_json(set_region):
...
+
+    response = s3_client.list_objects(
+        Bucket=bucket,
+        Prefix=prefix
+    )
+
+    keys = [obj["Key"] for obj in response["Contents"]]
+
+    for key, value in expected_json_objects.items():
+        assert key in keys
```
alexwlchan: I think we want an additional test that `keys` only contains what we expect. Also, this fails on the first mismatched key it finds – better would be to read the complete set of JSON files from S3, and compare them all together.

miro_preprocessor/xml_to_json_converter/test_run.py
```diff
@@ -55,15 +67,65 @@ def test_creates_txt_with_all_images_json(set_region):
...
+
+        get_object_response = s3_client.get_object(
+            Bucket=bucket,
+            Key=key
+        )
```
alexwlchan: nit: this and the S3 call above can probably be collapsed onto a single line

miro_preprocessor/xml_to_json_converter/run.py
```diff
@@ -30,19 +30,32 @@
...
     image_data = generate_images(bucket=bucket, key=src_key)
 
     tmp_json = tempfile.mktemp()
     os.makedirs(os.path.dirname(tmp_json), exist_ok=True)
+
+    s3 = boto3.client('s3')
+
     with open(tmp_json, 'w') as f:
         for img in image_data:
+            img_json_dump = json.dumps(img, separators=(',', ':'))
```
alexwlchan: Because we’re making assertions about the exact output in the tests, I’d add `sort_keys=True` to the parameters to `json.dumps()` – that ensures we have a stable sort order in the JSON. Otherwise you’re relying on dict iteration order, which isn’t guaranteed..

miro_preprocessor/xml_to_json_converter/run.py
```diff
@@ -30,19 +30,32 @@
 from utils import generate_images
 
 
-def main(bucket, src_key, dst_key):
+def main(bucket, src_key, dst_key, js_path="json"):
```
alexwlchan: Nice idea to make the prefix configurable. 👍

PR #911, opened 2017-09-12
Add Lambda to run ECS task from SNS trigger

kenoir:

Add a lambda that can start ECS tasks when triggered from SNS. This constitutes part of the pipeline for running a task to convert Miro XML output to JSON.

### Who is this change for?

💻 

- [ ] Run `terraform apply`.

### DIFF
--- a/lambdas/common/utils/ecs_utils.py
+++ b/lambdas/common/utils/ecs_utils.py
@@ -157,3 +157,32 @@ def describe_service(ecs_client, cluster_arn, service_arn):
         cluster=_name_from_arn(cluster_arn),
         services=[_name_from_arn(service_arn)]
     )['services'][0]
+
+
+def run_task(
+        ecs_client,
+        cluster_name,
+        task_definition,
+        started_by,
+        container_name="app",
+        command=[]):
+    """
+    Run a given command against a named container in a task definition on a particular cluster.
+
+    Returns the response from calling run_task
+    """
+    return ecs_client.run_task(
+        cluster=cluster_name,
+        taskDefinition=task_definition,
+        overrides={
+            'containerOverrides': [
+                {
+                    'name': container_name,
+                    'command': command
+                },
+            ],
+            'taskRoleArn': 'string'
+        },
+        count=1,
+        startedBy=started_by,
+    )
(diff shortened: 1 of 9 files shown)


kenoir:
Part of: https://github.com/wellcometrust/platform/issues/882

lambdas/common/utils/ecs_utils.py
```diff
@@ -157,3 +157,32 @@ def describe_service(ecs_client, cluster_arn, service_arn):
...
     )['services'][0]
+
+
+def run_task(
+        ecs_client,
+        cluster_name,
+        task_definition,
+        started_by,
+        container_name="app",
+        command=[]):
```
alexwlchan: I’d suggest adding a `overrides` parameter to this function, and adding its value to the existing `overrides` in the RunTask API – some tasks require a bit of configuration (e.g. environment variables), and that’s quite a flexible way to do it.

lambdas/run_ecs_task/src/run_ecs_task.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python
+# -*- encoding: utf-8 -*-
+"""
+Run an
```
alexwlchan: "run an… ?"

lambdas/run_ecs_task/src/run_ecs_task.py
```diff
@@ -0,0 +1,34 @@
...
+        ecs_client=boto3.client('ecs'),
+        cluster_name=message_data['cluster_name'],
+        container_name=message_data['container_name'],
+        task_definition=message_data['task_definition'],
+        command=message_data['command'],
+        started_by=message_data['started_by']
+    )
+
+    print(f'response = {response!r}')
+    assert len(response['failures']) == 0
```
alexwlchan: Maybe assert the number of tasks or pending tasks is 1 as well?

lambdas/common/utils/ecs_utils.py
kenoir: I think there's possibly utility in keeping the interface as simple as possible here. I'd prefer to update this as required.

PR #912, opened 2017-09-13
Miro preprocessor Part 3

alicefuzier:
Create lambda that reads a image json from sns and sends it to the MiroData table in DynamoDB
### Who is this change for?
Everyone

- [ ] Deployed new versions
- [x] Run `terraform apply`.


miro_preprocessor/Makefile
```diff
@@ -26,13 +25,31 @@ xml_to_json_converter-publish: xml_to_json_converter-build $(ROOT)/.docker/publi
...
+xml_to_json_run_task-test: $(ROOT)/.docker/python3.6_ci
+	docker run \
+		-v $(ROOT)/miro_preprocessor/xml_to_json_run_task:/data \
+		-e OP=test -e FIND_MATCH_PATHS='/data/src' \
+		python3.6_ci:latest
+miro_image_to_dynamo-test: $(ROOT)/.docker/python3.6_ci
+	$(ROOT)/scripts/run_docker_with_aws_credentials.sh \
+	-v $(ROOT)/miro_preprocessor/miro_image_to_dynamo:/data \
+	-e OP=test \
+	-e FIND_MATCH_PATHS='./*/target common/tests' \
```
kenoir: Why `target`?

miro_preprocessor/Makefile
```diff
@@ -26,13 +25,31 @@ xml_to_json_converter-publish: xml_to_json_converter-build $(ROOT)/.docker/publi
...
+	python3.6_ci:latest
+
+xml_to_json_run_task-build: $(ROOT)/.docker/python3.6_ci
```
kenoir: Probably worth adding a build task for your lambda too.

miro_preprocessor/miro_image_to_dynamo/main.tf
```diff
@@ -0,0 +1,46 @@
...
+    actions = [
+      "dynamodb:PutItem",
+    ]
+
+    resources = [
+      "${var.miro_data_table_arn}",
+    ]
+  }
+
+  statement {
```
kenoir: I don't think you need this. You can pass an existing policy document down from the root.

miro_preprocessor/miro_image_to_dynamo/src/miro_image_to_dynamo.py
```diff
@@ -0,0 +1,54 @@
...
+    try:
+        s3_client.head_object(Bucket=bucket_name, Key=key)
+    except ClientError as client_error:
+        if client_error.response['Error']['Code'] == '404':
+            print(f"No image found for MiroId {miro_id}: skipping")
+            pass
+        else:
+            raise
+    else:
+        print(f"Image found for MiroId {miro_id}: sending to Dynamodb")
```
kenoir: Maybe you could split some of these control paths into functions?

miro_preprocessor/miro_image_to_dynamo/src/miro_image_to_dynamo.py
```diff
@@ -0,0 +1,54 @@
...
+        table_name = os.environ["TABLE_NAME"]
+        print('Pushing image with ID %s' % (miro_id))
```
kenoir: I think we need to move the image into the public folder (or at least signal for another process to do so here).

kenoir: Ping @alexwlchan to confirm.

miro_preprocessor/miro_image_to_dynamo/main.tf
alicefuzier: Yeah, but there is no policy document that matches the permissions that this lambda needs, I still have to create it. I think I prefer if it lives in the same file as the lambda module that uses it (unless it's a generic thing that can be reused) so that we don't have to follow too many links to see what is going on

miro_preprocessor/miro_image_to_dynamo/src/miro_image_to_dynamo.py
alicefuzier: 👍

miro_preprocessor/miro_image_to_dynamo/src/miro_image_to_dynamo.py
alicefuzier: right... I wonder whether we should split this into two lambdas then. One that receives the json from SNS, checks if the file exists and, if does, copies the file in the public bucket and forwards the json into another SNS topic. 
The lambda that pushes to dynamo listens to this topic, so we always know that anything that is pushed to dynamo has an image in public bucket (i.e if copying fails, the json is never forwarded and we don't show broken images in the API)

PR #913, opened 2017-09-13
XML to JSON Run ECS Task lambda

kenoir:

Depends on: https://github.com/wellcometrust/platform/pull/911

Runs the XML to JSON ECS task from an S3 event stream when a new XML file is uploaded.

### Who is this change for?

💻 

- [ ] Run `terraform apply`.


api/src/main/scala/uk/ac/wellcome/platform/api/finatra/exceptions/CaseClassMappingExceptionMapper.scala
```diff
@@ -24,13 +24,15 @@ class CaseClassMappingExceptionWrapper @Inject()(
 
   val contextUri: String = s"${apiScheme}://${apiHost}${apiContext}"
 
-  override def toResponse(request: Request, e: CaseClassMappingException): Response = {
+  override def toResponse(request: Request,
```
kenoir: `scalafmt`!

alexwlchan: Fixing it!

alexwlchan: (And working out why it didn't run on my PR)

docker/python3.6_ci/build_lambda.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+
+if [ -d "src" ]
```
alexwlchan: nit: I’d pull the `lambda_dir` variable out of the `if` block, then reference it here:

```bash
if [[ -d "$lambda_dir" ]]
```

docker/python3.6_ci/Dockerfile
```diff
@@ -12,6 +12,7 @@ RUN pip install moto==1.0.1
 
 COPY is_up_to_date_with_master.py /app/is_up_to_date_with_master.py
 
+COPY build_lambda.sh /app/build_lambda.sh
```
alexwlchan: thought: does `COPY . /app` work?

Might save us having to edit the Dockerfile when we add new scripts.

miro_preprocessor/Makefile
```diff
@@ -26,13 +26,25 @@ xml_to_json_converter-publish: xml_to_json_converter-build $(ROOT)/.docker/publi
 		--project=xml_to_json_converter \
 		--infra-bucket=platform-infra
 
-miro_preprocessor-test: xml_to_json_converter-test
+xml_to_json_run_task-test: $(ROOT)/.docker/python3.6_ci
+	docker run \
+		-v $(ROOT)/miro_preprocessor/xml_to_json_run_task:/data \
+		-e OP=test -e FIND_MATCH_PATHS='/data/src' \
```
alexwlchan: For consistency with `xml_to_json_converter-test`, do we want these paths to be:

```
-v $(ROOT)/miro_preprocessor/xml_to_json_run_task/src:/data \
-e OP=test -e FIND_MATCH_PATHS='/data'
```

?

alexwlchan: Also, I wonder if it would be helpful to add a second variable to our Makefiles:

```make
ROOT = $(shell git rev-parse --show-toplevel)
PROJECT = $(ROOT)/miro_preprocessor
```

miro_preprocessor/iam_policy_document.tf
```diff
@@ -0,0 +1,24 @@
...
+  statement {
+    actions = [
+      "s3:Put*",
+    ]
+
+    resources = [
+      "${data.terraform_remote_state.platform.bucket_miro_data_arn}/json",
+    ]
+  }
+}
```
alexwlchan: 👌 

Nice separation of privileges.

miro_preprocessor/xml_to_json_run_task/main.tf
```diff
@@ -0,0 +1,21 @@
...
+}
+
+module "trigger_application_restart_on_config_change" {
+  source = "../../terraform/lambda/trigger_s3"
+
+  lambda_function_name = "${module.lambda_xml_to_json_run_task.function_name}"
+  lambda_function_arn  = "${module.lambda_xml_to_json_run_task.arn}"
+  s3_bucket_arn        = "${var.bucket_miro_data_arn}"
+  s3_bucket_id         = "${var.bucket_miro_data_id}"
+  filter_prefix        = "xml"
```
alexwlchan: So this has a filter prefix of `xml`, but the role policy has read permissions for a prefix of `source`.

Are they intentionally different? Could we get a brief explanation in the README of the different folders, maybe?

miro_preprocessor/xml_to_json_run_task/main.tf
```diff
@@ -0,0 +1,21 @@
+# Lambda for publishing ECS service schedules to an SNS topic
+
+module "lambda_xml_to_json_run_task" {
+  source     = "../../terraform/lambda"
+  source_dir = "${path.module}/target"
+
+  name        = "xml_to_json_run_task"
+  description = "Run the Miro XML to JSON ECS task"
```
alexwlchan: nit: The more I read it, the more this description bugs me. Maybe "*Start* the Miro XML to JSON ECS task"? Just to make it clear this Lambda isn’t actually doing the work.

miro_preprocessor/xml_to_json_run_task/src/conftest.py
```diff
@@ -0,0 +1,53 @@
+import boto3
```
alexwlchan: Not to be addressed right now, but we should really find a way to stop just copying these files around.

miro_preprocessor/xml_to_json_run_task/src/xml_to_json_run_task.py
```diff
@@ -0,0 +1,61 @@
...
+
+def command_for_xml_to_json_task(event, cluster_name, container_name, task_definition):
+    bucket_name = event["bucket_name"]
+    object_src = event["object_key"]
+    object_dst = f'{object_src}.txt'
+
+    cmd = [
+        f'--bucket={bucket_name}',
+        f'--src={object_src}',
+        f'--src={object_dst}'
```
alexwlchan: Should this be `--dst`?

miro_preprocessor/xml_to_json_run_task/src/xml_to_json_run_task.py
```diff
@@ -0,0 +1,61 @@
...
+        sns_client,
+        topic_arn,
+        cmd
+    )
+
+
+def main(event, _):
+    print(f'event = {event!r}')
+
+    topic_arn = os.environ["TOPIC_ARN"]
```
alexwlchan: How do these environment variables end up in the Lambda? I don’t see any associated Terraform config, or have I missed it?

miro_preprocessor/xml_to_json_run_task/src/xml_to_json_run_task.py
```diff
@@ -0,0 +1,61 @@
+# -*- encoding: utf-8 -*-
+
```
alexwlchan: nit: lose this newline

miro_preprocessor/xml_to_json_run_task/src/xml_to_json_run_task.py
```diff
@@ -0,0 +1,61 @@
...
+"""
+Lambda to run the Miro XML to JSON task when receiving an S3 event.
```
alexwlchan: minor: I’d rephrase this to make it clear "run the task" means a specific AWS API:

> Lambda to call the RunTask API for the Miro XML-to-JSON ECS task when a file in S3 is updated.

miro_preprocessor/xml_to_json_run_task/src/xml_to_json_run_task.py
```diff
@@ -0,0 +1,61 @@
...
+
+    s3_events = s3_utils.parse_s3_record(event=event)
+    task_commands = [command_for_xml_to_json_task(
+        event,
+        cluster_name,
+        container_name,
+        task_definition
+    ) for event in s3_events]
+
+    [post_to_sns(sns_client, topic_arn, cmd) for cmd in task_commands]
```
alexwlchan: Using list comprehensions for side effects in this way is unusual. You’re collecting the results even though you don’t plan to use it – here, a `for` loop is more common, i.e.,

```python
for cmd in task_commands:
    post_to_sns(sns_client, topic_arn, cmd)
```

which makes it explicit that we’re discarding the results.

alexwlchan: And at this point, I’d drop the `post_to_sns` wrapper, and call `sns_utils.publish_message()` directly. Cut out an unnecessary redirection.

miro_preprocessor/xml_to_json_run_task/src/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+def moto_start(set_region):
+    mock_autoscaling().start()
+    mock_ec2().start()
+    mock_ecs().start()
+    mock_sns().start()
+    mock_sqs().start()
+    yield
+    mock_autoscaling().stop()
+    mock_ec2().stop()
+    mock_ecs().stop()
```
alicefuzier: don't think you need ```
mock_autoscaling()
mock_ec2()
mock_ecs()
```

File: miro_preprocessor/terraform.tf [hunk 15]
```diff
@@ -18,3 +18,13 @@ data "terraform_remote_state" "platform" {
     region = "eu-west-1"
   }
 }
+
+data "terraform_remote_state" "lambda" {
```
alicefuzier: This is only here beacuse of the `lambda_error_alarm_arn` output. I wonder whether this should be in the shared infra instead

miro_preprocessor/Makefile
kenoir: Wouldn't this cause an issue if multiple projects used this variable?

alexwlchan: Mmm. I think it would still work – `include` just runs the Makefile straight through, so it would only last until the variable was next defined. But might get messy, so let’s leave it for now.

miro_preprocessor/terraform.tf
kenoir: 👍  I think we can consolidate this after.

miro_preprocessor/xml_to_json_run_task/src/xml_to_json_run_task.py
kenoir: Yeah - good point, have added these.

docker/python3.6_ci/Dockerfile
```diff
@@ -2,18 +2,9 @@ FROM python:3.6
...
-
-RUN pip install pytest
-
-RUN pip install moto==1.0.1
-
-COPY is_up_to_date_with_master.py /app/is_up_to_date_with_master.py
-
-COPY run.sh /app/run.sh
-COPY test.sh /app/test.sh
+RUN pip install -r /app/requirements.txt
```
alexwlchan: 👍 much nicer

docker/python3.6_ci/src/build_lambda.sh
```diff
@@ -0,0 +1,27 @@
...
+
+  if [ -e src/requirements.txt ]
+  then
+    echo "Found requirements.txt, installing."
+    pip install -r src/requirements.txt --target=$target_dir
+  else
+    echo "No requirements.txt present. Skipping."
+  fi
+else
+  echo "No ./src directory found! Exiting."
```
alexwlchan: nit: let’s use `src` throughout

docker/python3.6_ci/src/run.sh
```diff
@@ -27,6 +35,7 @@ function run_tests {
   echo "Testing ..."
 
   install_dependencies
+  FIND_MATCH_PATHS=${FIND_MATCH_PATHS:=/data}
```
alexwlchan: See here: https://stackoverflow.com/a/27445490/1558022

We want to assign this even if the variable is unset, not just set and null.

miro_preprocessor/README.md
```diff
@@ -13,4 +13,10 @@ An ECS Task with inputs:
 - S3 Source location (a Miro formatted XML dump)
 - S3 Target location
 
-This task downloads and converts a Miro formatted XML dump to JSON and puts it at the S3 target location specified.
\ No newline at end of file
+This task downloads and converts a Miro formatted XML dump to JSON and puts it at the S3 target location specified.
+
+## xml_to_json_run_task
+
+A lambda which is triggered by an S3 upload of a new Miro xml dump.
```
alexwlchan: Detail on where it should be uploaded to?

terraform/ecs_script_task/templates/default.json.template
```diff
@@ -4,7 +4,7 @@
     "essential": true,
     "image": "${app_uri}",
     "memory": ${memory},
-    "name": "app",
+    "name": "${name}",
```
alexwlchan: Why has this become a variable?

miro_preprocessor/main.tf
```diff
@@ -1,56 +1,24 @@
...
-    actions = [
-      "s3:Put*",
-    ]
+  s3_read_miro_data_json = "${data.aws_iam_policy_document.s3_read_miro_data.json}"
 
-    resources = [
-      "${data.terraform_remote_state.platform.bucket_miro_data_arn}/json",
-    ]
-  }
+  task_defintion_arn = "${module.xml_to_json_converter.task_arn}"
```
alexwlchan: `task_defintion_arn`?

`definition`?

miro_preprocessor/xml_to_json_run_task/main.tf
```diff
@@ -0,0 +1,31 @@
...
+  description = "Start the Miro XML to JSON ECS task"
+
+  alarm_topic_arn = "${var.lambda_error_alarm_arn}"
+
+  environment_variables = {
+    TOPIC_ARN           = "${var.topic_arn}"
+    CLUSTER_NAME        = "${var.cluster_name}"
+    CONTAINER_NAME      = "${var.container_name}"
+    TASK_DEFINITION_ARN = "${var.task_defintion_arn}"
+  }
```
alexwlchan: 👍

terraform/ecs_script_task/templates/default.json.template
kenoir: We need to pass this along to the run task lambda and i prefer this approach to hardcoding the value.

miro_preprocessor/main.tf
kenoir: Good point - task_arn actually has a different meaning in AWS as the `arn` of a running instance of a task definition, as distinct from the definition itself.

kenoir: In case it's not clear - this should be `task_defintion_arn`.

Issue #213, opened 2017-05-02
Return JSON error responses


jtweed:
Details TBC, as part of the design guidelines.

alexwlchan:
https://api.wellcomecollection.org/catalogue/v0/works?includes=foo

alexwlchan:
Current model:

```json
{
    "type": "Error",
    "errorType": "http",
    "httpStatus": 404,
    "description": "Work xyzzy was not found"
}
```

We’ll need to do errors for, at least:

- 404 you looked up a bad work in Elasticsearch
- 400 you paged too far off the end, or 400 bad parameters (re-wrapping the Finatra responses)
- 500 Server error

alexwlchan:
We also need to get the new Error type in the Swagger docs, where currently it doesn't appear.

alexwlchan:
New Error type is in the Swagger docs, waiting on @kenoir for a deployment and publish to Stoplight.

alexwlchan:
API docs are now in Stoplight, so closing as done.

PR #916, opened 2017-09-13
Return JSON 400 responses for bad pagination parameters

### LINKED ISSUE #848: Return a 400 Error if the user tries to fetch beyond the end of the ES result size
Trying to fetch beyond the first 10000 results causes Elasticsearch to break. This triggers an HTTP 500 error in the API. e.g. https://api.wellcomecollection.org/catalogue/v0/works?page=10000

We should wrap this to return a nicer message and be a 400 instead.

Required for #490.

alexwlchan:

Our Elasticsearch index can only return the first 10000 results. If you try to page beyond that, you get a 500 Internal Server Error. Make that a 400 Bad Request error instead.

Resolves #848. Related: #213. Requires #915 to be merged first.

### Who is this change for?

Users who want more useful error messages.

Developers who don’t want to be alerted to this sort of error.

- [ ] Reviewed by @jtweed
- [ ] Deployed new versions


api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
```diff
@@ -75,35 +76,45 @@ class WorksController @Inject()(@Flag("api.prefix") apiPrefix: String,
...
-            includes = request.includes.getOrElse(WorksIncludes()),
-            index = request._index
-          )
-        case None =>
-          worksService.listWorks(
-            pageSize = pageSize,
-            pageNumber = request.page,
-            includes = request.includes.getOrElse(WorksIncludes()),
-            index = request._index
+      if (pageSize * request.page > esMaxResultWindow) {
```
kenoir: Not sure whether it's better to just allow the bad query to run and catch that exception in the error mapper.

alexwlchan: I think easier to catch upfront – if run the query and catch the exception, we’re paying the penalty for a query we could skip, and have to work out if the exception is this particular failure mode, or something different (which should still raise as a 500).

kenoir: Yeah, but this does move logic about what an appropriate search size is into this controller when actually ES is controlling that. Also having to work out  if the exception is this particular failure mode - because it means we can be sure this is actually the error that has occurred.

alexwlchan: Mmm. Yeah, okay, convinced – time to bring back the custom exception mapper!

And it occurs to me we might want to surface other stuff as 400 errors in future, e.g. malformed query syntax.

alexwlchan:
I went back and re-implemented this with an exception mapper, so I think it’s ready for another review.

api/src/main/scala/uk/ac/wellcome/platform/api/finatra/exceptions/ElasticsearchResponseExceptionMapper.scala
```diff
@@ -0,0 +1,111 @@
...
+  //
+  //     Result window is too large, from + size must be less than or equal
+  //     to: [10000] but was [100000]. See the scroll api for a more
+  //     efficient way to request large data sets. This limit can be set by
+  //     changing the [index.max_result_window] index level setting.
+  //
+  // In this case, we return a 400 Bad Request error, but we rephrase it
+  // to be shorter and to avoid talking about Elasticsearch concepts.
+  val resultSizePattern =
+    """Result window is too large, from \+ size must be less than or equal to: \[([0-9]+)\]""".r.unanchored
```
alicefuzier: Sorry for being a bit late in reviewing this, but I wonder if it's worth trying doing this without matching on the message. It seems a bit brittle to me: we might upgrade the library and find out that this is broken because the message has changed slightly.

alexwlchan: I’m not opposed to the idea, but I couldn’t see anything more specific to match on in the JSON:

```json
{
  "error": {
    "root_cause": [
      {
        "type": "query_phase_execution_exception",
        "reason": "Result window is too large, from + size must be less than or equal to: [10000] but was [100010]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."
      }
    ],
    "type": "search_phase_execution_exception",
    "reason": "all shards failed",
    "phase": "query",
    "grouped": true,
    "failed_shards": [
      {
        "shard": 0,
        "index": ".kibana",
        "node": "nrhHcUIZQxiEASp_g93UaQ",
        "reason": {
          "type": "query_phase_execution_exception",
          "reason": "Result window is too large, from + size must be less than or equal to: [10000] but was [100010]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."
        }
      }
    ]
  },
  "status": 500
}
```

Also, if the message changes, we have a test for this code path, so it wouldn’t fail silently.

alicefuzier: Yeah, I know, but it still feels wrong to have code that works on what is essentially a logging message. It feels a bit of a hack. Although I don't really have an answer as to how to do it, so fine to leave as is. I'll try poking around a bit

PR #927, opened 2017-09-15
Copy Miro s3 assets lambda

alicefuzier:
Copy assets to another bucket, so that they are available for the Catalogue API (for now - eventually this could be used to move them in separate buckets for TV or Glacier)
### Who is this change for?
Images

- [ ] Run `terraform apply`.


miro_preprocessor/miro_copy_s3_asset/src/miro_copy_s3_asset.py
```diff
@@ -0,0 +1,66 @@
...
+            print(f"No image found for MiroId {miro_id}: skipping")
+            pass
+        else:
+            raise
+    else:
+        destination_key = f"{shard}/{miro_id}.jpg"
+        try:
+            destination_head_response = s3_client.head_object(Bucket=destination_bucket_name, Key=destination_key)
+        except ClientError as client_error:
+            if client_error.response['Error']['Code'] == '404':
```
kenoir: giphy 3

Don't go too deep ...

PR #937, opened 2017-09-18
Whitelist the User-Agent header to be passed to origin

kenoir:

Pass the `User-Agent` header to origin (Loris) so that logs contain this info.

**Note:** The headers will have `Cloudfront-` prefixed to them when they are passed from CloudFront as per https://aws.amazon.com/blogs/aws/enhanced-cloudfront-customization/

### Who is this change for?

💻 

- [ ] Run `terraform apply`.


alexwlchan:
Merging by fiat, as Travis isn't going to tell us anything useful about this.

PR #946, opened 2017-09-18
Remove Terraform for the Imgix read-only user

alexwlchan:

Since the Experience team have removed their Imgix stack, we don’t need this user for giving read-only access to our bucket. We can delete the Terraform and the corresponding users.

### Who is this change for?

Devs who don’t want unnecessary users hanging around, and lean and mean Terraform stacks!

- [ ] Run `terraform apply`.

### DIFF
--- a/shared_infra/iam_policy_document.tf
+++ b/shared_infra/iam_policy_document.tf
@@ -198,19 +198,6 @@ data "aws_iam_policy_document" "s3_put_gatling_reports" {
   }
 }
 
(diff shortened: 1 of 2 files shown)


kenoir:
👍  but I reckon this should be approved by someone on the experience team.

jamesgorrie:
@gestchild do we ever read from the origin?

kenoir:
@jamesgorrie I think you do - but that is also public.

PR #948, opened 2017-09-19
Sorter rules

kenoir:

Adding the rules to the sorter for the miro preprocessor pipeline

### Who is this change for?

💻  

- [ ] Deployed new versions


miro_preprocessor/image_sorter/src/sorter_logic.py
```diff
@@ -10,6 +10,45 @@
 import re
 
 
+class Rules:
```
kenoir: Looks good 👍

miro_preprocessor/image_sorter/src/test_sorter_logic.py
```diff
@@ -5,6 +5,29 @@
...
+            "image_library_dept": "Paperclips and hairnets",
+            "image_tech_captured_mode": "Frog retina",
+            "image_copyright_cleared": "N",
+            "image_access_restrictions": "CC-BY",
+            "image_general_use": "N",
+            "image_innopac_id": "12345678"
+        }}
+    collection = image['collection']
+    if 'collection' in kwargs.keys():
+        collection = kwargs.pop('collection')
```
alexwlchan: One slightly neater way of doing this might be to use the `.get()` method on dictionaries, which takes a key and a default:

```python
collection = kwargs.get('collection', 'source/images-A')
```

Then you could declare `image_data` directly, rather than through a level of dict structure.

miro_preprocessor/image_sorter/src/sorter_logic.py
```diff
@@ -15,14 +77,55 @@ class Decision(enum.Enum):
...
+    r = Rules(collection, image_data)
+    print(r)
 
-    raise Undecidable(
-        f'No decision for collection={collection!r}, image_data={image_data!r}'
-    )
+    if r.is_f_collection or \
+            (r.is_l_or_m_or_v_collection and r.image_library_dept_is_Archives_and_Manuscripts) or \
+            (r.is_l_or_m_or_v_collection and r.image_tech_captured_mode_is_videodisc) or \
+            (r.is_l_or_m_or_v_collection and not r.is_innopac_id_8_digits and r.is_title_blank and r.is_image_pub_title_blank and r.is_image_pub_periodical_blank):
```
kenoir: Consider reversing this logic so that we don't need the not here so we can be clearer about what this is doing without worrying about operator precedence, i.e, make this `is_innopac_id__not_8_digits

miro_preprocessor/image_sorter/src/test_sorter_logic.py
```diff
@@ -5,16 +5,52 @@
...
+    image_data = image['image_data']
+    image_data.update(kwargs)
+
+    return collection, image_data
 
 
 @pytest.mark.parametrize('collection, image_data', [
-    # TODO: Write some examples...
+    collection_image_data(collection='source/images-F'),
+    collection_image_data(collection='source/images-L', image_library_dept="Archives and Manuscripts"),
```
kenoir: 👍  on only filling out the necessary test data

alicefuzier:
So, _I think_ rules are implemented. What remains to be done is changing `miro_image_sorter` to expect a list of `Decision` from `sort_image` and to make it deal with 5 possible types of `Decision` instead of 3. Unfortunately, as it stands the lambda is broken so we cannot merge yet

miro_preprocessor/image_sorter/src/sorter_logic.py
```diff
@@ -6,23 +6,206 @@
...
+    def image_library_dept_is_Public_programmes(self):
+        return self._compare("image_library_dept", "Public programmes")
+
+    @property
+    def image_tech_captured_mode_is_videodisc(self):
+        return self._compare("image_tech_captured_mode", "videodisc")
+
+    @property
+    def is_innopac_id_8_digits(self):
+        return self._search(r"[0-9]{7}[0-9|x]{1}", "image_innopac_id") is not None
```
alexwlchan: This regex is a bit too permissive: in particular, it matches `1234567|`, which surely isn't the intention? You don’t need the `|` in the second group: the following is sufficient:

```python
return self._search(r"[0-9]{7}[0-9x]{1}", "image_innopac_id")
```

That everything is in square brackets is enough to implicitly add the OR for you, so the `|` is treated as a literal.

Also, do we care about case?

miro_preprocessor/image_sorter/src/sorter_logic.py
```diff
@@ -6,23 +6,206 @@
...
+        return self._get_normalised(key) is None
+
+    def _key_matches(self, key, string_list):
+        return any([self._compare(key, string) for string in string_list])
+
+    def _parse_date(self, date_string):
+        return dateutil.parser.parse(date_string, dayfirst=True)
+
+    def is_collection(self, *args):
+        return any([self._is_collection(collection) for collection in args])
```
alexwlchan: If you drop the square brackets, `any()` will lazily generate the conditions, and stop as soon as it finds a truthy condition. This is slightly more efficient.

(Why: Python has _generator comprehensions_, which are like list comprehensions, but lazily evaluated. e.g.

```python
my_generator = (self._is_collection(collection) for collection in args)
```

There's then a bit of syntactic sugar that says if the only argument to a function is a generator, you can drop one pair of parentheses.)

miro_preprocessor/image_sorter/src/sorter_logic.py
```diff
@@ -6,23 +6,206 @@
...
+
+    @property
+    def is_image_pub_periodical_blank(self):
+        return self._is_blank("image_pub_periodical")
+
+    @property
+    def is_after_first_march_2016(self):
+        scanned_date = self._get("image_tech_scanned_date")
+        if scanned_date is not None:
+            return self._parse_date(scanned_date) > self._parse_date("01/03/2016")
```
alexwlchan: nit: I can see above the parser treats it as day first, but might be nice to write this as `self._parse_date("1 March 2016")` so the human reader can see it will always be parsed correctly.

miro_preprocessor/image_sorter/src/sorter_logic.py
alexwlchan: Also, nice use of `*args`! :)

miro_preprocessor/image_sorter/src/sorter_logic.py
```diff
@@ -6,23 +6,206 @@
...
+    def is_for_public_access(self):
+        return not self.is_not_for_public_access
+
+    @property
+    def satisfies_api_filters(self):
+        return self.is_copyright_cleared and self.is_cleared and self.has_not_use_restrictions
+
+    @property
+    def is_cold_store(self):
+        return self.is_collection("F") or \
```
alexwlchan: nit: Python treats everything inside a set of parentheses as one line, so you don't need line continuations. This means you can write:

```python
return (
    self.is_collection("F") or
    (self.is_collection("L", "M", "V") and self.image_library_dept_is_Archives_and_Manuscripts) or
    ...
)
```

which I find slightly neater.

Up to you though, it makes no functional difference.

miro_preprocessor/image_sorter/src/sorter_logic.py
```diff
@@ -6,23 +6,206 @@
...
-    # For now reproduce the behaviour of the miro_adapter and send all images to the catalogue API.
-    return Decision.catalogue_api
+    print(collection)
+    print(image_data)
+    r = Rules(collection, image_data)
+
+    decisions = []
+
+    if not r.is_collection("F", "L", "V", "M"):
+        raise InvalidCollectionException
```
alexwlchan: Maybe include the metadata in the exception? Then you can search logs for the exception and it's all on one line, rather than searching and then having to work back up.

(And if we have Lambdas running concurrently, that may get harder)

miro_preprocessor/image_sorter/src/sorter_logic.py
alicefuzier: I added the day first thing because it was parsing the string "29/02/2016" incorrectly otherwise. No idea why as it doesn't look ambiguous to me, but it was thinking that it was after the first of March 2016, so, for now, I think it's ok to be restrictive in the formatting of the date

PR #952, opened 2017-09-19
The Miro adapter is pining for the fjords

alexwlchan:

Now we have the Miro preprocessor pipeline, this adapter unnecessary. It is no more. It has ceased to be. It is an ex-adapter. The only reason it’s still in the repo is because it’s been nailed there.

So, let’s delete it instead.

### Who is this change for?

A platform team who want a neat and tidy repo.

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


kenoir:
👍 Will wait for proof we can reproduce this behaviour with the new pipeline before merging.

alexwlchan:
I reckon we’re close enough and I want to get this in before the other build changes, merging.

PR #959, opened 2017-09-19
Add logging to XML conversion task

kenoir:

More visibility of what's going in our xml to json conversion task.

### Who is this change for?

💻 

- [ ] Run `terraform apply`.

### DIFF
--- a/miro_preprocessor/xml_to_json_converter/src/run.py
+++ b/miro_preprocessor/xml_to_json_converter/src/run.py
@@ -38,6 +38,7 @@ def _wrap_image_data(collection, image_data):
 
 
 def main(bucket, src_key, dst_key, js_path="json"):
+    print(f"Starting to process s3://{bucket}/{src_key}.")
     image_data = generate_images(bucket=bucket, key=src_key)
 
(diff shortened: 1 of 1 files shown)


miro_preprocessor/xml_to_json_converter/src/run.py
```diff
@@ -47,6 +48,8 @@ def main(bucket, src_key, dst_key, js_path="json"):
 
     s3 = boto3.client('s3')
 
+    image_data_count = 1
```
alexwlchan: You don’t need to manage the loop yourself.

Replace L54 with:

```python
for count, img in enumerate(image_data, start=1):
```

See https://docs.python.org/3/library/functions.html#enumerate

kenoir: Will that work on a generator?

alexwlchan: Yes!

```python
def infinite():
    x = 1
    while True:
        x *= 2
        yield x

for i, x in enumerate(infinite()):
    print(i, x)
    if i > 10:
        break
```

PR #968, opened 2017-09-20
Return a 404 error if you look up a non-existent index

### LINKED ISSUE #965: Selecting the wrong index should be an HTTP 404 error
Right now, if you hit https://api.wellcomecollection.org/catalogue/v0/works?_index=foo you can search an alternative index.

This is useful when we’re doing reindexing, and we’ll probably use it more as we start tweaking the search weightings.

But if you try it and the index doesn’t exist, you get a 500 error, and an alarm pops up in Slack. Sadness! We should make this a 404 error instead.

One note, however: we’re using the non-existence of an index to test that 500s return correctly, because it’s a reliable source of errors. What can we use instead for those tests?

alexwlchan:

In the title. Which is really a way to stop alarms in our Slack channel for what’s a pretty benign error. Plus some tidying the internal logic of the ExceptionMapper, which is still a bit crappy but now in a different way. (A pox on the Elasticsearch SDK for not returning proper error types!)

Resolves #965.

### Who is this change for?

Me, because I’m bored of investigating those alarms.

oh also i guess users who want helpful error messages \*sigh*

- [ ] Deployed new versions

### DIFF
--- a/api/src/main/scala/uk/ac/wellcome/platform/api/finatra/exceptions/ElasticsearchResponseExceptionMapper.scala
+++ b/api/src/main/scala/uk/ac/wellcome/platform/api/finatra/exceptions/ElasticsearchResponseExceptionMapper.scala
@@ -58,6 +58,7 @@ class ElasticsearchResponseExceptionMapper @Inject()(
 
   private def jsonToError(jsonDocument: String,
                           exception: Exception): DisplayError = {
+    val mapper = new ObjectMapper()
     val exceptionData = mapper.readTree(jsonDocument)
     val reason = exceptionData
       .get("error")
@@ -101,12 +102,11 @@ class ElasticsearchResponseExceptionMapper @Inject()(
       // a JSON response, so there's only a single-line message -- then `.head`
       // fails: "NoSuchElementException: next on empty iterator".  I think
       // this happens when the cluster doesn't reply, but I'm not sure.
-      val mapper = new ObjectMapper()
       val jsonDocument = exception.getMessage
         .split("\n")
         .tail
         .head
-      jsonToError(jsonDocument)
+      jsonToError(jsonDocument = jsonDocument, exception = exception)
     } catch {
       case e: Exception =>
         serverError(s"Error ($e) parsing Elasticsearch response", exception)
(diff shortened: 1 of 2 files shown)


api/src/main/scala/uk/ac/wellcome/platform/api/finatra/exceptions/ElasticsearchResponseExceptionMapper.scala
```diff
@@ -29,52 +29,91 @@ class ElasticsearchResponseExceptionMapper @Inject()(
     error(
       s"Sending HTTP 400 from ElasticsearchResponseExceptionMapper ($message)",
       exception)
-    DisplayError(Error(variant = "http-400", description = Some(message)))
+    DisplayError(Error(variant = s"http-400", description = Some(message)))
```
alexwlchan: s-strings here are unnecessary, and an artefact of an earlier version of this code

api/src/main/scala/uk/ac/wellcome/platform/api/finatra/exceptions/ElasticsearchResponseExceptionMapper.scala
```diff
@@ -29,52 +29,91 @@ class ElasticsearchResponseExceptionMapper @Inject()(
...
+    val exceptionData = mapper.readTree(jsonDocument)
+
+    val esError = exceptionData.get("error")
+    val esErrorType = esError
+      .get("type")
+      .asText
+
+    esErrorType match {
+      // This occurs if the user requests a non-existent index as ?_index=foo.
+      // We return this as a 404 error to the user.
```
alexwlchan: this line is redundant

api/src/main/scala/uk/ac/wellcome/platform/api/finatra/exceptions/ElasticsearchResponseExceptionMapper.scala
```diff
@@ -29,52 +29,91 @@ class ElasticsearchResponseExceptionMapper @Inject()(
...
+      case "index_not_found_exception" => {
+        val index = esError
+          .get("index")
+          .asText
+        notFound(s"There is no index $index", exception)
+      }
+
+      // This may occur if the user requests an overly large page of results.
+      // We return this as a 400 error to the user.
```
alexwlchan: and this, just needs a comment about the error type

PR #969, opened 2017-09-21
Bump the API to v1

alexwlchan:

Bump the API version from v0 to v1. We’re production-ready! Or at least, we have a shiny new version number.

The aim is that new deployments of the API should response to catalogue/v1, but we’ll continue to serve the current version on catalogue/v0, at least until we can get the Experience site migrated over.

Related: #582.

### Who is this change for?

Our quarterly goals! Also, folks who want a guarantee of stability from the API – we’re not going to make backwards-incompatible changes with v1.

- [ ] Reviewed by @jtweed
- [ ] Deployed new versions
- [ ] Run `terraform apply`.

### DIFF
--- a/api/src/main/scala/uk/ac/wellcome/platform/api/Server.scala
+++ b/api/src/main/scala/uk/ac/wellcome/platform/api/Server.scala
@@ -41,7 +41,7 @@ class Server extends HttpServer {
   private final val apiName =
     flag(name = "api.name", default = "catalogue", help = "API name path part")
   private final val apiVersion =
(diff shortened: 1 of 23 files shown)


jamesgorrie:
@alexwlchan @kenoir do we loose V1?

alexwlchan:
> do we lose v0?

Not immediately. At least initially, we’ll be running both v0 and v1, so we can flip the switch without the Experience site breaking.

alexwlchan:
I’m going to wait until I’ve had lunch to merge this.

PR #973, opened 2017-09-22
Build correct collection from object src key

kenoir:

The `xml_to_json_convertor` ECS task wraps the image data from the Miro XML and adds a collection identifier.

It should pass along an identifier understood by later steps in the pipeline.

### Who is this change for?

Folks who want new images in the API

- [ ] Deployed new versions

### DIFF
--- a/miro_preprocessor/xml_to_json_converter/src/test_run.py
+++ b/miro_preprocessor/xml_to_json_converter/src/test_run.py
@@ -128,3 +128,12 @@ def test_creates_json_file_for_each_image(s3_fixture, xml_file_contents):
     actual_json_objects = {key: _get_body(key) for key in keys}
 
     assert expected_json_objects == actual_json_objects
+
+
+def test_build_collection_id():
+    src_key = "source/Images-V.xml"
+    expected_collection_id = "Images-V"
+
+    actual_collection_id = run._build_collection_id(src_key)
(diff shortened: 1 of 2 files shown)


miro_preprocessor/xml_to_json_converter/src/test_run.py
```diff
@@ -128,3 +128,12 @@ def _get_body(key):
     actual_json_objects = {key: _get_body(key) for key in keys}
 
     assert expected_json_objects == actual_json_objects
+
+
+def test_build_collection_id():
+    src_key = "source/Images-V.xml"
```
alicefuzier: I'm not sure whether that will cause issues in the transformer or not, but the files we've put in S3 so far have key `sources/images-V.xml`, without the capital `I`. So the expected collection, in that case, would be `images-V` instead of `Images-V`. The logic in the transformer checks for `Images-V` with capital `I` if I remember correctly, so _I think_ we would probably still have issues with this logic. Maybe the transformer should be a bit more robust in wjat it looks for

kenoir: 👍  https://github.com/wellcometrust/platform/pull/974/files#diff-6e1939a684d42b780d5ea33c26b2aab8R22

PR #974, opened 2017-09-22
Don't expect collection to match, rather contain

kenoir:

In the Transformer when trying to extract title information, use `contains` rather than equality to be more pessimistic about what we are sent.

### Who is this change for?

💻 

- [ ] Deployed new versions.

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/models/transformable/miro/MiroTransformable.scala
+++ b/common/src/main/scala/uk/ac/wellcome/models/transformable/miro/MiroTransformable.scala
@@ -149,7 +149,7 @@ case class MiroTransformable(MiroID: String,
   }
 
   def getCreatedDate(miroData: MiroTransformableData): Option[Period] =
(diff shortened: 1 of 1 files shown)


common/src/main/scala/uk/ac/wellcome/models/transformable/miro/MiroTransformable.scala
```diff
@@ -61,8 +61,9 @@ case class MiroTransformable(MiroID: String,
     val titleIsTruncatedDescription = candidateTitle
       .startsWith(miroData.title.get)
 
-    val useDescriptionAsTitle = (titleIsTruncatedDescription &&
-      MiroCollection == "Images-V") || (miroData.title.get == "-" || miroData.title.get == "--")
+    val useDescriptionAsTitle =
+      (titleIsTruncatedDescription && MiroCollection.contains("Images-V")) ||
```
alexwlchan: lowercase?

PR #976, opened 2017-09-22
Remove Transformer rules for rejecting images

kenoir:

This PR removes the rules in the miro-transformer for rejecting images. This _should_ now have passed to the miro-preprocessor pipeline, we'll need to review these changes in conjunction with those in order to ensure the changes don't clash.

### Who is this change for?

Christy and us and anyone who wants a complete collection.

- [ ] Deployed new versions


common/src/main/scala/uk/ac/wellcome/models/transformable/miro/MiroTransformChecks.scala
```diff
@@ -1,98 +0,0 @@
...
-   * in the same API result, but for now, we just exclude them from
-   * the API.  They aren't useful for testing image search.
-   */
-  private def checkImageExists(
-    data: MiroTransformableData): List[FieldIssues] =
-    if (data.techFileSize.getOrElse(List[String]()).isEmpty) {
-      List(
-        FieldIssues(
-          "image_tech_file_size",
-          data.techFileSize,
```
alicefuzier: I think we might be able to remove this file entirely from `MiroTransformableData`. I don't think anything else is using it

Issue #980, opened 2017-09-22
Put our monitoring services in a separate stack


alexwlchan:
Move monitoring out of the two monolith stacks (shared_infra and lambdas), and pull them into a dedicated stack.

### Who is this change for?

Platform devs who want a manageable repo/stacks.

Looking more widely, I would guess our monitoring is the most easily-reusable part of the platform, so having it all in one place will probably be helpful for third-party devs who want to use our stuff. 🎉

PR #981, opened 2017-09-22
Add a script for building Lambda ZIPs

alexwlchan:

This is a new script, deliberately not called from anywhere, which uploads prepared Lambdas as ZIP files to Amazon S3. You can see the output here:

https://console.aws.amazon.com/s3/buckets/platform-infra/lambdas/?region=eu-west-1&tab=overview

Related: #971, #980.

### Who is this change for?

We need this if we want Lambdas elsewhere in the repo.

### DIFF
--- a/lambdas/build_lambda_zips.py
+++ b/lambdas/build_lambda_zips.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python
+# -*- encoding: utf-8 -*-
+"""
+This script builds a deployment ZIP for every Lambda in the repository,
+and uploads it to Amazon S3.  It handles copying in the common lib, and
+installing any necessary dependencies.
+"""
+
+import os
+import shutil
+import subprocess
+import tempfile
+import zipfile
+
+import boto3
+
+
+ROOT = subprocess.check_output([
+    'git', 'rev-parse', '--show-toplevel']).decode('ascii').strip()
+
+UTILS_LIB = os.path.join(ROOT, 'lambdas', 'common', 'utils')
+
+ZIP_DIR = os.path.join(ROOT, '.lambda_zips')
+
+
+def create_zip(src, dst):
+    """
+    Zip a source directory into a target directory.
+
+    Based on https://stackoverflow.com/a/14569017/1558022
+    """
+    name = f'{dst}.zip'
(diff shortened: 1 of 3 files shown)


lambdas/build_lambda_zips.py
```diff
@@ -0,0 +1,117 @@
...
+                dst=os.path.join(target, os.path.basename(f))
+            )
+
+    # Copy the contents of the Lambda common lib
+    shutil.copytree(UTILS_LIB, os.path.join(target, 'utils'))
+
+    # Now install any additional pip dependencies.
+    reqs_file = os.path.join(path, 'requirements.txt')
+    if os.path.exists(reqs_file):
+        print(f'*** Installing dependencies from requirements.txt')
```
kenoir: 👍

lambdas/build_lambda_zips.py
```diff
@@ -0,0 +1,117 @@
...
+def upload_zip_to_s3(client, filename):
+    print(f'*** Uploading {filename} to S3')
+    client.upload_file(
+        Bucket='platform-infra',
+        Filename=filename,
+        Key=f'lambdas/{os.path.basename(filename)}'
+    )
+
+
+def find_lambdas():
```
kenoir: I'd much rather this were handled centrally - more like there is a config file in any directory that contains something we wish to build stating the _kind_ or project it is, we can then hand the responsibility for building a project of type _x_ to the correct builder.

alexwlchan: Okay, so how about a `.project` file that contains one of: `lambda`, `ecs_task`, `ci_docker_container`, and so on?

alexwlchan: Actually, `.project` should probably be a more sophisticated file, say JSON:

```json
{
    "type": "lambda"
}
```

so we can add more metadata later.

kenoir: 👍

alexwlchan:
@kenoir I’ve reduced the scope of the script: now it just gets handed a path, a key and a bucket, and produces the ZIP given those arguments. How it gets called can be handled later – but this reduces this to a mergeable patch.

Re-review please? :-)

PR #984, opened 2017-09-25
Allow the ASG to dictate whether it's spot powered

kenoir:

A stable uniform, spot powered services cluster, described by a single ASG module. 

### Who is this change for?

💻 

- [ ] Run `terraform apply`.


kenoir:
**Note:** it's likely that running this against the API would cause an outage if not managed properly. Don't just merge and deploy this change!

terraform/ecs_asg/ecs_asg_launch_config/outputs.tf
```diff
@@ -0,0 +1,15 @@
+output "name" {
+  value = "${element(concat(aws_launch_configuration.spot_launch_config.*.name, aws_launch_configuration.ondemand_launch_config.*.name), 0)}"
```
kenoir: IMHO this is batshit but: https://github.com/hashicorp/terraform/issues/12520

terraform/ecs_asg/ecs_asg_launch_config/security_groups.tf
```diff
@@ -1,7 +1,7 @@
 resource "aws_security_group" "instance_sg" {
   description = "controls direct access to application instances"
   vpc_id      = "${var.vpc_id}"
-  name        = "${var.asg_name}_instance_sg"
+  name        = "${var.asg_name}_instance_sg_${random_id.sg_append.hex}"
```
kenoir: The random ID along with `create_before_destroy` avoids a dependency issue where the old SGs cannot be deleted as they are in use.

terraform/spot_fleet_request/iam.tf
```diff
@@ -1,36 +0,0 @@
-data "aws_iam_policy_document" "assume_spot_fleet_role" {
```
kenoir: Removing these for now, but they are still valuable if we wish to use _only_ spot fleet requests in the future.

PR #985, opened 2017-09-26
Do not crash if the innopac_id is empty

alicefuzier:
Make image_sorter lambda not crash if innopac_id is empty
### Who is this change for?
Devs

### DIFF
--- a/miro_preprocessor/image_sorter/src/sorter_logic.py
+++ b/miro_preprocessor/image_sorter/src/sorter_logic.py
@@ -145,7 +145,8 @@ class Rules:
             self.is_copyright_cleared and \
             self.is_cleared and \
             (not self.has_use_restrictions) and \
-            (not self.use_restrictions_are_none)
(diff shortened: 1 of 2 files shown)


miro_preprocessor/image_sorter/src/sorter_logic.py
```diff
@@ -85,7 +85,7 @@ def image_tech_captured_mode_is_videodisc(self):
 
     @property
     def is_innopac_id_8_digits(self):
-        return self._search(r"[0-9]{7}[0-9xX]{1}", "image_innopac_id") is not None
+        return self._get("image_innopac_id") is not None and self._search(r"[0-9]{7}[0-9xX]{1}", "image_innopac_id") is not None
```
alexwlchan: Would this logic be better folded inside `_search`? If you search on an empty key, you automatically get a None result?

alicefuzier: 👍

Issue #955, opened 2017-09-19
Run Loris in its own ECS cluster


alexwlchan:
This would be nice, but I ran into ELB health check problems the first time I tried to do this. 😕

alexwlchan:
Closed by #986.

Issue #782, opened 2017-08-11
Investigate moving Loris caches to EBS


alexwlchan:

alicefuzier:
More experiments on Loris performance are in progress at #1039

PR #986, opened 2017-09-26
Move Loris to its own ECS cluster

alicefuzier:
What the title says so that we can provision the ASG in different ways for Loris and for the API. This will allow us to add EBS volumes for #782 without having to add EBS volumes for API instances as well
### Who is this change for?
Devs

- [x] Run `terraform apply`.


alicefuzier:
I think Loris can stay in the same VPC as the API and can continue to share the same ALB, but I'm not too sure about that, so let me know what you guys think

loris/terraform/asg.tf
```diff
@@ -0,0 +1,20 @@
+
+module "loris_cluster_asg" {
+  source                = "../../terraform/ecs_asg"
+  asg_name              = "loris-cluster"
+  subnet_list           = ["${module.vpc_api.subnets}"]
```
alexwlchan: You’ll need to pull this in from the shared stack.

loris/terraform/asg.tf
```diff
@@ -0,0 +1,20 @@
...
+  key_name              = "${var.key_name}"
+  instance_profile_name = "${module.ecs_loris_iam.instance_profile_name}"
```
alexwlchan: This needs to be copied across.

alexwlchan:
Resolves #955.

loris/terraform/ecr.tf
```diff
@@ -1,9 +1,9 @@
 module "ecr_nginx_loris" {
-  source = "../terraform/ecr"
+  source = "../../terraform/ecr"
```
alexwlchan: Mildly surprised that this has changed, has something in the Makefile changed?

PR #988, opened 2017-09-26
Changes to allow Loris to use the new filtered MIRO images bucket

kenoir:

Allow Loris to serve more collections in order to increase the size of the collection which is visible to the public!

### Who is this change for?

💻  & 👩 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


docker/nginx/loris.nginx.conf
```diff
@@ -42,8 +42,8 @@ http {
     # For example V0002007.jpg is actually stored in V0002000/V0002007.jpg in the s3 bucket.
     # Therefore every request for V0002007.jpg needs to be routed to s3:V0002000/V0002007.jpg
     #
-    location ~ "^/image/V[0-9]{4}.*" {
-      rewrite "^/image/(?<filename>(?<shard>V[0-9]{4}).*)" "/s3:${shard}000/${filename}" break;
+    location ~ "^/image/[A-Z][0-9]{4}.*" {
```
alicefuzier: I think there can be more than one letter at the beginning of a miro id. There is a similar regex [here](https://github.com/wellcometrust/platform/blob/42700ed7a65d1eade1bb95d9d92371b6ec615076/miro_preprocessor/miro_copy_s3_asset/src/miro_copy_s3_asset.py#L70). You might want to nick that

docker/nginx/loris.nginx.conf
```diff
@@ -42,8 +42,8 @@ http {
...
+      rewrite "^/image/(?<filename>(?<shard>{[A-Z][0-9]{4}).*)" "/s3:${shard}000/${filename}" break;
```
alicefuzier: ditto

docker/nginx/loris.nginx.conf
kenoir: 👍

loris/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM ubuntu:14.04
 
-RUN apt-get update
+RUN apt-get update --fix-missing
```
alexwlchan: Mmm. Why do we need this? It should run afresh every time.

Is the issue that this layer is already cached locally, I wonder?

kenoir: It rebuilt for me and didn't work without this :(

alexwlchan: I think the correct fix is to tie this to the second line:

```
RUN apt-get update &&
    apt-get install --yes ...
```

so we get an update *and* an install in the same, cached step.

(Let’s talk about exactly why this works when you’re free.)

shared_infra/iam_policy_document.tf
```diff
@@ -1,3 +1,20 @@
...
+  statement {
+    actions = [
+      "s3:GetObject",
+    ]
+
+    resources = [
+      "arn:aws:s3:::${aws_s3_bucket.wellcomecollection-miro-images-public.id}/*",
+    ]
+
+    principals {
```
alexwlchan: Do we need this `principals` block? Is it used anywhere else?

terraform/services/config/templates/loris.ini.template
```diff
@@ -51,7 +51,7 @@ templates = 'wordpress,s3,prismic'
     url = 'https://wellcomecollection.files.wordpress.com/%s'
 
     [[s3]]
-    url = 'http://s3-eu-west-1.amazonaws.com/miro-images-public/%s'
+    url = 'http://s3-eu-west-1.amazonaws.com/wellcomecollection-miro-images-public/%s'
```
alexwlchan: Since we’ve never really exposed the `s3://` prefix publicly (it’s in our info.json's, but nobody looks at those), is it worth rethinking the name of this template?

Maybe call it `miro`? To distinguish from other collections we may store in S3 later?

shared_infra/s3.tf
```diff
@@ -70,6 +70,15 @@ resource "aws_s3_bucket" "mets-ingest" {
...
   acl    = "public-read"
+
+  lifecycle {
+    prevent_destroy = true
+  }
+}
+
+resource "aws_s3_bucket_policy" "wellcomecollection-miro-images-public" {
+  bucket = "${aws_s3_bucket.wellcomecollection-miro-images-public.id}"
+  policy = "${data.aws_iam_policy_document.wellcomecollection-miro-images-public.json}"
```
alexwlchan: Mmm. I wonder if this bucket should live in the Loris stack? Because it’s primarily used by Loris?

Not saying it definitely should, just an idea.

loris/README.md
```diff
@@ -32,3 +32,34 @@ You can run Loris locally with the following two commands:
...
+### curl localhost:9000
+
+You can then query the local version of loris (in order to avoid redirects it's necessary to use the `ELB-HealthChecker/2.0` User-Agent).
+
+```sh
+curl \
+    -A ELB-HealthChecker/2.0 \
+    -v \
+    -L \
+    http://localhost:9000/image/V0017087.jpg/full/300,/0/default.jpg > default.jpg
```
alexwlchan: minor: could we get the full names of these options spelt out? Most people will copy-paste this, so verbosity beats terseness.

e.g. `--user-agent "ELB-HealthChecker/2.0"`

Also, you might want to use `--output default.jpg` rather than redirecting stdout.

loris/docker-compose.yml
```diff
@@ -0,0 +1,16 @@
...
+    ports:
+     - "8888:8888"
+    environment:
+      - INFRA_BUCKET=platform-infra
+      - CONFIG_KEY=${CONFIG_KEY}
+      - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
+      - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
+
+  nginx:
+    image: "nginx_loris:${NGINX_TAG}"
```
alexwlchan: Can you do default values in docker-compose env variables? So this defaults to "latest"?

loris/README.md
```diff
@@ -32,3 +32,34 @@ You can run Loris locally with the following two commands:
 where `$PLATFORM` is the root of the platform repo, and `$LORIS` is the root of a local checkout of the Loris repo.
 
 You'll then have a development version of Loris running at <http://localhost:8888>.
+
+## Testing Loris with nginx
+
+In order to test Loris with it's nginx config this project provides a `docker-compose.yml`
```
alexwlchan: nit: /s/it's/its

loris/Dockerfile
kenoir: I think I can guess, just being lazy!

alexwlchan: :D

shared_infra/iam_policy_document.tf
kenoir: This is a bucket policy, so it's a bit different from the others policy documents. The principals block here is granting everyone permission to read any object.

If you look at the `alb_logs` policy it also has a principals block which grants permission to our account to put logs in an S3 bucket (basically giving our account permission to record ALB logs).

alexwlchan: 👍

terraform/services/config/templates/loris.ini.template
kenoir: I think you're correct but as this change is already quite significant (and what you suggest is unnecessary to meet the goal of this PR) i'd prefer to do that in a later patch.

shared_infra/s3.tf
kenoir: I'm not sure TBH, I did think about that - my feeling is it probably could be in the Loris stack as we only use it as a container for images that Loris serves (should be another PR though).

kenoir: Actually changed my mind :) as this PR introduces the change it makes sense to address this here.

loris/README.md
kenoir: 👍 good point

shared_infra/s3.tf
alexwlchan: 👍

loris/docker-compose.yml
kenoir: https://docs.docker.com/compose/env-file/ seems to be the way to do it, which is less than ideal.

alexwlchan:
I’ll merge this once tests pass.

alexwlchan:
Actually, I’ll hold it until we have images in the new index, then merge it.

PR #989, opened 2017-09-26
Revamp the way we publish and deploy Lambdas

### LINKED ISSUE #971: Build Lambda ZIPs in Travis CI, not locally

alexwlchan:

This is a complete rethink of the way we deploy Lambdas. Rather than installing files locally, we:

* Have Travis build a ZIP “deployment package” inside a Docker container
* Upload the ZIP to S3 (but only if there are changes against the current version, and in *non-test code*)
* Use that S3 object when deploying the Lambda

Resolves #971.

Still to do is adding the Travis config, but that’s blocked on a merge of #987.

### Who is this change for?

Devs who want neater deployment procedures.

Also, me when I come to pull apart the Lambdas stack.

- [ ] Deployed new versions
- [x] Run `terraform apply`.

### DIFF
--- a/.travis/run_task.py
+++ b/.travis/run_task.py
@@ -52,6 +52,8 @@ def main():
         make('loris-publish')
     elif task == 'miro_preprocessor-test':
         make('miro_preprocessor-publish')
+    elif task == 'lambdas-test':
+        make('lambdas-publish')
     else:
         task = task.replace('build', 'deploy').replace('test', 'deploy')
         make(task)
--- a/.travis/run_task.py
+++ b/.travis/run_task.py
@@ -49,14 +49,18 @@ def main():
         return 0
 
     if task == 'loris-build':
-        make('loris-publish')
+        publish_task = 'loris-publish'
     elif task == 'miro_preprocessor-test':
-        make('miro_preprocessor-publish')
+        publish_task = 'miro_preprocessor-publish'
     elif task == 'lambdas-test':
-        make('lambdas-publish')
+        publish_task = 'lambdas-publish'
     else:
-        task = task.replace('build', 'deploy').replace('test', 'deploy')
-        make(task)
+        publish_task = (task
+            .replace('build', 'deploy')
+            .replace('test', 'deploy')
+        )
+
+    make(publish_task)
 
     return 0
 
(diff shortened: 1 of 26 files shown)


builds/publish_lambda_zip.py
```diff
@@ -38,15 +38,14 @@ def create_zip(src, dst):
...
-    with zipfile.ZipFile(name, 'w', zipfile.ZIP_DEFLATED) as zf:
+    with zipfile.ZipFile(dst, 'w', zipfile.ZIP_DEFLATED) as zf:
         abs_src = os.path.abspath(src)
         for dirname, subdirs, files in os.walk(src):
             for filename in files:
                 absname = os.path.abspath(os.path.join(dirname, filename))
                 arcname = absname[len(abs_src) + 1:]
                 zf.write(absname, arcname)
-    return name
+    return dst
```
alexwlchan: We shouldn’t return here if `dst` is unmodified.

alexwlchan:
Mmm. This appears to be missing some code – in particular, my hand-rolled ZIP comparison code, which should be replaced with a call to `zipcmp`.

.travis/run_task.py
```diff
@@ -52,6 +52,8 @@ def main():
         make('loris-publish')
     elif task == 'miro_preprocessor-test':
         make('miro_preprocessor-publish')
+    elif task == 'lambdas-test':
+        make('lambdas-publish')
```
kenoir: Do we want to publish on test or is the task misnamed?

alexwlchan: This only runs if we've pushed to master – let me see if I can make the code a little clearer.

builds/publish_lambda_zip.py
```diff
@@ -38,15 +41,15 @@ def create_zip(src, dst):
 
     Based on https://stackoverflow.com/a/14569017/1558022
     """
-    name = f'{dst}.zip'
-    with zipfile.ZipFile(name, 'w', zipfile.ZIP_DEFLATED) as zf:
+    with zipfile.ZipFile(dst, 'w', zipfile.ZIP_DEFLATED) as zf:
         abs_src = os.path.abspath(src)
         for dirname, subdirs, files in os.walk(src):
             for filename in files:
+                if filename.startswith('.'):
```
kenoir: 👍

alexwlchan:
Merging with approval from @kenoir in Slack.

PR #992, opened 2017-09-27
Migrate some of our Lambdas into the monitoring stack

alexwlchan:

Migrate specific Lambdas into the monitoring stack. At least initially:

* post_to_slack
* gatling_to_cloudwatch
* update_service_list

### Who is this change for?

A platform team who want stacks that make sense.

Part of #980.

- [x] Run `terraform apply`.


alexwlchan:
This is applied in both stacks, so I’d like to merge this soon before it gets blatted by something else.

builds/publish_lambda_zip.py
```diff
@@ -64,7 +64,7 @@ def build_lambda_local(path, name):
 
     # Copy all the associated files to the Lambda directory.
     for f in os.listdir(path):
-        if not f.startswith(('test_', '.', 'requirements.txt')):
+        if not f.startswith(('test_', '.', 'requirements.txt', '__pycache__')):
```
kenoir: 👍

kenoir:
Not just lambdas in here, but still good 👍

kenoir:
Looks like there's a failing test in `master` for the `id_minter` that wasn't introduced by these changes (in addition to the lambda tests that are currently re-running. Good to merge if the lambda tests pass - and we'll address the `id_minter` later.

PR #998, opened 2017-09-28
ReindexTrackerService tracks shards

kenoir:

In order to run reindexes against particular shards the ReindexTrackerService (and its underlying table) need to keep track of the particular shard that we wish to reindex.

### Who is this change for?

Folk who want effective reindexes of individual image collections (and whatever else we choose to shard on in the future).

- [ ] Deployed new versions
- [ ] Run `terraform apply`.

### DIFF
--- a/reindexer/src/test/scala/uk/ac/wellcome/platform/reindexer/services/ReindexTrackerServiceTest.scala
+++ b/reindexer/src/test/scala/uk/ac/wellcome/platform/reindexer/services/ReindexTrackerServiceTest.scala
@@ -23,10 +23,11 @@ class ReindexTrackerServiceTest
   it(
     "should return the index in need of reindexing"
   ) {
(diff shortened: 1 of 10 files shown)


reindexer/src/test/scala/uk/ac/wellcome/platform/reindexer/services/ReindexTrackerServiceTest.scala
```diff
@@ -23,18 +23,20 @@ class ReindexTrackerServiceTest
   it(
```
alicefuzier: I think we should add another test in this Suite that justifies the code change in RendexTrackerService

kenoir: 👍 check that it actually references another shard ...

PR #999, opened 2017-09-28
Increase possible batch size in dynamo_to_sns lambda

kenoir:

In order to speed up the dynamo_to_sns lambda it needs to be able to handle larger batches of records.

### Who is this change for?

🚤 💻 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


lambdas/common/utils/dynamo_utils.py
```diff
@@ -3,22 +3,28 @@
...
+    def create(event):
+        return [DynamoImage(
+            record['dynamodb'],
+            record['eventSourceARN']
+        )for record in event['Records'] if 'NewImage' in record['dynamodb']]
+
+
+class DynamoImage:
+    def __init__(self, record, source_arn):
+        self.record = record
```
alicefuzier: Should this be renamed to `records` to avoid confusion?

kenoir:
@alicefuzier should be ready for review

kenoir:
@alicefuzier should be ready for re-review

PR #1032, opened 2017-10-04
Loris cluster experiments

alicefuzier:
Trying again to make Loris faster. M4 images are memory optimized and the network performance is high. If Loris is slow because of network we should see an improvement here
### Who is this change for?
Anyone using Loris

- [x] Run `terraform apply`.


alicefuzier:
So, small update on what has been going on with this branch. I tried different underlying EC2 instances to prove different hypothesis on what is the bottleneck in Loris.
- First hypothesis: EFS cache is the bottleneck -> tried backing the instances with a EBS cache. There is a slight improvement after leaving it running for a few days but not huge (Check monitoring screenshots below).
- Second hypothesis: network is the bottleneck -> tried a m4.xlarge and a r4.xlarge both of which have a better network provision than the current t2.large. Result: Nope, it's not that
- Third hypothesis: CPU is the bottleneck -> tried a c4.large and a c4.xlarge. c4.large didn't seem to make a difference (if anything it is slower). But c4.xlarge did improve response times as screenshots below show. 
loris_2d_stats
loris_ebs_2d_stats
loris_c4_2d_stats


c4.xlarge is expensive though and I think it might be worth checking if we can reduce Loris CPU usage. Average CPU usage seems to sit around 20-30% of provisioned CPU but there are consistently a lot of spikes that go up to 100% of CPU. One way to try and reduce these spikes might be to reduce the number of wusgi processes in the Loris docker images. It might be that if we have too many, that could cause the CPU to constantly switch between them.


Additionally, combining EBS cache with c4 instances might combine and enhance the 
improvements seen with both of them individually.

So, 

- [ ] I'm going to shut down Loris C4 cluster and migrate the Loris EBS cluster to use C4 instances to see how much that improves things

- [ ] Build another Loris with half the number of processes and check what happens to CPU usage

alicefuzier:
Closing this one and coninuing work in #1039

PR #1035, opened 2017-10-05
Split API and Pipeline stacks

kenoir:

Splits the API and pipeline stacks! Small stacks == less chance of breaking things in anther stack!

### Who is this change for?

💻 

- [ ] Run `terraform apply`.


alexwlchan:
⚡️🎉⚡️

PR #1039, opened 2017-10-09
Experiments to improve Loris performance

alicefuzier:
improve Loris performance in the most cost effective way. Spin off of #1032 
### Who is this change for?
Everyone

- [ ] Deployed new versions

- [ ] Run `terraform apply`.


alicefuzier:
We've decided to stick with C4 xlarge backed with an EBS volume for now. C4.large was not performing well at for the time being we are ok with the c4xlarge cost. Possibly worth revisiting this once we make some changes to loris to keep tthe images in memory instead of passing around temporary files

PR #1048, opened 2017-10-11
Copy Miro master assets 

alicefuzier:
Move miro master assets to their final destination (for now)
### Who is this change for?
👩  👨  👨‍🚀  👩‍🎨 🤸‍♀️ 👨‍🍳  💻 🚶‍♀️ 🐺 

- [ ] Run `terraform apply`.


miro_preprocessor/miro_copy_s3_asset/miro_copy_s3_derivative_asset/src/install_deps.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+pip install -r requirements.txt
```
alexwlchan: This needs

```bash
set -o errexit
set -o nounset
```

at the top so it can bail out if installing requirements fails.

miro_preprocessor/miro_copy_s3_asset/miro_copy_s3_derivative_asset/src/requirements.txt
```diff
@@ -0,0 +1 @@
+wellcome_lambda_utils==0.0.0.4
```
alexwlchan: what is this version number

miro_preprocessor/miro_copy_s3_asset/miro_copy_s3_derivative_asset/src/test_miro_copy_s3_derivative_asset.py
```diff
@@ -171,42 +170,9 @@ def test_should_replace_asset_if_already_exists_with_different_content(
...
-    miro_copy_s3_asset.main(event, None)
+    miro_copy_s3_derivative_asset.main(event, None)
 
     s3_response = s3_client.get_object(Bucket=destination_bucket_name, Key=destination_key)
     assert s3_response['Body'].read() == image_body
 
     assert_sns_message_forwarded(image_json, queue_url, sqs_client)
-
-
-def test_should_not_copy_asset_if_already_exists_with_same_checksum(create_source_and_destination_buckets, sns_sqs,
```
alexwlchan: Why has this test been removed?

miro_preprocessor/s3.tf
```diff
@@ -0,0 +1,20 @@
+resource "aws_s3_bucket" "wellcomecollection-images" {
+  bucket = "wellcomecollection-images"
```
alexwlchan: Is this bucket storing the master assets? If so, is it appropriate that it lives in the Miro preprocessor stack?

shared_infra/lambdas/common/setup.py
```diff
@@ -0,0 +1,25 @@
...
+
+SOURCE = local_file('src')
+
+setup(
+    name='wellcome_lambda_utils',
+    packages=find_packages(SOURCE),
+    package_dir={'': SOURCE},
+    version='0.0.0.4',
+    install_requires=['boto'],
+    python_requires='>=3',
```
alexwlchan: Python 3.6 – the f-strings we‘re using are a very recent feature.

shared_infra/lambdas/common/src/wellcome_lambda_utils/miro_utils.py
```diff
@@ -0,0 +1,20 @@
+import re
+
+
+class MiroImage:
+    def __init__(self, image_info):
+        self.collection = image_info['collection'].split("-")[-1]
+        self.image_data = image_info['image_data']
+
```
alexwlchan: It would be helpful to add a `__repr__` method. This gets called to print the debug representation in tracebacks, interactive console, etc. Right now you just get `<__main__.MiroImage object at 0x107d678d0>`.

Something like:

```python
def __repr__(self):
    image_info = {'collection': self.collection, 'image_data': self.image_data}
    return f'{type(self).__name}(image_info={image_info!r})'
```

Ideally it’s a string you could `eval()` to get an equivalent object.

shared_infra/lambdas/common/src/wellcome_lambda_utils/miro_utils.py
```diff
@@ -0,0 +1,20 @@
...
+    @property
+    def miro_id(self):
+        return self.image_data['image_no_calc']
+
+    @property
+    def shard(self):
+        result = re.match(r"(?P<shard>[A-Z]+[0-9]{4})", self.miro_id)
```
alexwlchan: nice use of named regex groups

shared_infra/lambdas/common/src/wellcome_lambda_utils/s3_utils.py
```diff
@@ -0,0 +1,51 @@
...
+                f"Destination bucket has no image")
+            _copy_image_asset(s3_client, source_identifier, destination_identifier)
+            pass
+        else:
+            raise
+    else:
+        if not source_head_response['ETag'] == destination_head_response['ETag']:
+            print(
+                f"Destination bucket has image with different hash")
+            _copy_image_asset(s3_client, source_identifier, destination_identifier)
```
alexwlchan: I might be tempted to structure this slightly differently – rather than multiple calls to `_copy_image_asset`, instead have one call at the end of the function, and early returns if we detect we can skip the copy. Saves a bit of copying.

And minor idiom note: `if foo != bar` is slightly preferred to `if not foo == bar`.

shared_infra/lambdas/common/src/wellcome_lambda_utils/sns_utils.py
```diff
@@ -39,3 +39,11 @@ def publish_sns_message(sns_client, topic_arn, message):
     assert response['ResponseMetadata']['HTTPStatusCode'] == 200
 
     return response
+
+
+def extract_json_message(event):
+    """
+    Extracts a JSON message from an SNS event sent to a lambda
+    """
+    message = event['Records'][0]['Sns']['Message']
```
alexwlchan: Nice! But I see two places where you’re still repeating this line of code in this diff – could you push it out to other places please?

miro_preprocessor/miro_copy_s3_asset/miro_copy_s3_derivative_asset/src/test_miro_copy_s3_derivative_asset.py
alicefuzier: this is in test_s3_utils now

shared_infra/lambdas/common/src/wellcome_lambda_utils/sns_utils.py
alicefuzier: Tha could go in sns_utils somehow as it's how we read events from sns in _every_ lambda that is triggered by sns. I think that should be a different consolidation patch though

alicefuzier:
Thanks for the comments @alexwlchan . I'm going to address them now. Just to let you know, changes to the lambda utils module are now in a separate patch: #1078 , so I'll address those regarding that code in there. Once that is merged to master I'll fix the conflicts here

miro_preprocessor/miro_copy_s3_asset/miro_copy_s3_derivative_asset/src/requirements.txt
alicefuzier: This was a bogus version number while I was trying to figure out how to publish a package in python

miro_preprocessor/miro_copy_s3_asset/variables.tf
```diff
@@ -1,13 +1,23 @@
...
 variable "lambda_error_alarm_arn" {}
 
 variable "topic_miro_copy_s3_asset_arn" {}
 
-variable "bucket_miro_images_public_arn" {}
+variable "bucket_destination_asset_arn" {}
 
-variable "topic_miro_image_to_dynamo_arn" {}
+variable "topic_forward_sns_message_arn" {
+  default = ""
```
kenoir: remove this?

alicefuzier: It's used by the lambda that copies the derivative, but not by the one that copies the master. That's why it's there but empty by default

kenoir: Maybe add that into the description?

alicefuzier: Yep, makes sense

miro_preprocessor/main.tf
```diff
@@ -55,21 +55,42 @@ module "miro_image_sorter" {
...
+  lambda_error_alarm_arn       = "${local.lambda_error_alarm_arn}"
+  bucket_destination_asset_arn = "${aws_s3_bucket.wellcomecollection-images.arn}"
+  bucket_destination_name      = "${aws_s3_bucket.wellcomecollection-images.id}"
+  bucket_source_asset_arn      = "${local.bucket_miro_images_sync_arn}"
+  bucket_source_asset_name     = "${local.bucket_miro_images_sync_name}"
+
+  destination_key_prefix = "library"
+  lambda_description     = "Copy catalogue miro master assets to private s3 bucket"
+  lambda_name            = "miro_copy_catalogue_master"
+  is_master_asset        = "true"
```
alexwlchan: It’s a small thing, but I’d rather the fields in these two modules were ordered in such a way to highlight the different settings. Put all the keys that are the same at the top, then put the differing keys in a second block.

miro_preprocessor/main.tf
```diff
@@ -55,21 +55,42 @@ module "miro_image_sorter" {
...
+
+  lambda_error_alarm_arn       = "${local.lambda_error_alarm_arn}"
+  bucket_destination_asset_arn = "${local.bucket_miro_images_public_arn}"
+  bucket_destination_name      = "${local.bucket_miro_images_public_name}"
+  bucket_source_asset_arn      = "${local.bucket_miro_images_sync_arn}"
+  bucket_source_asset_name     = "${local.bucket_miro_images_sync_name}"
+
+  lambda_description     = "Copy catalogue miro derivatives to Loris s3 bucket"
+  lambda_name            = "miro_copy_catalogue_derivative"
+  is_master_asset        = "false"
```
alexwlchan: Terraform supports booleans as a first-class type. Let’s use those, rather than strings.

https://www.terraform.io/docs/configuration/syntax.html

miro_preprocessor/miro_copy_s3_asset/miro_copy_s3_derivative_asset/src/miro_copy_s3_derivative_asset.py
```diff
@@ -0,0 +1,50 @@
...
+
+def copy_and_forward_message(s3_client,
+                             sns_client,
+                             image_info,
+                             topic_arn,
+                             source_identifier,
+                             destination_identifier,
+                             source_head_response):
+    s3_utils.copy_asset_if_not_exists(s3_client=s3_client,
+                                      source_head_response=source_head_response,
```
alexwlchan: Passing this around seems decidedly iffy to me; is it really correct to be passing around S3 responses between our functions? Seems like that would make it very hard to test.

I might have missed this function going into the s3_utils module – I think a neater way to do this would be to have `s3_utils.copyfile(src, dst, lazy=False)`. Then the work of looking up this HEAD response is owned entirely by the module, and not exposed to the caller.

miro_preprocessor/miro_copy_s3_asset/miro_copy_s3_derivative_asset/src/miro_copy_s3_derivative_asset.py
```diff
@@ -0,0 +1,50 @@
...
+                                      source_identifier=source_identifier,
+                                      destination_identifier=destination_identifier)
+    sns_utils.publish_sns_message(
+        sns_client,
+        topic_arn,
+        image_info)
```
alexwlchan: nit: keyword arguments for clarity please.

miro_preprocessor/miro_copy_s3_asset/miro_copy_s3_derivative_asset/src/miro_copy_s3_derivative_asset.py
```diff
@@ -0,0 +1,50 @@
...
+
+def main(event, _):
+    print(f"Received event:\n{event}")
+    sns_client = boto3.client("sns")
+    s3_client = boto3.client("s3")
+    source_bucket_name = os.environ["S3_SOURCE_BUCKET"]
+    destination_bucket_name = os.environ["S3_DESTINATION_BUCKET"]
+    topic_arn = os.environ["TOPIC_ARN"]
+
+    image_info = json.loads(event['Records'][0]['Sns']['Message'])
```
alexwlchan: This should use the load_message_wrapper from sns_utils.py.

miro_preprocessor/miro_copy_s3_asset/miro_copy_s3_derivative_asset/src/miro_copy_s3_derivative_asset.py
```diff
@@ -0,0 +1,50 @@
...
+    miro_image = MiroImage(image_info)
+    source_key = f"fullsize/{miro_image.image_path}.jpg"
```
alexwlchan: Should we pass in `fullsize` as an environment variable?

miro_preprocessor/miro_copy_s3_asset/miro_copy_s3_derivative_asset/src/miro_copy_s3_derivative_asset.py
```diff
@@ -0,0 +1,50 @@
...
+    destination_key = f"{miro_image.image_path}.jpg"
```
alexwlchan: Should this be using the S3_DESTINATION_PREFIX environment variable that I think we’ve passed into this Lambda?

miro_preprocessor/miro_copy_s3_asset/miro_copy_s3_master_asset/src/miro_copy_s3_master_asset.py
```diff
@@ -0,0 +1,27 @@
...
+
+    image_info = json.loads(event['Records'][0]['Sns']['Message'])
+    miro_image = MiroImage(image_info)
+    key = f"Wellcome_Images_Archive/{miro_image.collection} Images/{miro_image.image_path}.jp2"
+    print(key)
+    destination_key = f"{destination_prefix}/{miro_image.image_path}.jp2"
+    source_identifier = S3_Identifier(source_bucket_name, key)
+    destination_identifier = S3_Identifier(destination_bucket_name, destination_key)
+    s3_utils.exec_if_key_exists(s3_client, source_identifier=source_identifier,
+                                function=partial(s3_utils.copy_asset_if_not_exists, destination_identifier=destination_identifier))
```
alexwlchan: This looks similar to `miro_copy_s3_derivative_asset.py`, but I thought the idea was that we’d have two Lambdas feeding from the same source tree?

miro_preprocessor/main.tf
alicefuzier: Yeah I did try that, but I couldn't get this to work properly with booleans 
```
locals {
  source_prefix  = "${var.is_master_asset == "true" ? "Wellcome_Images_Archive": "fullsize/"}"
  lambda_s3_type = "${var.is_master_asset == "true" ? "master": "derivative"}"
}
```

miro_preprocessor/main.tf
alicefuzier: 👍

miro_preprocessor/miro_copy_s3_asset/miro_copy_s3_derivative_asset/src/miro_copy_s3_derivative_asset.py
alicefuzier: which sns_utils.py? The one in common doesn't have it

miro_preprocessor/miro_copy_s3_asset/miro_copy_s3_derivative_asset/src/miro_copy_s3_derivative_asset.py
alicefuzier: maybe, but derivative assets will always come from the `fullsize` "folder" and that might give the wrong impression that you can get them from somewhere else, so I'm in favour of hardcoding it

miro_preprocessor/miro_copy_s3_asset/miro_copy_s3_derivative_asset/src/miro_copy_s3_derivative_asset.py
alicefuzier: 👍

miro_preprocessor/miro_copy_s3_asset/miro_copy_s3_master_asset/src/miro_copy_s3_master_asset.py
alicefuzier: I'm not sure I follow. Initially, the idea was to instantiate two lambdas from the same code, but it quickly turned out to be tricky to make it generic enough as there are significant differences in where the source images are for master and derivative assets. 
So, what I've done in this PR is having a different source code for copying master and derivatives respectively. Both of those can then be instantiated as many times as we need. For example, the copy master asset will have to be instantiated three times (digital library, tandem_vault, cold_store)

miro_preprocessor/miro_copy_s3_asset/miro_copy_s3_derivative_asset/src/miro_copy_s3_derivative_asset.py
alicefuzier: 👍  I'll use that from #1086 when it's ready

PR #1050, opened 2017-10-11
Pin Loris dependencies

kenoir:

In order to avoid builds breaking when they are advanced upstream inadvertently.

### Who is this change for?

Developers who want passing builds.

- [ ] Deployed new versions


loris/Dockerfile
```diff
@@ -1,8 +1,13 @@
-FROM ubuntu:14.04
+FROM ubuntu:16.04
 
 RUN apt-get update && \
-    apt-get install --yes python python-pip python-setuptools python-dev \
-    uwsgi uwsgi-plugin-python
+    apt-get install --yes python uwsgi uwsgi-plugin-python wget
+
+# Install pip.  We don't use pip from the Ubuntu package repositories
```
kenoir: 👍

PR #1063, opened 2017-10-12
Send WIA winners to Tandem Vault

### LINKED ISSUE #1058: Sort Miro images for Wellcome Image Awards
Miro images that are Wellcome Image Awards winners are going to be separately organised in Tandem Vault to give staff easy browsing and access to previous winners.

To do this we should:

* Create another path for these images (or re-use the Tandem Vault one if we can treat them differently on import into Tandem Vault)
* Match images based on the `image_award` field - note that this should match both "Biomedical Image Awards" and "Wellcome Image Awards"

Images that match this rule should still be matched by other rules as appropriate.

alexwlchan:

All images that won WIA get sent to Tandem Vault in addition to anywhere else.

Resolves #1058.

### Who is this change for?

Folks who want correctly sorted images.

- [ ] Deployed new versions

### DIFF
--- a/miro_preprocessor/Makefile
+++ b/miro_preprocessor/Makefile
@@ -73,7 +73,7 @@ image_sorter-test: $(ROOT)/.docker/python3.6_ci
 	$(ROOT)/builds/docker_run.py --aws -- \
 		--volume $(ROOT)/miro_preprocessor/image_sorter/src:/data \
 		--env OP=test \
-		--env FIND_MATCH_PATHS="/data" \
+		--env FIND_MATCH_PATHS="/data" --tty \
 		python3.6_ci:latest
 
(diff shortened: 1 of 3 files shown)


miro_preprocessor/Makefile
```diff
@@ -73,7 +73,7 @@ image_sorter-test: $(ROOT)/.docker/python3.6_ci
 	$(ROOT)/builds/docker_run.py --aws -- \
 		--volume $(ROOT)/miro_preprocessor/image_sorter/src:/data \
 		--env OP=test \
-		--env FIND_MATCH_PATHS="/data" \
+		--env FIND_MATCH_PATHS="/data" --tty \
```
kenoir: Why this?

Issue #131, opened 2017-03-31
Create API ontology for ResultList and Error


jtweed:
We need to define types that only exist in the API.

Initially these are ResultSet (for a list response) and Error (for error responses).

PR #1066, opened 2017-10-12
Lint our Turtle ontologies in CI

alexwlchan:
Builds on #1065. A precursor to #131, because I don't trust myself to write valid TTL! And indeed, it threw up one error in my error ontology.


kenoir:
Why is CI not running here?

alexwlchan:
> Why is CI not running here?

Because it was being merged into a branch (`pythonci`) where we don’t run tests.

alexwlchan:
What this exposes is that apparently JS linting wasn't running previously? Sadness.

PR #1068, opened 2017-10-12
Write Miro Image data to ES Cluster

kenoir:

Keep track of where Miro images are ending up after being sorted in a way that's accessible and searchable.

### Who is this change for?

💻 

- [ ] Run `terraform apply`.

### DIFF
--- a/miro_preprocessor/image_sorter/src/test_miro_image_sorter.py
+++ b/miro_preprocessor/image_sorter/src/test_miro_image_sorter.py
@@ -76,11 +76,17 @@ def _get_msg(sqs_client, queue_url):
     )
 
     message_body = messages['Messages'][0]['Body']
-    inner_message = json.loads(message_body)['Message']
+    json_decoded_message = json.loads(message_body)
 
-    return json.loads(
-        json.loads(inner_message)['default']
-    )
+    inner_message = json_decoded_message['Message']
+    subject = json_decoded_message['Subject']
+
+    return {
+        "subject": subject,
+        "message": json.loads(
+            json.loads(inner_message)['default']
+        )
+    }
 
 
 def collection_image_data(**kwargs):
@@ -144,7 +150,7 @@ def test_image_sorter_catalogue_api(image_sorter_sns_sqs, s3_put_event):
 
     catalogue_api_msg = _get_msg(sqs_client, sns_sqs["catalogue_api"]["queue"])
 
-    assert catalogue_api_msg == metadata
+    assert catalogue_api_msg['message'] == metadata
 
 
 @mock_s3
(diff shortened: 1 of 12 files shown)


kenoir:
Fixes: https://github.com/wellcometrust/platform/issues/1053

miro_preprocessor/image_sorter/src/test_miro_image_sorter.py
```diff
@@ -76,11 +76,17 @@ def _get_msg(sqs_client, queue_url):
...
-        json.loads(inner_message)['default']
-    )
+    inner_message = json_decoded_message['Message']
+    subject = json_decoded_message['Subject']
+
+    return {
+        "subject": subject,
+        "message": json.loads(
+            json.loads(inner_message)['default']
+        )
```
alexwlchan: Why modify the return structure if you never assert on anything but the subject?

miro_preprocessor/miro_image_to_dynamo/main.tf
```diff
@@ -10,7 +10,7 @@ module "miro_image_to_dynamo_lambda" {
 
   environment_variables = {
     TABLE_NAME      = "${var.miro_data_table_name}"
-    REINDEX_VERSION = 1
+    REINDEX_VERSION = 2
```
alexwlchan: ???

miro_preprocessor/miro_inventory/main.tf
```diff
@@ -0,0 +1,60 @@
+module "miro_inventory_lambda" {
+  source = "git::https://github.com/wellcometrust/terraform.git//lambda?ref=v1.0.0"
```
alexwlchan: nit: cleaner just to do `github.com/wellcometrust`, you don't need the `git::https://` prefix

miro_preprocessor/miro_inventory/src/miro_inventory.py
```diff
@@ -0,0 +1,79 @@
+# -*- encoding: utf-8 -*-
+"""
+Description
```
alexwlchan: 😞

miro_preprocessor/miro_inventory/src/miro_inventory.py
```diff
@@ -0,0 +1,79 @@
...
+    post_url = f'{es_cluster_url}/{es_index}/{es_type}'
+
+    print(f'POSTing to {post_url}')
+
+    ascii_token = token.decode('ascii')
+
+    return requests.post(
+        post_url,
+        data=payload,
+        headers={'Authorization': f'Basic {ascii_token}'}
```
alexwlchan: Try:

```python
return requests.post(
   post_url,
   data=payload,
   auth=(es_username, es_password)
)
```

Because requests has builtin support for HTTP Basic Auth – is that what you’re trying to do here?

miro_preprocessor/image_sorter/src/test_miro_image_sorter.py
kenoir: I want to, there's a but there's a bug in moto which hard codes the value.

miro_preprocessor/miro_inventory/main.tf
kenoir: I _think_ this is a necessity when using the ref param (at least the terraform docs indicate so).

kenoir:
Going to merge this now as I want to be able to demo it later ...

PR #1073, opened 2017-10-13
Send holding images straight to cold store

### LINKED ISSUE #1070: Exclude holding images from the Catalogue API
We need to exclude the “holding images” that are duplicate image records from Miro.

Waiting for Christy to tell us exactly what this rule looks like.

alexwlchan:

Remove duplicate images which are holding thumbnails for other records.

Resolves #1070.

### Who is this change for?

Folks who don’t want duplicate, downsized images in their API.

- [ ] Deployed new versions

### DIFF
--- a/miro_preprocessor/image_sorter/src/sorter_logic.py
+++ b/miro_preprocessor/image_sorter/src/sorter_logic.py
@@ -184,6 +184,13 @@ def _get_decisions_from_id_exceptions(exceptions, image_data):
             return [getattr(Decision, key) for key, value in exception.items()
                     if value is not "" and not value.strip().lower() == "false"]
 
+    # There are "holding images" in MIRO, which are thumbnails put into
+    # duplicate image records for some explicit AIDS posters.  All the
+    # posters are available, so we delete these records.  They all have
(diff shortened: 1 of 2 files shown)


miro_preprocessor/image_sorter/src/sorter_logic.py
```diff
@@ -184,6 +184,13 @@ def _get_decisions_from_id_exceptions(exceptions, image_data):
             return [getattr(Decision, key) for key, value in exception.items()
                     if value is not "" and not value.strip().lower() == "false"]
 
+    # There are "holding images" in MIRO, which are thumbnails put into
+    # duplicate image records for some explicit AIDS posters.  All the
+    # posters are available, so we delete these records.  They all have
+    # image numbers in the L sequence ending with "FX", e.g. "L0052198FX".
+    # if re.match(r'^L\d+FX$', image_data['image_no_calc']):
```
kenoir: Is this supposed to be commented out?

alexwlchan: 🙈

PR #1080, opened 2017-10-17
Fixes for recent Lambda alarms

alexwlchan:

Two alarms we’ve seen recently:

* Timeouts in service_deployment_status
* 429 Rate Limit Exceeded in miro_inventory

So let’s fix them both!

- [ ] Deployed new versions
- [ ] Run `terraform apply`.

### DIFF
--- a/miro_preprocessor/miro_inventory/src/miro_inventory.py
+++ b/miro_preprocessor/miro_inventory/src/miro_inventory.py
@@ -5,6 +5,7 @@ Lambda which passes on miro_image data from a SNS topic to an Elasticsearch clus
 
 import json
 import os
+import time
 
 from botocore.vendored import requests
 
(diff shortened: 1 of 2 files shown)


miro_preprocessor/miro_inventory/src/miro_inventory.py
```diff
@@ -75,4 +76,17 @@ def main(event, _):
...
+    # load on the cluster.  So if we detect an error, insert an artificial
+    # delay to slow down the Lambda rescheduler, and reduce cluster pressure.
+    #
+    # Note: if this hits the timeout instead, the Lambda will still register
+    # as failed and be rescheduled, so this doesn't cause us to lose events.
+    try:
+        response.raise_for_status()
+    except Exception:
+        time.sleep(10)
+        raise
```
alexwlchan: This Lambda currently has a timeout of 120s (!), so I don’t think this will make much of a difference.

kenoir: I don't think this is necessary since the cluster size has been increased. The solution is to _run less through the pipeline in any case. Also what @alexwlchan said.

PR #1081, opened 2017-10-17
Copy master assets that have key containing the id and a suffix

alicefuzier:
Master assets keys in S3 are in the form:`Wellcome_images_archive/A Images/${miro_id}${something}.jp2`
`something` can be:
- empty (most cases)
- a suffix starting with a hyphen: i.e. `-RA-RA`
- a suffix with letters and hyphens (potentially?)

Miro Ids don't have the same length so we have situations where assignment a key to an id is ambiguous: 
```
id1 = "A1234567"
id2 = "A1234567EL"

key = "Wellcome_images_archive/A Images/A1234567EL-RA-RA.jp2"
```

When we try to match an id to a key we cannot just assume that we can build the key from the id as we do for the derivative assets. 
So, the solution in this PR:

- lists all of the objects that match the `Wellcome_images_archive/A Images/${miro_id}` prefix
- try to find the exact match in the returned keys: i.e. `Wellcome_images_archive/A Images/${miro_id}.jp2`
- if there is no exact match => try to find an exact match before a hyphen: i.e `Wellcome_images_archive/A Images/${miro_id}-RA-RA.jp2`
- if there is no exact match nor an exact match before a hyphen => throw an exception

There was another condition which was that, if there was only one key matching the prefix, that key would be assigned to the id regardless of the structure of the key. This condition has been removed as it could cause images for child miro images to be assigned to the parent as well as the child.


miro_preprocessor/miro_copy_s3_asset/miro_copy_s3_derivative_asset/src/install_deps.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
kenoir: Does this file still need to exist?

PR #1085, opened 2017-10-17
Move building and publishing lambda-utils package into its own project

kenoir:


lambda_utils/Makefile
```diff
@@ -0,0 +1,29 @@
...
+		--env FIND_MATCH_PATHS="/data" \
+		python3.6_ci:latest
+
+lambda_utils-build:
+	docker run \
+		--volume $(LAMBDA_UTILS):/src \
+		--workdir /src \
+		python:3-alpine python setup.py sdist
+
+lambda_utils-publish: lambda_utils-build
```
kenoir: This should be `deploy`

alexwlchan: Mmm. It’s publishing a new version of something to a package repository, but not actually deploying it anywhere in our stack. I prefer "publish", a la ECR publishing.

kenoir: It should be publish, but our build system relies on this being "deploy".

kenoir: No, strike that run_task.py says otherwise.

Issue #1054, opened 2017-10-11
Sierra adapter prototype


jtweed:

alexwlchan:
Earlier script: https://gist.github.com/alexwlchan/d0e9886a4429284c57f17a6505283f48

PR #1092, opened 2017-10-19
Add a skeleton stack for the Sierra adapter

alexwlchan:

Lay down the basic pieces:

* Source code and Dockerfile
* Terraform stack
* Travis config for testing

Note that testing is quite different to how we usually do it – rather than using the Python 3.6 CI image (which then needs to know about all the dependencies of projects it tests), instead we build a new `sierra_adapter:test` image that subclasses the Docker image that runs in prod, adds py.test, and runs that.

### Who is this change for?

Us, as we work on #1054.

- [x] Deployed new versions
- [x] Run `terraform apply`.


.travis/run_task.py
```diff
@@ -54,6 +54,7 @@ def main():
         'miro_preprocessor-test',
         'lambdas-test',
         'monitoring_lambdas-test',
+        'sierra_adapter-test',
```
kenoir: Don't like this list, but not an issue for merge.

sierra_adapter/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM python:3.6-alpine
+
+LABEL maintainer "Alex Chan <a.chan@wellcome.ac.uk>"
```
kenoir: As you're not going to be the sole maintainer, probably best to make this a team rather than an individual.

.travis/run_task.py
alexwlchan: Yeah, we should settle this publish/deploy disparity and then we can axe this.

sierra_adapter/Dockerfile
alexwlchan: Good point.

sierra_adapter/terraform/generate_tfvars.sh
```diff
@@ -0,0 +1,28 @@
...
+# Name of tfvars file
+TF_VARS=terraform.tfvars
+RELEASE_IDS_FILE="release_ids.tfvars"
+
+# Ensure we don't have stale variables from a previous run
+rm -f $TF_VARS
+rm -rf "$RELEASE_DIR"
+
+echo "Getting variables from S3"
+aws s3 cp s3://platform-infra/terraform-miro-preproc.tfvars terraform.tfvars
```
kenoir: sierra?

alexwlchan: 🙈

kenoir:
Travis is sad because pylint: ./sierra_adapter/src/test_adapter.py:3:1

PR #1095, opened 2017-10-19
Another step to consolidate all common util functions

kenoir:

Add all s3_utils functions in use across the project to common utils package.

### Who is this change for?

💻 

- [ ] Deployed new versions


lambda_utils/src/wellcome_lambda_utils/s3_utils.py
```diff
@@ -1,5 +1,7 @@
 # -*- encoding: utf-8 -*-
 
+import dateutil.parser
```
alexwlchan: You need to add _dateutil_ to the _install_requires_ for the common package.

lambda_utils/src/wellcome_lambda_utils/s3_utils.py
```diff
@@ -53,3 +55,23 @@ def should_copy():
...
+    event_datetime = dateutil.parser.parse(record["eventTime"])
+
+    return {
+        "event_name": record["eventName"],
+        "event_time": event_datetime,
+        "bucket_name": record["s3"]["bucket"]["name"],
+        "object_key": record["s3"]["object"]["key"],
+        "size": record["s3"]["object"]["size"],
+        "versionId": record["s3"]["object"].get("versionId")
+    }
```
alexwlchan: There’s potentially a neater way to do this with something called _descriptors_ – I’ll show you both later.

lambda_utils/src/wellcome_lambda_utils/s3_utils.py
kenoir: 👍

PR #1101, opened 2017-10-19
Annotate where images that pass through the miro preprocessor go

kenoir:

Clarity about the location of master images from Miro

### Who is this change for?

@jtweed


kenoir:
Fixes: https://github.com/wellcometrust/platform/issues/1098

PR #1102, opened 2017-10-19
Fix the lambda_utils-publish task

kenoir:

Ensure we can publish the `lambda_utils` package to PyPi

### Who is this change for?

Folk who want to make use of our shared `lambda_utils` package 

- [ ] Deployed new versions


kenoir:
Worth noting we won't be _sure_ this works in Travis until it's merged.

PR #1103, opened 2017-10-20
Some more Loris experiments

alicefuzier:
After our tests we've decided that the c4.xlarge backed by EBS volume cache is the best one for now.
Each Ec2 instance can fit only one Loris task because of Loris CPU usage. That means that, to have 4 tasks running we need 4 instances.
The on demand price for c4.xlarge is $0.226000, so the overall cost per year of each instance would be 0.226000 * 24 * 365 = $1979.76 which means a total of $1979.04 * 4= $7919.04

The EBS volume currently in use is an IO1 volume with 2000 IOPS. The cost of the volume is $0.138 per GB-month of provisioned storage plus $0.072 per provisioned IOPS-month, which means 0.138 * 180 + 0.072 * 2000 = 24.84 + 144 = $168.84 per volume per month. 4 volumes (a volume per instance) would be 168.84 * 4 * 12= $8104.32 per year

I think the volumes cost, at least, could be reduced quite a bit, so I want to keep two clusters running to experiment on that

### DIFF
--- a/loris/terraform/asg.tf
+++ b/loris/terraform/asg.tf
@@ -11,11 +11,16 @@ module "loris_cluster_asg" {
   asg_max     = "4"
 
   image_id      = "${data.aws_ami.stable_coreos.id}"
-  instance_type = "t2.xlarge"
+  instance_type = "c4.xlarge"
 
   sns_topic_arn         = "${local.ec2_terminating_topic_arn}"
   publish_to_sns_policy = "${local.ec2_terminating_topic_publish_policy}"
   alarm_topic_arn       = "${local.ec2_instance_terminating_for_too_long_alarm_arn}"
+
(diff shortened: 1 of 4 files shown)


alicefuzier:
I thought it was good to get it into master so that the changes to loris cluster don't get blatted, but it's probably still worth for @jtweed to read the above cost breakdown.

I didn't really look into the EBS cost when we started doing the experiments and just gave it as much as possible, but it turns out that the volume is as expensive as the instance itself and it's definitely overprovisioned, so I think it's worth evaluating the impact

jtweed:
Agreed, I didn't realise the EBS cost was so high or that we have 4 loris instances. It's definitely worth doing some work to ensure the EBS cost is optimal. Also, do we need 4 loris instances running all the time?

PR #1105, opened 2017-10-20
Refines functionality of miro_inventory lambda

kenoir:

Modify `miro_inventory` lambda to record _decisions_ and not images in order that we can record data on the decisions themselves.

### Who is this change for?

💻 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


kenoir:
@alicefuzier if you're adding a copy asset from the none topic, we'll need to coordinate ensuring that is recorded in the inventory.

PR #1106, opened 2017-10-20
None bucket

### LINKED ISSUE #1099: Copy images from the None category into an S3 bucket

alicefuzier:
Copy master assets for which we couldn't make a decision in their own bucket. Resolves #1099 
### Who is this change for?
Everyone

- [x] Run `terraform apply`.


miro_preprocessor/topics.tf
```diff
@@ -27,3 +27,8 @@ module "none_topic" {
   source = "git::https://github.com/wellcometrust/terraform.git//sns?ref=v1.0.0"
   name   = "miro_none"
 }
+
+module "none_topic_forward" {
```
kenoir: I don't think you need this.

alicefuzier: 🤦‍♀️

alexwlchan:
This is only broken in the format task, so I’m going to merge and deploy this so we can re-run the pipeline.

PR #1111, opened 2017-10-23
End loris experiments

alicefuzier:
Settle for c4.xlarge with 180 gb gp2 EBS volume for Loris. 
The on demand price for c4.xlarge is $0.226000, so the overall cost per year of each instance would be 0.226000 * 24 * 365 = $1979.76 which means a total of $1979.04 * 4= $7919.04
The EBS volume cost is $0.11 per GB-month of provisioned storage so one volume is 0.11 * 180= $19.8 per month. 4 volumes (one for each instance) will be 19.8 * 4 = $79.2 per month which means 79.2 * 12 = $950.4 per year.

In total Loris provision will come at about 7919.04 + 950.4 = $8869.44 per year

- [x] Run `terraform apply`.

### DIFF
--- a/loris/terraform/alb.tf
+++ b/loris/terraform/alb.tf
@@ -13,19 +13,3 @@ module "loris_alb" {
 
   alb_access_log_bucket = "${local.bucket_alb_logs_id}"
 }
-
-module "loris_alb_ebs" {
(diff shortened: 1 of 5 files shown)


kenoir:
Don't forget to change the Route53 rules!

jamesgorrie:
@alicefuzier Do you have the perf results? It'd be great to run through them.

PR #1134, opened 2017-10-27
Add versioning to master assets bucket 

alicefuzier:
What the title says (and increase time before master assets get sent to glacier as asked by the production team)
### Who is this change for?
Everyone

- [x] Run `terraform apply`.


kenoir:
👍

PR #1144, opened 2017-11-03
Add non-MPL references from Miro

### LINKED ISSUE #1129: Add non-MPL references from Miro

alexwlchan:

Expose other legacy identifiers in the API content.

Resolves #1129.

### Who is this change for?

Folks who have old references lying around.

- [ ] Deployed new versions


alexwlchan:
Moving back to WIP as the zipping-two-fields logic will be useful for the WIA stuff. It’s already a bit hacky, so I’d like to do it properly.

alexwlchan:
And now that’s done. Ready for review.

PR #1146, opened 2017-11-03
Add information about Wellcome Image Awards winners to descriptions

### LINKED ISSUE #1100: Expose Wellcome Image Awards field from MIRO in the API
From @jennpb:
Would it be possible to expose this MIRO field in the API?
Usage History > Wellcome Exhibitions > Exhibitions > e.g. Wellcome Image Awards 2012

It doesn’t have to be visible by 6 November, but it’d be nice if it could be by the end of the quarter.

alexwlchan:

Resolves #1100.

### Who is this change for?

API users who want to know about Award-Winning Images.

- [ ] Deployed new versions

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/models/transformable/miro/MiroTransformable.scala
+++ b/common/src/main/scala/uk/ac/wellcome/models/transformable/miro/MiroTransformable.scala
@@ -127,7 +127,7 @@ case class MiroTransformable(MiroID: String,
     // Finally, remove any leading/trailing from the description, and drop
     // the description if it's *only* whitespace.
     val description =
-      if ((rawDescription + wiaAwardsString).trim.length > 0) {
+      if ((rawDescription + wiaAwardsString).trim.isNotEmpty) {
         Some((rawDescription + wiaAwardsString).trim)
       } else None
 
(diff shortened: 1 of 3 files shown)


alexwlchan:
This is a bit of a hack job; doing this in the truly general case is both dull and unnecessary.

common/src/main/scala/uk/ac/wellcome/models/transformable/miro/MiroTransformable.scala
```diff
@@ -85,9 +85,50 @@ case class MiroTransformable(MiroID: String,
...
+      }
+
+      // Any more than two award-related entries in these fields would be
+      // unexpected, and we let it error as an unmatched case.
+    }
+
+    // Finally, remove any leading/trailing from the description, and drop
+    // the description if it's *only* whitespace.
     val description =
-      if (rawDescription.trim.length > 0) {
```
kenoir: `rawDescription.trim.isNotEmpty`

alexwlchan: 👌

alexwlchan:
I think this is ready for another review. Once that’s done, we can get this merged and deployed ahead of the reindex on Monday.

Issue #982, opened 2017-09-23
Who is asking for PNGs as JPEGs?


alexwlchan:
We keep getting Loris alarms in Slack about somebody trying to request images like this:

/wordpress:2016/04/viral-nova.png/full/320,/0/default.jpg

If you try to fetch that image from Loris, it complains because it can’t convert an RGBA PNG to an RGB JPEG. What should it do with the transparent regions? The IIIF Image spec doesn’t seem to mention this possibility.

That it’s appearing early in the morning (UK time) suggests that maybe this is an automated crawler?

Whatever it is, we should find out and shut it up, so it stops spamming Slack with alarms!

alexwlchan:
Inspecting the ALB logs from today’s error:

```
760097843905_elasticloadbalancing_eu-west-1_app.api.e87a4a5f32874d8b_20170926T0730Z_52.51.25.193_5f7glel8.log:531:https 2017-09-26T07:28:03.894196Z app/api/e87a4a5f32874d8b 35.167.191.134:27087 10.30.0.99:32790 0.001 0.219 0.000 500 500 527 483 "GET https://iiif-origin.wellcomecollection.org:443/image/wordpress:2016/12/victoria_and_albert_christmas_tree.png/full/320,/0/default.jpg HTTP/1.1" "Mozilla/5.0 (compatible; bingbot/2.0; +http://www.bing.com/bingbot.htm)" ECDHE-RSA-AES128-GCM-SHA256 TLSv1.2 arn:aws:elasticloadbalancing:eu-west-1:760097843905:targetgroup/loris/137b9ef8b7bbfbf8 "Root=1-59ca0183-578d8d20733d06387e5799c3" "iiif-origin.wellcomecollection.org" "session-renegotiated-or-reused"
```

and pulling out the user agent string:

```
Mozilla/5.0 (compatible; bingbot/2.0; +http://www.bing.com/bingbot.htm)
```

Blergh. Wait it out, I guess? 😕

alexwlchan:
At this point, these alerts have died down, and we have enough alarm visibility to see if this is the cause of a particular alarm or something else. I don’t think we need to do anything else here for now.

alexwlchan:
Closing again, I haven’t seen these errors in quite a while.

PR #1147, opened 2017-11-03
Return 410s on certain URLs to shut up Bingbot

alexwlchan:
Another go at #982.

### DIFF
--- a/docker/nginx/loris.nginx.conf
+++ b/docker/nginx/loris.nginx.conf
@@ -31,6 +31,22 @@ http {
 
     add_header Cache-Control "max-age=31536000;";
 
+    # A while back, the Collection site started serving all content through
+    # Loris, but requesting JPG format.  This breaks when you request a
+    # transparent PNG, because it falls over trying to fill in the transparent
+    # regions, which triggers a 500 alarm in Slack.
+    #
+    # But those URLs don't exist any more, right?
+    #
+    # Nope, Bingbot keeps requesting them.  So this is meant to shut it up,
+    # eventually we'll remove these, but 410s should tell it to go away.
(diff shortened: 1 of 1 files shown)


docker/nginx/loris.nginx.conf
```diff
@@ -31,6 +31,22 @@ http {
...
+    # A while back, the Collection site started serving all content through
+    # Loris, but requesting JPG format.  This breaks when you request a
+    # transparent PNG, because it falls over trying to fill in the transparent
+    # regions, which triggers a 500 alarm in Slack.
+    #
+    # But those URLs don't exist any more, right?
+    #
+    # Nope, Bingbot keeps requesting them.  So this is meant to shut it up,
+    # eventually we'll remove these, but 410s should tell it to go away.
+    location /image/wordpress:2016/12/victoria_and_albert_christmas_tree.png/full/320,/0/default.jpg {
```
kenoir: Should we be serving pngs at all? can we make this a pattern for all pngs?

kenoir: Or maybe something like this: https://stackoverflow.com/questions/8237761/is-it-possible-to-change-the-http-status-code-returned-when-proxy-pass-gateway-i

alexwlchan: We do serve PNGs, and most of them would serve fine as JPEGs – if they don’t have transparent regions. This is just a short-term hack to get Bingbot to leave these URLs alone.

kenoir: Yup. I just don't like adding these one off things to the nginx config, with no exact timescale for removing them.

PR #1151, opened 2017-11-06
Move the Turtle linter into its own container

alexwlchan:
If you follow the “one job per image” model, this is a bit neater. It should also make builds that only require one or the other a bit faster, as both images now have a smaller dependency footprint.


kenoir:
🐢

PR #1153, opened 2017-11-07
Write a basic ResultList ontology

### LINKED ISSUE #131: Create API ontology for ResultList and Error

alexwlchan:
Resolves #131.


ontologies/Schema/resultlist.ttl
```diff
@@ -0,0 +1,37 @@
...
+
+<http://wellcomecollection.org/ontologies/resultlist>
+    rdf:type owl:Ontology ;
+    dcterms:title "Result list ontology"@en ;
+    dcterms:created "2017-11-07"^^xsd:date ;
+    dcterms:description "Ontology describing the properties of a result list"@en .
+
+wresultlist:ResultList rdf:type owl:Class ;
+    rdfs:label "Result list"@en ;
+    rdfs:comment "A collection of works."@en ;
```
jamesgorrie: Only used for works?
Why not a WorksList then?

silveroliver: Its a good point. Would we use result lists for lists of items, holdings and potentially concepts in the future. @jamesgorrie I will pick up with platform team tomorrow.

ontologies/Schema/resultlist-ontology.ttl
```diff
@@ -15,13 +14,18 @@
...
 wresultlist:ResultList rdf:type owl:Class ;
     rdfs:label "Result list"@en ;
-    rdfs:comment "A collection of works."@en ;
+    rdfs:comment "A collection of things."@en ;
     rdf:isDefinedBy <http://wellcomecollection.org/ontologies/resultlist> .
 
+wresultlist:Thing rdf:type owl:Class ;
+    rdfs:label "Thing"@en ;
+    rdfs:comment "A thing"@en ;
+    rdf:isDefinedBy <http://wellcomecollection.org/ontologies/wresultlist:> .
```
jtweed: Stray colon.

alexwlchan: Thanks, fixed.

alexwlchan:
I’ve discussed this with @silveroliver; we’re both happy with this as-is.

Issue #1016, opened 2017-09-29
Ingest uncatalogued images into Tandem Vault


jtweed:
Preparatory work:

- [x] Create upload sets, one for each Miro prefix
- [x] Create collections, one for each Miro prefix
- [x] Create collections, one for each year of the Wellcome Image awards

Upload step:

- [x] Logic to determine if an image is a WIA, and if so, which year
- [x] Code for uploading an image from our S3 bucket into the Tandem Vault S3 bucket
- [x] Code for sorting an uploaded image into the Miro collection
- [x] Code for sorting an uploaded image into the WIA award collection, if necessary
- [x] Code for adding the appropriate Miro metadata to an asset

kenoir:
See https://github.com/wellcometrust/platform/issues/1058 for WIA criteria.

kenoir:
Talk to @alexwlchan about > Code for adding the appropriate Miro metadata to an asset

jtweed:
```
year,collection
1997,110075
1998,110076
1999,110077
2001,110078
2002,110079
2005,110080
2006,110081
2008,110082
2009,110083
2011,110084
2012,110085
2014,110086
2015,110087
2016,110088
2017,110089
```

kenoir:
@alexwlchan adds:

```py
import requests

resp = requests.put(
    'https://wellcome.tandemvault.com/api/v1/assets/2965980',
    params={
        'api_key': '0fd483ff51ba9dbea0bfe497b788594c2ce9578088cff6ecb9e33b87cda501e1',
        'asset[description]': 'First description by Alex',
        'asset[usage_terms]': 'These are my usage terms'
    }
)
resp.text
resp
```

kenoir:
Two things remain:
- Provide a list of WIA winners from the enrichment queue.
- Provide a list of missing source files from the upload queue

kenoir:
Calling this one done.

PR #1155, opened 2017-11-07
Check for asset existence before upload

kenoir:

Make TV uploads idempotent(ish)

### Who is this change for?

💻 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


kenoir:
For https://github.com/wellcometrust/platform/issues/1016

miro_preprocessor/miro_image_to_tandem_vault/src/tandem_vault_uploader.py
```diff
@@ -64,12 +64,27 @@ def main():
...
-                s3_client,
-                src_bucket,
-                miro_image
-            )
+            assets = api.get_assets(miro_image.miro_id)
+            num_assets = len(assets)
+
+            if num_assets > 1:
+                raise Exception(
+                    f"No exact match for {miro_image.miro_id} (found {num_assets} matches)."
```
alexwlchan: Rather than including the number, why not include the list?

PR #1157, opened 2017-11-07
No copyright when image_source_code is WEL

kenoir:

"Wellcome Collection" is _never_ valid copyright when referring to `image_source_code` so reflect that.

### Who is this change for?

@jennpb 

- [ ] Deployed new versions
- [ ] Run reindex

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/models/transformable/miro/MiroTransformable.scala
+++ b/common/src/main/scala/uk/ac/wellcome/models/transformable/miro/MiroTransformable.scala
@@ -308,6 +308,7 @@ case class MiroTransformable(MiroID: String,
       case None =>
         miroData.sourceCode match {
           case Some(code) => Some(contributorMap(code.toUpperCase))
+          case Some("WEL") => None
           case None => None
         }
     }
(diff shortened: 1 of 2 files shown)


alicefuzier:
In our contrib_map.json there are more codes that map to Wellcome Collection. Do we want to exclude those as well?

jennpb:
From a UI point of view, we need to be able to distinguish between Copyright and Source. For the vast majority of images, Wellcome Collection will be Source, but not Copyright.

For non-Wellcome Collection images, there may be 
- a copyright holder, e.g. Royal Army Medical Corps AND
- a Source, namely Wellcome Collection

For most of Wellcome Collection historical images, we are most likely to have no information for copyright. 

By keeping these two distinct, we can provide correct copyright information while also linking the image back to Wellcome Collection.

kenoir:
> By keeping these two distinct, we can provide correct copyright information while also linking the image back to Wellcome Collection. 

Paging @silveroliver to this thread!

kenoir:
Closing until we get clarification and agreement on what's required.

PR #1158, opened 2017-11-08
Dockerise end taskicise sqs_redrive

alicefuzier:
Wrap the sqs_redrive script in an ECS task
### Who is this change for?
Lazy devs

- [x] Deployed new versions
- [x] Run `terraform apply`.


shared_infra/sqs_redrive/run.py
```diff
@@ -0,0 +1,47 @@
...
+        QueueUrl=queue_url,
+        MessageBody=msg
+    )
+
+
+def delete_message(sqs_client, msg, queue_url):
+    sqs_client.delete_message(
+        QueueUrl=queue_url,
+        ReceiptHandle=msg['ReceiptHandle']
+    )
```
alexwlchan: I’m not sure these wrappers add anything, but YMMV.

shared_infra/sqs_redrive/run.py
```diff
@@ -0,0 +1,47 @@
...
+
+def main(sqs_client):
+    source_queue_url = os.environ["SQS_SOURCE_URL"]
+    target_queue_url = os.environ["SQS_TARGET_URL"]
+    while True:
+        response = sqs_client.receive_message(
+            QueueUrl=source_queue_url
+        )
+
+        if "Messages" not in response:
```
alexwlchan: Easier-to-Ask-For-Permission (EAFP) is generally considered more Pythonic than Look-Before-You-Leap. So it’s slightly more idiomatic to write:

```python
try:
    for msg in response['Messages']:
        ...
except KeyError:
    break
```

But this approach is fine, and I don’t care strongly about which one you use.

shared_infra/sqs_redrive/run.py
```diff
@@ -0,0 +1,47 @@
...
+def delete_message(sqs_client, msg, queue_url):
+    sqs_client.delete_message(
+        QueueUrl=queue_url,
+        ReceiptHandle=msg['ReceiptHandle']
+    )
+
+
+def main(sqs_client):
+    source_queue_url = os.environ["SQS_SOURCE_URL"]
+    target_queue_url = os.environ["SQS_TARGET_URL"]
```
alexwlchan: We should log these variables.

PR #1159, opened 2017-11-08
Remove the lambda_utils project

alexwlchan:
It’s superseded by https://github.com/wellcometrust/aws_utils.

I’m not removing old uses of it yet – the old versions will continue to work, and we can switch to the newer package when we’re next working in that part of the code.


alexwlchan:
WTF Travis?

PR #1160, opened 2017-11-08
Sierra API class

kenoir:

Adding a Python lib for interacting with the Sierra API.

### Who is this change for?

Folk who want to read from the Sierra API>

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


kenoir:
For https://github.com/wellcometrust/platform/issues/1054

kenoir:
Do we reckon this should be a separate package entirely, or should we keep it here for the time being ... i'm leaning towards keep it here until it matures a bit.

shared.Makefile
```diff
@@ -11,6 +11,8 @@ export INFRA_BUCKET = platform-infra
 
 ROOT = $(shell git rev-parse --show-toplevel)
 
+CURRENT_DIR = $(shell pwd)
```
alexwlchan: You could also use `$(CURDIR)`, I think.

PR #1163, opened 2017-11-09
Expose contributor codes in the creator field

alexwlchan:

Make sure we give enough information to credit people for their images in the "Creator" field.

- [ ] Deployed new versions


alexwlchan:
Resolves #1162.

PR #1164, opened 2017-11-09
sierra_objects_to_s3 task

kenoir:

Add a task capable of placing objects from the Sierra API (given a date range) into S3 one file per object.

### Who is this change for?

Folk who want Sierra items in the catalogue api.

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


kenoir:
This PR is missing the terraform to actually create these tasks. Will add in follow-up PR.

sierra_adapter/sierra_objects_to_s3/src/test_sierra_objects_to_s3.py
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    with recorder['betamax'].use_cassette('sierra_api'):
+        sierra_objects_to_s3.main(args, mock_s3_client, recorder['session'])
+
+        response = mock_s3_client.list_objects(
+            Bucket=test_bucket_name
+        )
+
+        assert len(response['Contents']) == 29
```
kenoir: Not sure yet how to go about testing this more thoroughly. We could check the path for each is as expected, but as to the contents of these objects i'd rather leave that to the _other_ test here.

PR #1176, opened 2017-11-14
drain_ecs_container_instance

kenoir:

The API docs are not clear but for some messages on ASG lifecycle hooks we don't have all the fields: http://docs.aws.amazon.com/autoscaling/latest/userguide/lifecycle-hooks.html#lifecycle-hook-considerations 

Our lambda expects all the fields so errors, stop quietly instead.

### Who is this change for?

💻 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.

### DIFF
--- a/shared_infra/drain_ecs_container_instance/src/drain_ecs_container_instance.py
+++ b/shared_infra/drain_ecs_container_instance/src/drain_ecs_container_instance.py
@@ -65,9 +65,9 @@ def drain_ecs_container_instance(asg_client, ec2_client, ecs_client, sns_client,
     message_data = json.loads(message)
 
     # Check this is an interesting message
-    if not 'AutoScalingGroupName' in message_data:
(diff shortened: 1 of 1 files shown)


shared_infra/drain_ecs_container_instance/src/drain_ecs_container_instance.py
```diff
@@ -35,21 +36,25 @@ def continue_lifecycle_action(
...
+
+    print(f'response = {response!r}')
 
 
 def get_ec2_tags(ec2_client, ec2_instance_id):
-    ec2_instance_info = ec2_client.describe_instances(InstanceIds=[
-        ec2_instance_id,
-    ])
+    ec2_instance_info = ec2_client.describe_instances(
+        InstanceIds=[
```
alexwlchan: nit: Pull this onto a single line?

PR #1199, opened 2017-11-23
Use wellcome/scalafmt container

kenoir:

Simplify this repo and share dockerfiles in an institution wide repo.

### Who is this change for?

💻 

- [ ] Deployed new versions


alexwlchan:
Restarted the build, as I don't think the container had been built the first time.

PR #1203, opened 2017-11-23
Move Scala projects next to their infra

kenoir:

A simpler more understandable repo structure.

### Who is this change for?

💻 

- [ ] Deployed new versions


Makefile
```diff
@@ -49,18 +49,23 @@ sbt-test-transformer: .docker/sbt_test
 		--file=builds/sbt_image_builder.Dockerfile
 
 sbt-build-api: .docker/sbt_image_builder
+	TARGET=catatlogue_api/api \
```
alexwlchan: I’d be shocked if this is working like you expect, "catatlogue" vs. "catalogue".

PR #1216, opened 2017-11-28
Move Terraform services to read their config from the local repo

alexwlchan:

This moves our application config out of the terraform-modules repo, and stores it alongside the Terraform in this repo.

See also: https://github.com/wellcometrust/terraform/pull/14

### Who is this change for?

Primarily the Platform team, so we don't need to cut a new release of terraform-modules every time we want to make a config change.

- [x] Run `terraform apply`. (Although this should be a no-op.)


alexwlchan:
Note: this needs to be merged *after* the patch in terraform-modules, so I can update the refs to point to the new release.

Issue #1193, opened 2017-11-22
Sierra object to Dynamo


kenoir:
Service for asking sierra for a given date range and updating in dynamo.

Should work for items and Bibs. 

Reads from job queue posted to by lambda.

Issue #1194, opened 2017-11-22
Deploy Sierra object to dynamo services for Items & Bibs


kenoir:
Instances of the Sierra object to dynamo services for both Items & Bibs.

Depends https://github.com/wellcometrust/platform/issues/1193

alexwlchan:
It’s deployed! We still need to write a bit of the code, but all the Terraform, ECR repo, Travis publishing etc. are done.

PR #1217, opened 2017-11-28
Add Terraform config for the Sierra-to-Dynamo app

alexwlchan:

Add the Terraform pieces, so we can actually deploy this application!

TODO:

- [x] Create the DynamoDB table
- [x] Plumb through the flags, config variables, etc.
- [x] Set up IAM permissions on everything

Related: #1194. Closes: #1193.

- [x] Run `terraform apply`.


alexwlchan:
The output from this stack is the stream ARNs of the two new DynamoDB tables, which can be plugged into dynamo_to_sns to get an SNS topic of notifications for the merger, but I’ll leave that for a separate patch. This is big enough!

PR #1220, opened 2017-11-29
Make SQSWorker deal with bad messages in a more graceful way

alicefuzier:
Make SQSReader and SQSWorker deal gracefully with bad sqs messages, without triggering TryBackoff failures if we keep receiving them
### Who is this change for?
Devs


common/src/main/scala/uk/ac/wellcome/sqs/SQSReader.scala
```diff
@@ -67,15 +67,18 @@ class SQSReader @Inject()(sqsClient: AmazonSQS, sqsConfig: SQSConfig)
           info(s"Processing message ${message.getMessageId}")
           process(message)
         })
+        .flatMap(_ => deleteMessage(message))
         .recover {
+          case e :SQSReaderGracefulException =>
```
alexwlchan: nit: spacing

PR #1215, opened 2017-11-27
[WIP] Adds sierra_bib_merger project

kenoir:
### What is this PR trying to achieve?

Merging Sierra objects from the sierra_to_dynamo app (as provided via a DynamoDB Event Stream -> Lambda -> SNS).

In order that these objects can then be introduced to the ingestion pipeline.

This PR adds a simple skeleton for that project.

### Who is this change for?

📖 😍 

### Have the following been considered/are they needed?

- [ ] Deployed new versions
- [ ] Run `terraform apply`.

PR #1228, opened 2017-11-30
Move SierraRecord into the common lib

alexwlchan:
Make this model available to the bib/item merger. Spun out as a separate patch from #1215.


alexwlchan:
Since this is a mechanical change, I’ll merge as soon as tests pass.

PR #1231, opened 2017-11-30
Use default nginx in worker services

kenoir:

Use the default nginx container from: https://github.com/wellcometrust/dockerfiles/tree/master/nginx

### Who is this change for?

💻 

- [ ] Run `terraform apply`.


catalogue_pipeline/terraform/lambdas.tf
```diff
@@ -26,7 +26,7 @@ module "trigger_reindexer_lambda" {
 }
 
 module "dynamo_to_sns" {
-  source = "git::https://github.com/wellcometrust/platform.git//shared_infra/dynamo_to_sns?ref=dynamo_to_sns-redux"
+  source = "git::https://github.com/wellcometrust/platform.git//shared_infra/dynamo_to_sns"
```
alexwlchan: 😱

kenoir: yeah

PR #1232, opened 2017-11-30
Add the Sierra bib merger

alexwlchan:

Supersedes #1215.

- [ ] Deployed new versions

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/models/MergedSierraObject.scala
+++ b/common/src/main/scala/uk/ac/wellcome/models/MergedSierraObject.scala
@@ -1,3 +1,8 @@
 package uk.ac.wellcome.models
 
-case class MergedSierraObject(id: String)
+case class MergedSierraObject(id: String, bibData: Option[String] = None)
+
+object MergedSierraObject {
+  def apply(id: String, bibData: String): MergedSierraObject =
+    MergedSierraObject(id = id, bibData = Some(bibData))
+}
--- a/common/src/main/scala/uk/ac/wellcome/models/MergedSierraObject.scala
+++ b/common/src/main/scala/uk/ac/wellcome/models/MergedSierraObject.scala
@@ -1,0 +1,3 @@
+package uk.ac.wellcome.models
+
+case class MergedSierraObject(id: String)
--- a/common/src/main/scala/uk/ac/wellcome/models/MergedSierraObject.scala
+++ b/common/src/main/scala/uk/ac/wellcome/models/MergedSierraObject.scala
@@ -1,8 +1,0 @@
-package uk.ac.wellcome.models
-
-case class MergedSierraObject(id: String, bibData: Option[String] = None)
-
-object MergedSierraObject {
-  def apply(id: String, bibData: String): MergedSierraObject =
-    MergedSierraObject(id = id, bibData = Some(bibData))
-}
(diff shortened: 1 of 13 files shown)


common/src/main/scala/uk/ac/wellcome/models/MergedSierraObject.scala
```diff
@@ -0,0 +1,8 @@
+package uk.ac.wellcome.models
+
+case class MergedSierraObject(id: String, bibData: Option[String] = None)
```
kenoir: After talking to @alexwlchan we might end up with something like:
 
```scala
case class MergedSierraRecord(bib: SierraRecord, items: List[SierraRecord], version: Int)
```

sierra_adapter/sierra_bib_merger/src/main/scala/uk/ac/wellcome/platform/sierra_bib_merger/modules/SierraBibMergerModule.scala
```diff
@@ -1,10 +1,35 @@
...
+import uk.ac.wellcome.models.aws.DynamoConfig
+import uk.ac.wellcome.platform.sierra_bib_merger.services.{
+  SierraBibMergerUpdaterService,
+  SierraBibMergerWorkerService
+}
+import uk.ac.wellcome.utils.TryBackoff
+
+object SierraBibMergerModule extends TwitterModule with TryBackoff {
+
+  override lazy val continuous: Boolean = false
```
alicefuzier: Why is this false?

sierra_adapter/sierra_bib_merger/src/main/scala/uk/ac/wellcome/platform/sierra_bib_merger/services/SierraBibMergerUpdaterService.scala
```diff
@@ -0,0 +1,60 @@
...
+          )
+          .put(record)
+        val x = Scanamo.exec(dynamoDBClient)(ops) match {
+          case Right(_) =>
+            logger.info(s"$record saved successfully to DynamoDB")
+          case Left(error) =>
+            logger.warn(s"Failed saving $record to DynamoDB", error)
+        }
+      }
+      case None => ()
```
alicefuzier: You don't need to return `()`. You can just do `case None => `

alexwlchan:
I’m going to rebase against current master, now that #1241 and #1242 are merged.

sierra_adapter/sierra_bib_merger/src/main/scala/uk/ac/wellcome/platform/sierra_bib_merger/modules/SierraBibMergerModule.scala
alexwlchan: This is a TryBackoff parameter – it means that after the first success, we should continue to process messages rather than exit as succeeded.

alicefuzier: probably late, but _i think_ the behaviour is the opposite. If it's false it will stop after the first message

alexwlchan:
```scala
  * Additionally, the operation can run again after it succeeds (reverting
  * to the base wait time), or give up after the first success.  This is
  * controlled by a third attribute:
  *
  *     @param continuous      true if the operation should repeat, false
  *                            if it should return after the first success
```

oops, you're right

Issue #1196, opened 2017-11-22
Sierra Bib merger


kenoir:
Merge Item records into a dynamo table providing the entry point into the ingest pipeline.

Needs to do the consistent read, conditional update shuffle.

Shares a dynamo table with the Sierra Item merger (https://github.com/wellcometrust/platform/issues/1195).

Depends: https://github.com/wellcometrust/platform/issues/1194

alexwlchan:
Closed by #1232.

PR #1233, opened 2017-12-01
Terraform for the Sierra bib merger

alexwlchan:

Write all the Terraform (I think) required for the Sierra bib merger.

Related: #1196.

- [x] Run `terraform apply`.


sierra_adapter/terraform/dynamo.tf
```diff
@@ -0,0 +1,28 @@
...
+  stream_enabled   = true
+  stream_view_type = "NEW_AND_OLD_IMAGES"
+
+  attribute {
+    name = "id"
+    type = "S"
+  }
+
+  attribute {
+    name = "version"
```
kenoir: Do we need the version to be a range key? There'll only ever be one version in the table right?

sierra_adapter/terraform/sierra_merger/variables.tf
```diff
@@ -0,0 +1,34 @@
...
+variable "cluster_id" {}
+
+variable "vpc_id" {}
+
+variable "ecr_repository_url" {}
+variable "release_id" {}
+variable "build_env" { default = "prod" }
+
+variable "nginx_ecr_repository_url" {}
+variable "nginx_release_id" {}
```
kenoir: Do we need these given the new default nginx setup?

sierra_adapter/terraform/sierra_merger/services.tf
```diff
@@ -0,0 +1,52 @@
...
+}
+
+module "sierra_merger_service" {
+  source             = "git::https://github.com/wellcometrust/terraform.git//services?ref=s3-mystery"
+  name               = "sierra_${var.resource_type}_merger"
+  cluster_id         = "${var.cluster_id}"
+  task_role_arn      = "${module.ecs_sierra_merger.task_role_arn}"
+  vpc_id             = "${var.vpc_id}"
+  app_uri            = "${var.ecr_repository_url}:${var.release_id}"
+  nginx_uri          = "${var.nginx_ecr_repository_url}:${var.nginx_release_id}"
```
kenoir: Don't think we need the `nginx_uri` as this is defaulted if we use `v1.3.0` of the services module.

alexwlchan: yep, removing now

sierra_adapter/terraform/sierra_merger/services.tf
```diff
@@ -0,0 +1,52 @@
...
+
+  config_vars = {
+    windows_queue_url = "${module.update_events_queue.id}"
+    metrics_namespace = "sierra_${var.resource_type}_merger"
+    dynamo_table_name = "${var.target_dynamo_table_name}"
+  }
+
+  loadbalancer_cloudwatch_id   = "${var.alb_cloudwatch_id}"
+  server_error_alarm_topic_arn = "${var.alb_server_error_alarm_arn}"
+  client_error_alarm_topic_arn = "${var.alb_client_error_alarm_arn}"
```
kenoir: Needs `https_domain = services.wellcomecollection.ac.uk` as this get passed to nginx when it starts.

See: https://github.com/wellcometrust/terraform/blob/master/services/variables.tf#L161

sierra_adapter/terraform/main.tf
```diff
@@ -64,14 +63,40 @@ module "sierra_to_dynamo_bibs" {
...
   sierra_oauth_secret = "${var.sierra_oauth_secret}"
 
   vpc_id = "${local.services_vpc_id}"
 
   account_id = "${data.aws_caller_identity.current.account_id}"
+
+  lambda_error_alarm_arn = "${local.lambda_error_alarm_arn}"
+}
+
+module "sierra_bib_merger" {
```
kenoir: Like this structure 👍

Issue #1236, opened 2017-12-01
Split SierraRecord into two classes: SierraBibRecord and SierraItemRecord


alexwlchan:
So we can have type-safety on methods on MergedSierraRecord.

PR #1237, opened 2017-12-01
Add a MergedSierraRecord class, and the ability to merge SierraBibRecords

alexwlchan:
This model implements the logic for combining a bib record from the MergedSierraData table, and a new bib record that comes from the Sierra API (via our Sierra To Dynamo app).

Item merging is **explicitly out of scope**.

This is for #1196, spun out of #1232, and makes a start on #1236.

### DIFF
--- a/common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
+++ b/common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
@@ -55,7 +55,7 @@ class MergedSierraRecordTest extends FunSpec with Matchers {
       id = "777",
       bibData = Some(
         sierraBibRecord(
-          id = "888",
+          id = "777",
           title = "Shiny McNewRecordz 2.0",
           modifiedDate = "2009-09-09T09:09:09Z"
         )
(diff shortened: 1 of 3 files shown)


common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
```diff
@@ -0,0 +1,130 @@
...
+        )
+      )
+    )
+
+    val result = mergedSierraRecord.mergeBibRecord(record)
+    result shouldBe None
+  }
+
+  it("should update bibData when merging bib records with newer data") {
+
```
alicefuzier: Empty test?

common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  it("should only merge bib records with matching ids") {
+    val record = sierraBibRecord(id = "333")
+    val originalRecord = MergedSierraRecord(id = "444")
+
+    val caught = intercept[RuntimeException] {
+      originalRecord.mergeBibRecord(record)
+    }
+    caught.getMessage shouldEqual "Non-matching bib ids 333 != 444"
```
alicefuzier: it it can throw exceptions, would it be morth making it return a Try or an Either?

PR #1244, opened 2017-12-04
Sierra transformer

alicefuzier:
Make MergedSierraRecord transform into a work (with just title)

### DIFF
--- a/catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/SierraTransformerFeatureTest.scala
+++ b/catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/SierraTransformerFeatureTest.scala
@@ -1,0 +1,40 @@
+package uk.ac.wellcome.transformer
+
+import java.time.Instant
+
+import org.scalatest.{FunSpec, Matchers}
+import uk.ac.wellcome.transformer.utils.{TransformableSQSMessageUtils, TransformerFeatureTest}
+
+class SierraTransformerFeatureTest
+  extends FunSpec
+    with Matchers
+    with TransformerFeatureTest
+    with TransformableSQSMessageUtils {
+
+  val queueUrl: String = createQueueAndReturnUrl("test_sierra_transformer")
+  override val flags: Map[String, String] = Map(
+    "transformer.source" -> "SierraData",
+    "aws.region" -> "eu-west-1",
+    "aws.sqs.queue.url" -> queueUrl,
+    "aws.sqs.waitTime" -> "1",
+    "aws.sns.topic.arn" -> idMinterTopicArn,
+    "aws.metrics.namespace" -> "sierra-transformer"
+  )
+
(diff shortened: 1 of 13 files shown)


common/src/main/scala/uk/ac/wellcome/models/MergedSierraRecord.scala
```diff
@@ -38,8 +42,23 @@ case class MergedSierraRecord(
       None
     }
   }
+
+  override def transform: Try[Work] =
+    JsonUtil.fromJson[SierraBibData](bibData.get.data).map { sierraBibData =>
```
alexwlchan: Should we have a branch that recovers in the case that `bibData.get` is None? And then test for that?

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/SierraTransformerFeatureTest.scala
```diff
@@ -0,0 +1,53 @@
...
+    "aws.sqs.queue.url" -> queueUrl,
+    "aws.sqs.waitTime" -> "1",
+    "aws.sns.topic.arn" -> idMinterTopicArn,
+    "aws.metrics.namespace" -> "sierra-transformer"
+  )
+
+  it("should transform sierra records, and publish them to the given topic") {
+
+    val id = "001"
+    val title = "A pot of possums"
```
alexwlchan: 😄

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/SierraTransformerFeatureTest.scala
```diff
@@ -0,0 +1,53 @@
...
+
+    eventually {
+      val snsMessages = listMessagesReceivedFromSNS()
+      snsMessages should have size 1
+
+      val expectedWork = Work(
+        title = title,
+        identifiers =
+          List(SourceIdentifier(IdentifierSchemes.sierraSystemNumber, id)))
+      snsMessages.head.message shouldBe JsonUtil.toJson(expectedWork).get
```
alexwlchan: minor: I’d rather unpack the JSON message as a structured object and compare those, than cast to JSON and trust that we get equal serialisations.

(I feel like we had a method for comparing JSON objects somewhere?)

common/src/main/scala/uk/ac/wellcome/models/MergedSierraRecord.scala
```diff
@@ -23,36 +27,56 @@ case class MergedSierraRecord(
...
           version = this.version + 1
         ))
     } else {
       None
     }
   }
+
+  override def transform: Try[Option[Work]] =
+    maybeBibData
+      .map { foo =>
```
alexwlchan: Can we have a better variable name than `foo`?

PR #1245, opened 2017-12-05
Use default service healthcheck paths

alexwlchan:
Requires https://github.com/wellcometrust/terraform-modules/pull/18.


alexwlchan:
Merging because although the check-format task is failing, it's a Scala issue, not Terraform.

PR #1255, opened 2017-12-07
Have Travis push a commit for auto-format if errors are detected

alexwlchan:

Streamline our build process – if a pull request is opened with a linting error that can be fixed with `make format`, Travis will create a commit and push it directly to the branch.

### Who is this change for?

Devs who want a streamlined build process. No more "Run 'make format'" commits for us to make.

Also, the way Travis short-circuits PR builds means we should reduce build contention and improve CI times – the autoformat commit will cancel the previous build, and save us a few build VMs.


.travis/run_autoformat.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python
+# -*- encoding: utf-8
+"""
+Wrapper script for doing auto-formatting in Travis.
```
kenoir: Stick it in a container, base it off: https://github.com/wellcometrust/dockerfiles/tree/master/build_tooling

alexwlchan: Yeah, this is just a POC/WIP.

alexwlchan:
I’ve also done a bunch of work to improve the Travis logging; adding `flush=True` means that a print appears when it occurs, rather than being buffered to stdout and emitted at the end of the script.

PR #1259, opened 2017-12-07
Use latest version of service module

kenoir:

Keep platform up to date with the latest service definition.

### Who is this change for?

💻 

- [ ] Deployed new versions


catalogue_pipeline/terraform/service_reindexer.tf
```diff
@@ -1,10 +1,11 @@
 module "miro_reindexer" {
-  source             = "git::https://github.com/wellcometrust/terraform.git//services?ref=v1.3.1"
-  name               = "miro_reindexer"
-  cluster_id         = "${aws_ecs_cluster.services.id}"
-  task_role_arn      = "${module.ecs_miro_reindexer_iam.task_role_arn}"
```
alexwlchan: Care to delete the old ecs_miro_reindexer_iam module while you’re here?

PR #1262, opened 2017-12-07
Remove calm from the identifiers table

alicefuzier:


catalogue_pipeline/id_minter/src/main/resources/db/migration/V5__remove calm.sql
```diff
@@ -0,0 +1,12 @@
+-- Now we're ingesting Sierra records, we're actually using Sierra IDs in
+-- the ID minter.
+
+USE ${database};
+
+-- Mysql automatically sets the index name to the name of the first
```
kenoir: 👍

Issue #1234, opened 2017-12-01
Run the sierra_adapter in a dedicated cluster (maybe?)


alexwlchan:
Right now it’s using the services cluster, which is a weird mixture of abstractions.

We might want to create a dedicated cluster, but we'd need to wrap the current autoscaling code first.

alexwlchan:
Closed in #1263.

PR #1263, opened 2017-12-07
Use new cluster module for Sierra

kenoir:

Use the new cluster module to split the sierra adapter out.

### Who is this change for?

💻 

- [x] Deployed new versions


alexwlchan:
Resolves #1234.

sierra_adapter/terraform/locals.tf
```diff
@@ -1,15 +1,17 @@
 locals {
-  services_alb_listener_http_arn  = "${data.terraform_remote_state.catalogue_pipeline.services_alb_listener_http_arn}"
-  services_alb_listener_https_arn = "${data.terraform_remote_state.catalogue_pipeline.services_alb_listener_https_arn}"
+  alb_cloudwatch_id      = "${module.sierra_adapter_cluster.alb_cloudwatch_id}"
```
alexwlchan: I’d shortcut locals at this point; the few characters saved isn’t that much clearer now, and it’s easier to see where it comes from within the same stack.

alexwlchan:
It’s relying on at least one bugfix in the modules repo, so I’d like to get that reviewed + merged tomorrow. I’ll work on this first, and then we can get it closed off.

PR #1264, opened 2017-12-08
[WIP] Make sure the ID minter handles Sierra records and nulls correctly

alexwlchan:
Last night, when I ran the ID minter for the first time on Sierra records, it kept throwing an exception:

```
scalikejdbc.TooManyRowsException: null
```

Digging into the logs, I found the following SQL query:

```
Executing SQL query = 'select i.CanonicalID as C1_on_i, i.ontologyType as o_on_i, i.MiroID as M_on_i, i.CalmAltRefNo as C2_on_i, i.SierraSystemNumber as S_on_i from identifiers.identifiers i where i.ontologyType = work and ( i.SierraSystemNumber = b3010658 or i.SierraSystemNumber is null)'
```

which returns every row in the database. Oops!

Let’s not do that.

---

I've written a test case which should reproduce the error (if this build doesn't fail, that's suspicious!), and then two of us should pair on fixing it, as SQL will undoubtedly be fiddly.


alexwlchan:
Closing in favour of #1266.

PR #1265, opened 2017-12-08
Pass SourcedWork from transformer

kenoir:

Pass SourcedWork from transformer in order that the id_minter can know about the source system. 

### Who is this change for?

Folk who want to identify items by their source system/

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


kenoir:
We don't think this will work when items (or other things) need identifiers.

PR #1267, opened 2017-12-11
Use more prebuilt Docker images for CI tasks

alexwlchan:

Continuing the build work from Friday; I’m pushing more code into the dockerfiles repo and calling it from here.

See https://github.com/wellcometrust/dockerfiles/pull/7 for where these are all going.

### Who is this change for?

🚤


alexwlchan:
That shaves another 4 minutes off the total build time (hard to tell cumulative, as I rearranged the build tasks between the last point of comparison). 🎉

builds/sbt_image_builder.Dockerfile
```diff
@@ -1,15 +1,9 @@
-FROM pvansia/scala-sbt:0.13.13
+FROM wellcome/sbt_wrapper
```
alexwlchan: This change means we'll see a small saving in builds as well, as we don't need to install docker twice in a run.

alexwlchan:
I’d already pushed local builds so the tests on this branch would pass; merging.

Issue #1195, opened 2017-11-22
Sierra Item merger


kenoir:
Merge Item records into a dynamo table providing the entry point into the ingest pipeline.

Needs to do the consistent read, conditional update shuffle.

Shares a dynamo table with the Sierra Bib merger (https://github.com/wellcometrust/platform/issues/1196).

Depends: https://github.com/wellcometrust/platform/issues/1194

alexwlchan:
I’m working on the model code – creating a SierraItemRecord class, and a method mergeItemRecord which adds it to a MergedSierraRecord.

alexwlchan:
mmmm but how do we fan out the updates

more work and thinking required

PR #1270, opened 2017-12-11
Add model code for putting items on a MergedSierraRecord

alexwlchan:

This creates a new class `SierraItemRecord`, which is currently identical to `SierraBibRecord` but is a different type for the purposes of type safety.

It adds a way to store item records on a `MergedSierraRecord`, and adds a new method `mergeItemRecord()`, which adds information from a new item to the merged record.

This patch only covers the model changes; exercising the model outside our tests is beyond the scope of this patch.

Required for #1195.

### Who is this change for?

Anybody who wants more than bib data from Sierra.


alexwlchan:
I think this is ready for review.

PR #1273, opened 2017-12-11
Make the existing bib merger more resource-type agnostic

alexwlchan:
As part of work on #1195, this refactors the bib merger to take a `sierra.resourceType` flag, which can be "bibs" or "items", and it then parses SQS messages based on that flag.

That’s the theory, at least – I haven’t done any of the item merger logic yet, because I want to run this past Robert and Alice first. Is this a sensible design?

I did this because I’d built the boilerplate for a separate item merger, and I realised how much code is shared between the two applications.


sierra_adapter/sierra_merger/src/main/scala/uk/ac/wellcome/platform/sierra_merger/services/SierraMergerWorkerService.scala
```diff
@@ -13,17 +14,25 @@ import scala.concurrent.Future
...
+  @Flag("sierra.resourceType") resourceType: String
 ) extends SQSWorker(reader, system, metrics)
     with Logging {
 
   override def processMessage(message: SQSMessage): Future[Unit] =
-    JsonUtil.fromJson[SierraBibRecord](message.body) match {
-      case Success(record) => sierraBibMergerUpdaterService.update(record)
-      case Failure(e) => Future(logger.warn(s"Failed processing $message", e))
+    resourceType match {
+      case "bibs" => JsonUtil.fromJson[SierraBibRecord](message.body) match {
```
alexwlchan: So I’d add a similar case for "items" just below…

sierra_adapter/sierra_merger/src/main/scala/uk/ac/wellcome/platform/sierra_merger/services/SierraMergerUpdaterService.scala
```diff
@@ -29,7 +29,7 @@ class SierraBibMergerUpdaterService @Inject()(
       )
       .put(record)
 
-  def update(bibRecord: SierraBibRecord): Future[Unit] = Future {
+  def updateBib(bibRecord: SierraBibRecord): Future[Unit] = Future {
```
alexwlchan: …which calls into a new updateItem() method here.

sierra_adapter/sierra_merger/src/main/scala/uk/ac/wellcome/platform/sierra_merger/services/SierraMergerWorkerService.scala
alicefuzier: I don't think I like this: I'd go for something similar to the transformer where we read the flag once in `TransformableParserModule` and based on that flag we instantiate the right `TransformableParser` and from then onwards use that. At runtime I don't think we should be using this flag

PR #1276, opened 2017-12-11
Add envsubst to config loading

kenoir:

Allows environment variables in containers to be substituted for app config.

This is a step towards providing app config via env variable only, removing the S3 step.

### Who is this change for?

💻 

- [ ] Deployed new versions

### DIFF
--- a/builds/scala_service/run.sh
+++ b/builds/scala_service/run.sh
@@ -8,7 +8,9 @@ echo "Running project $PROJECT"
 
 echo "Fetching config from AWS..."
 aws s3 ls s3://$INFRA_BUCKET/$CONFIG_KEY
-aws s3 cp s3://$INFRA_BUCKET/$CONFIG_KEY /opt/docker/conf/application.ini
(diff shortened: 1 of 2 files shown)


builds/scala_service/run.sh
```diff
@@ -8,7 +8,9 @@ echo "Running project $PROJECT"
 
 echo "Fetching config from AWS..."
 aws s3 ls s3://$INFRA_BUCKET/$CONFIG_KEY
-aws s3 cp s3://$INFRA_BUCKET/$CONFIG_KEY /opt/docker/conf/application.ini
+aws s3 cp s3://$INFRA_BUCKET/$CONFIG_KEY /opt/docker/conf/application.ini.template
+
+envsubst < /opt/docker/conf/application.ini.template > /opt/docker/conf/application.ini
```
alexwlchan: Could we get a comment explaining exactly what this is meant to do? Double redirects like this are always fiddly.

kenoir:
Closing, as this is getting implemented on a per project basis.

PR #1277, opened 2017-12-11
Make the existing sierra_to_dynamo be bibs-only

alexwlchan:
We'll have two different apps: one for bibs, one for items, in the same way we have for the mergers – because the items ingestor also needs to be recording all the bibs it was attached to. I’ll add the new project in a separate PR.


alexwlchan:
Resolves #1236.

alexwlchan:
Replaced by #1280.

PR #1278, opened 2017-12-11
Proof of concept containers with environment variables

alicefuzier:
Don't merge. this is just an idea about how we might avoid pulling config files from s3 and use environment variables instead.

To build the container do 
```bash
sbt "project blergh" "stage" && docker build catalogue_api/blergh -t blerg
```

and to run it 
```bash
docker run -p 8888:8888 -e API_NAME=hwqejg  blergh
```


kenoir:
Closing as we've got a real one: https://github.com/wellcometrust/platform/pull/1279

PR #1279, opened 2017-12-11
Build id_minter with env config

kenoir:

Discard S3 config in favour of env variable config only to make running application containers more flexible.

This change uses a new base container for the finatra service and a project local `Dockerfile`.  We should eventually be able to get rid of the sbt image builder container locally and simpligy ouf build and configuration process. 

Depends: https://github.com/wellcometrust/dockerfiles/pull/8

### Who is this change for?

💻 

- [ ] Deployed new versions

### DIFF
--- a/catalogue_pipeline/id_minter/Dockerfile
+++ b/catalogue_pipeline/id_minter/Dockerfile
@@ -1,0 +1,7 @@
+FROM finatra_service_base:latest
+
+ADD target/universal/stage /opt/docker
+
+ENV project=id_minter
+
+CMD ["/run.sh"]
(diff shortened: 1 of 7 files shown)


catalogue_pipeline/id_minter/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM finatra_service_base:latest
```
kenoir: Not yet an automated build.

functions.Makefile
```diff
@@ -110,3 +110,16 @@ define sbt_test
...
 		"project $(1)" ";dockerComposeUp;test;dockerComposeStop"
 endef
+
+
+# Build an sbt project.
+#
+# Args:
+#   $1 - Name of the project.
+#
+define sbt_build
```
kenoir: We don't need a special build image!

PR #1285, opened 2017-12-12
If only .tf files have changed, we don't need to run non-format tasks

alexwlchan:

Make our builds go _even faster_.

Plus starting to add some tests for our Travis tooling. For now these are just for running locally, but eventually it would be nice to have Travis automatically test these scripts as well.

### Who is this change for?

🚤


alexwlchan:
This was originally written with #1286 in mind, although the Makefile changes mean we still need to test the adapter stack.

Issue #1004, opened 2017-09-29
Ingest items from Sierra


jtweed:

kenoir:
Closing as superseded by Q2 goals.

PR #1288, opened 2017-12-12
Skeleton for the sierra_items_to_dynamo app

alexwlchan:
For #1004.


alexwlchan:
I’ve also broken up the Travis tasks for the sierra_adapter stack, as that one job was getting a bit unwieldy.

PR #1293, opened 2017-12-13
Update MergedSierraRecord to take account of linked bib IDs

alexwlchan:
Required model logic for the Sierra item merger: now a SierraItemRecord can record which bib IDs it’s currently linked or unlinked to, we should take account of that when merging a record.

Related: #1195.

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/models/MergedSierraRecord.scala
+++ b/common/src/main/scala/uk/ac/wellcome/models/MergedSierraRecord.scala
@@ -50,7 +50,7 @@ case class MergedSierraRecord(
   /** Given a new item record, construct the new merged row that we should
     * insert into the merged database.
     *
-    * Returns None if there's nothing to do.
+    * Returns the merged record.
     */
   def mergeItemRecord(record: SierraItemRecord): MergedSierraRecord = {
 
--- a/common/src/main/scala/uk/ac/wellcome/models/MergedSierraRecord.scala
+++ b/common/src/main/scala/uk/ac/wellcome/models/MergedSierraRecord.scala
@@ -75,6 +75,8 @@ case class MergedSierraRecord(
     }
   }
 
+  def unlinkItemRecord(itemRecord: SierraItemRecord): MergedSierraRecord = ???
+
   override def transform: Try[Option[Work]] =
     maybeBibData
       .map { bibData =>
--- a/common/src/main/scala/uk/ac/wellcome/models/MergedSierraRecord.scala
+++ b/common/src/main/scala/uk/ac/wellcome/models/MergedSierraRecord.scala
@@ -52,7 +52,7 @@ case class MergedSierraRecord(
     *
     * Returns None if there's nothing to do.
     */
-  def mergeItemRecord(record: SierraItemRecord): Option[MergedSierraRecord] = {
+  def mergeItemRecord(record: SierraItemRecord): MergedSierraRecord = {
 
     // We can decide whether to insert the new data in two steps:
     //
@@ -69,9 +69,9 @@ case class MergedSierraRecord(
 
     if (isNewerData) {
       val itemData = this.itemData + (record.id -> record)
-      Some(this.copy(itemData = itemData))
+      this.copy(itemData = itemData)
     } else {
-      None
+      this
     }
   }
 
--- a/common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
+++ b/common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
@@ -205,15 +205,20 @@ class MergedSierraRecordTest extends FunSpec with Matchers {
           bibIds = List("i111"),
           unlinkedBibIds = List(unlinkedBibId)
         )
-        val mergedSierraRecord = MergedSierraRecord(id = unlinkedBibId,itemData = Map(record.id -> record))
-        mergedSierraRecord.unlinkItemRecord(record) shouldBe mergedSierraRecord.copy(itemData = Map.empty)
+        val mergedSierraRecord = MergedSierraRecord(id = unlinkedBibId,
+                                                    itemData =
+                                                      Map(record.id -> record))
+        mergedSierraRecord.unlinkItemRecord(record) shouldBe mergedSierraRecord
+          .copy(itemData = Map.empty)
       }
 
-      ignore("should return None when merging an unlinked record which is already absent") {
+      ignore(
+        "should return None when merging an unlinked record which is already absent") {
         true shouldBe false
       }
 
-      ignore("should return None when merging an unlinked record which has linked more recently") {
+      ignore(
+        "should return None when merging an unlinked record which has linked more recently") {
         true shouldBe false
       }
     }
--- a/common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
+++ b/common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
@@ -112,85 +112,111 @@ class MergedSierraRecordTest extends FunSpec with Matchers {
   }
 
   describe("merging with a SierraItemRecord") {
-    it("should add itemData when there isn't already an item") {
-      val record = sierraItemRecord(
-        id = "i888",
-        title = "Illustrious imps are ingenious",
-        modifiedDate = "2008-08-08T08:08:08Z",
-        bibIds = List("b888")
-      )
+    describe("when the merged record is linked to the item") {
+      it("should add the item if it doesn't exist already") {
+        val record = sierraItemRecord(
+          id = "i888",
+          title = "Illustrious imps are ingenious",
+          modifiedDate = "2008-08-08T08:08:08Z",
+          bibIds = List("b888")
+        )
 
-      val mergedSierraRecord = MergedSierraRecord(id = "b888")
-      val result = mergedSierraRecord.mergeItemRecord(record).get
+        val mergedSierraRecord = MergedSierraRecord(id = "b888")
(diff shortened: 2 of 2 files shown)


common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
```diff
@@ -7,238 +7,276 @@ import uk.ac.wellcome.utils.JsonUtil
...
+      caught.getMessage shouldEqual "Non-matching bib ids 333 != 444"
     }
-    caught.getMessage shouldEqual "Non-matching bib ids 333 != 444"
-  }
 
-  it("should be at version 1 when first created") {
-    val record = MergedSierraRecord(id = "555")
-    record.version shouldEqual 1
-  }
+    it("should always increment the version when mergeBibRecord is called") {
```
kenoir: I think we can omit the "shoulds" from tests safely, without the should it still makes sense and makes the descriptions less repetitive.

kenoir: e.g.

`it("should always increment the version when mergeBibRecord is called")`

vs. 

`it("always increments the version when mergeBibRecord is called")`

common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
```diff
@@ -7,238 +7,276 @@ import uk.ac.wellcome.utils.JsonUtil
...
+            modifiedDate = "2009-09-09T09:09:09Z"
+          )
         )
       )
-    )
 
-    val result = mergedSierraRecord.mergeBibRecord(record)
-    result shouldBe None
-  }
+      val result = mergedSierraRecord.mergeBibRecord(record)
```
kenoir: Do we want to follow the same pattern as in: https://github.com/wellcometrust/platform/pull/1295? I.e. this returns an unchanged record?

common/src/main/scala/uk/ac/wellcome/models/MergedSierraRecord.scala
```diff
@@ -52,7 +52,7 @@ case class MergedSierraRecord(
     *
     * Returns None if there's nothing to do.
```
alexwlchan: This comment is out-of-date.

common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
```diff
@@ -7,236 +7,289 @@ import uk.ac.wellcome.utils.JsonUtil
...
 
-  it("should allow creation from a SierraBibRecord and a version") {
-    val bibRecord = sierraBibRecord(id = "202")
-    val version = 10
-    val mergedRecord = MergedSierraRecord(
-      bibRecord = bibRecord,
-      version = version
-    )
-    mergedRecord.version shouldEqual version
+    it("should be at version 1 when first created") {
```
alexwlchan: This doesn’t match the test.

alexwlchan: This is also duplicated on L57–60, we probably don’t need both?

common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
```diff
@@ -7,236 +7,289 @@ import uk.ac.wellcome.utils.JsonUtil
...
-    newRecord.version shouldEqual 10
-  }
+    it("should never increment the version when mergeBibRecord is called") {
+      val record = sierraBibRecord(id = "666")
+      val originalRecord = MergedSierraRecord(id = "666", version = 10)
+      val newRecord = originalRecord.mergeBibRecord(record)
+      newRecord.version shouldEqual 10
+    }
+
+    it("should return None when merging bib records with stale data") {
```
alexwlchan: Description is out-of-date.

common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
```diff
@@ -7,236 +7,289 @@ import uk.ac.wellcome.utils.JsonUtil
...
-    )
-
-    val mergedSierraRecord = MergedSierraRecord(id = "b121")
-    val result1 = mergedSierraRecord.mergeItemRecord(record1).get
-    val result2 = result1.mergeItemRecord(record2).get
-
-    result1.itemData.get(record1.id).get shouldBe record1
-    result2.itemData.get(record2.id).get shouldBe record2
-  }
+        result.itemData(record.id) shouldBe newerRecord
```
alexwlchan: My Scala isn’t as good as I thought – I thought something like this would return an Option:

```scala
val myMap = Map(1 -> 1, 2 -> 2)
myMap(2)  // Option[Int]?
```

common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
```diff
@@ -7,236 +7,289 @@ import uk.ac.wellcome.utils.JsonUtil
...
+      }
+
+      it("should return itself when merging item records with stale data") {
+        val record = sierraItemRecord(
+          id = "i111",
+          title = "Only otters occupy the orange oval",
+          modifiedDate = "2001-01-01T01:01:01Z",
+          bibIds = List("i111")
```
alexwlchan: bibId is wrong.

common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
kenoir: test description is at odds with test.

common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
```diff
@@ -7,236 +7,289 @@ import uk.ac.wellcome.utils.JsonUtil
...
+        )
+        val mergedSierraRecord = MergedSierraRecord(id = unlinkedBibId,
+                                                    itemData =
+                                                      Map(record.id -> record))
+        mergedSierraRecord.unlinkItemRecord(record) shouldBe mergedSierraRecord
+          .copy(itemData = Map.empty)
+      }
+
+      ignore(
+        "should return None when merging an unlinked record which is already absent") {
```
alexwlchan: Although we’re ignoring these tests for now, do you want to update the descriptions (i.e. not *should return None*).

common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
```diff
@@ -7,236 +7,289 @@ import uk.ac.wellcome.utils.JsonUtil
...
+        true shouldBe false
+      }
+
+      ignore(
+        "should return None when merging an unlinked record which has linked more recently") {
+        true shouldBe false
+      }
+    }
+
+    describe("when the merged record is unrelated to the item") {
```
alexwlchan: I think we’d agreed to break this block up:

* a test for `mergeSierraItemRecord` that checks the ID is in bibIds
* a test for `unlinkSierraItemRecord` that checks the ID is in unlinkedBibIds

Could we get stubs added to this patch?

common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
kenoir: this test description is incorrect

common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
kenoir: 👍

common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
kenoir: `val x: Option[Int] = myMap.get(2)` Hava a look at https://docs.scala-lang.org/overviews/collections/maps.html

common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
alicefuzier: I don't think I understand these two extra tests...

common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
kenoir: interesting ... so this means that we can merge items to bibs without the correct bibids. I think this probably related to the unlinking problem, so i'm not going to add any more tests here.

kenoir: ah there are already tests for this `ignored` later on.

common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
kenoir: 👍

alexwlchan: Something like:

```scala
itemRecord = SierraItemRecord(..., bibIds = [123])
mergedRecord = MergedSierraRecord(id = [456])

assert mergedRecord.mergeSierraItemRecord(itemRecord) throws RuntimeException
```

and similar for `unlinkSierraItemRecord`

alexwlchan:
> Fix the tests and it’s good to merge.

???, self

PR #1295, opened 2017-12-13
Refactor the bib merging code in the sierra_bib_merger

alexwlchan:
* Return a complete record, not an Option[MergedSierraRecord]
* Push version handling code down into the merger

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/models/MergedSierraRecord.scala
+++ b/common/src/main/scala/uk/ac/wellcome/models/MergedSierraRecord.scala
@@ -23,7 +23,7 @@ case class MergedSierraRecord(
   id: String,
   maybeBibData: Option[SierraBibRecord] = None,
   itemData: Map[String, SierraItemRecord] = Map[String, SierraItemRecord](),
-  version: Int = 1
+  version: Int = 0
 ) extends Transformable {
 
   /** Given a new bib record, construct the new merged row that we should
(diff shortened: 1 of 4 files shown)


common/src/main/scala/uk/ac/wellcome/models/MergedSierraRecord.scala
```diff
@@ -23,15 +23,15 @@ case class MergedSierraRecord(
...
   maybeBibData: Option[SierraBibRecord] = None,
   itemData: Map[String, SierraItemRecord] = Map[String, SierraItemRecord](),
-  version: Int = 1
+  version: Int = 0
 ) extends Transformable {
 
   /** Given a new bib record, construct the new merged row that we should
     * insert into the merged database.
     *
     * Returns None if there's nothing to do.
```
kenoir: Not right now?

kenoir: I think conceptually this should have been "construct a list of merge operations to perform" ... I think that was the intention of returning an option in any case.

sierra_adapter/sierra_bib_merger/src/main/scala/uk/ac/wellcome/platform/sierra_bib_merger/services/SierraBibMergerUpdaterService.scala
```diff
@@ -36,48 +36,42 @@ class SierraBibMergerUpdaterService @Inject()(
...
         case Right(record) => {
           logger.info(s"Found $record, attempting merge.")
 
-          record
-            .mergeBibRecord(bibRecord)
-            .toRight(
-              new RuntimeException("Unable to merge record!")
-            )
+          val newRecord = record.mergeBibRecord(bibRecord)
+          if (record != newRecord) {
```
kenoir: I like it better the other way.

PR #1296, opened 2017-12-13
Create a common lib for the Sierra adapter

### LINKED ISSUE #1294: Create a sierra_common library

alexwlchan:

Create a common library for the Sierra adapter applications, but one which is distinct from our global common lib.

Resolves #1294.

### Who is this change for?

* Reduce the size of the common lib
* Reduce the test/build churn while we’re working on Sierra stuff
* Better isolation of code


.travis/tooling.py
```diff
@@ -94,6 +94,22 @@ def affects_tests(path, task):
             print("+++ %s is definitely part of the %s tests" % (path, task))
             return True
 
+    # A number of Sierra-related tasks share code/Makefiles in the
+    # sierra_adapter directory.  If we're definitely in a project which
+    # has nothing to do with Sierra, we can ignore changes in this dir.
+    sierra_free_tasks = (
```
kenoir: 👍  but is this because some things are in the wrong place?

sierra_adapter/sierra_items_to_dynamo/src/main/scala/uk/ac/wellcome/platform/sierra_items_to_dynamo/services/WindowExtractor.scala
```diff
@@ -1,44 +0,0 @@
-package uk.ac.wellcome.platform.sierra_items_to_dynamo.services
```
kenoir: Where did this go?

alexwlchan: Nowhere, we had two copies of it and now we have just one.

.travis/tooling.py
alexwlchan: Not sure I understand the question?

PR #1297, opened 2017-12-13
Sierra item merger (no unlinking logic)

alicefuzier:
Make sierra_item_merger actually do stuff
### Who is this change for?
👽


sierra_adapter/sierra_item_merger/src/main/scala/uk/ac/wellcome/platform/sierra_item_merger/dynamo/MergedSierraRecordDao.scala
```diff
@@ -0,0 +1,55 @@
...
+import com.gu.scanamo.syntax._
+import com.twitter.inject.Logging
+import uk.ac.wellcome.models.MergedSierraRecord
+import uk.ac.wellcome.dynamo._
+import uk.ac.wellcome.models.aws.DynamoConfig
+
+import scala.concurrent.Future
+import uk.ac.wellcome.utils.GlobalExecutionContext.context
+
+class MergedSierraRecordDao @Inject()(dynamoDbClient: AmazonDynamoDB,
```
alexwlchan: Very nice! Maybe we could put this straight into the sierra_adapter lib? I don’t see anything here which is item-merger specific, and it would allow us to tidy up the bib merger as well.

sierra_adapter/sierra_item_merger/src/main/scala/uk/ac/wellcome/platform/sierra_item_merger/modules/SierraDynamoConfigModule.scala
```diff
@@ -0,0 +1,15 @@
+package uk.ac.wellcome.platform.sierra_item_merger.modules
+
+import com.google.inject.Provides
+import uk.ac.wellcome.finatra.modules.DynamoConfigModule
+import uk.ac.wellcome.models.aws.DynamoConfig
+import javax.inject.Singleton
+
+object SierraDynamoConfigModule extends DynamoConfigModule {
```
alexwlchan: This could probably be copied to the common lib as well.

sierra_adapter/sierra_item_merger/src/main/scala/uk/ac/wellcome/platform/sierra_item_merger/modules/SierraItemMergerModule.scala
```diff
@@ -0,0 +1,29 @@
+package uk.ac.wellcome.platform.sierra_item_merger.modules
+
+import akka.actor.ActorSystem
+import com.twitter.inject.{Injector, TwitterModule}
+import uk.ac.wellcome.platform.sierra_item_merger.services.SierraItemMergerWorkerService
+import uk.ac.wellcome.utils.TryBackoff
+
+object SierraItemMergerModule extends TwitterModule with TryBackoff {
+
+  override lazy val continuous: Boolean = false
```
alexwlchan: I can’t remember where @alicefuzier and I got to on this – is non-continuous definitely the right behaviour?

sierra_adapter/sierra_item_merger/src/main/scala/uk/ac/wellcome/platform/sierra_item_merger/services/SierraItemMergerUpdaterService.scala
```diff
@@ -0,0 +1,37 @@
...
+              mergedSierraRecordDao.updateRecord(mergedRecord)
+            else Future.successful(())
+          case None =>
+            mergedSierraRecordDao.updateRecord(
+              MergedSierraRecord(bibId,
+                                 itemData = Map(itemRecord.id -> itemRecord)))
+        }
+    }
+    Future.sequence(updateFutures).map(_ => ())
+  }
```
alexwlchan: 👏 

That’s really nice.

sierra_adapter/sierra_item_merger/src/main/scala/uk/ac/wellcome/platform/sierra_item_merger/services/SierraItemMergerWorkerService.scala
```diff
@@ -0,0 +1,29 @@
...
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+import scala.util.{Failure, Success}
+
+class SierraItemMergerWorkerService @Inject()(
+  reader: SQSReader,
+  system: ActorSystem,
+  metrics: MetricsSender,
+  sierraBibMergerUpdaterService: SierraItemMergerUpdaterService
```
alexwlchan: `sierraBibMergerUpdaterService`?

sierra_adapter/sierra_item_merger/src/test/scala/uk/ac/wellcome/platform/sierra_item_merger/SierraItemMergerFeatureTest.scala
```diff
@@ -0,0 +1,109 @@
...
+import org.scalatest.{FunSpec, Matchers}
+import uk.ac.wellcome.models.aws.SQSMessage
+import uk.ac.wellcome.models.{MergedSierraRecord, SierraItemRecord}
+import uk.ac.wellcome.platform.sierra_item_merger.locals.DynamoDBLocal
+import uk.ac.wellcome.platform.sierra_item_merger.utils.SierraItemMergerTestUtil
+import uk.ac.wellcome.test.utils.{AmazonCloudWatchFlag, SQSLocal}
+import uk.ac.wellcome.utils.JsonUtil
+import uk.ac.wellcome.dynamo._
+
+class SierraItemMergerFeatureTest
```
alexwlchan: The tests as written are fine, but I think we need more of them. At a minimum, corresponding tests to the ones we have on the BibMergerFeatureTest:

* should update an item in DynamoDB if a newer version is sent to SQS
* should not update an item in DynamoDB if an older version is sent to SQS
* should put an item from SQS into DynamoDB if the ID exists but no itemData

I wonder, as well – do we want a test around interactions between bib/item data?

i.e. the item merger should not change bibData when it updates the items, and vice versa for bibs

sierra_adapter/sierra_item_merger/src/test/scala/uk/ac/wellcome/platform/sierra_item_merger/dynamo/MergedSierraRecordDaoTest.scala
```diff
@@ -0,0 +1,191 @@
...
+        itemData = Map(
+          "i111" -> SierraItemRecord(id = "i111",
+                                     data = "something",
+                                     modifiedDate = "2001-01-01T01:01:30Z",
+                                     bibIds = List("b1111"))),
+        version = 0
+      )
+
+      whenReady(mergedSierraRecordDao.updateRecord(mergedSierraRecord)) { _ =>
+        Scanamo
```
alexwlchan: You may want to nick `dynamoQueryEqualsValue` from the bib merger feature tests, or put that into a common lib – would allow you to clean up this code somewhat.

sierra_adapter/sierra_item_merger/src/test/scala/uk/ac/wellcome/platform/sierra_item_merger/dynamo/MergedSierraRecordDaoTest.scala
```diff
@@ -0,0 +1,191 @@
...
+          .get[MergedSierraRecord](dynamoDbClient)(tableName)('id -> id)
+          .get shouldBe Right(
+          mergedSierraRecord.copy(version = 1)
+        )
+      }
+    }
+
+    it(
+      "should update a merged sierra record if it already exists and has a higher version") {
```
alexwlchan: This description is a bit ambiguous; it’s not clear what has the higher version. How about:

> it updates an existing record if the update has a higher version

sierra_adapter/sierra_item_merger/src/test/scala/uk/ac/wellcome/platform/sierra_item_merger/dynamo/MergedSierraRecordDaoTest.scala
```diff
@@ -0,0 +1,191 @@
...
+          Scanamo
+            .get[MergedSierraRecord](dynamoDbClient)(tableName)('id -> id)
+            .get shouldBe Right(
+            newerMergedSierraRecord.copy(version = 3)
+          )
+      }
+    }
+
+    it(
+      "should update a merged sierra record if it already exists and has a the same version") {
```
alexwlchan: Is this the correct behaviour?

I would think that if you try to put an MSR with the same version as an existing database row, your update gets rejected – to prevent two instances trying to PUT competing updates. Isn’t this why we have the version field?

But I’m 👍 on the test case, checking this boundary is definitely important.

sierra_adapter/sierra_item_merger/src/test/scala/uk/ac/wellcome/platform/sierra_item_merger/dynamo/MergedSierraRecordDaoTest.scala
```diff
@@ -0,0 +1,191 @@
...
+        Scanamo
+          .get[MergedSierraRecord](dynamoDbClient)(tableName)('id -> id)
+          .get shouldBe Right(
+          mergedSierraRecord.copy(version = 2)
+        )
+      }
+    }
+
+    it(
+      "should not update a merged sierra record if it already exists and has a lower version") {
```
alexwlchan: Ditto ambiguity above; how about

> it should not update an existing record if the update has a lower version

sierra_adapter/sierra_item_merger/src/main/scala/uk/ac/wellcome/platform/sierra_item_merger/services/SierraItemMergerUpdaterService.scala
```diff
@@ -0,0 +1,37 @@
...
+  mergedSierraRecordDao: MergedSierraRecordDao,
+  metrics: MetricsSender
+) extends Logging {
+
+  def update(itemRecord: SierraItemRecord): Future[Unit] = {
+
+    val updateFutures = itemRecord.bibIds.map { bibId =>
+      mergedSierraRecordDao
+        .getRecord(bibId)
+        .flatMap {
```
kenoir: Like this 👍

sierra_adapter/sierra_item_merger/src/test/scala/uk/ac/wellcome/platform/sierra_item_merger/locals/DynamoDBLocal.scala
```diff
@@ -0,0 +1,63 @@
...
+trait DynamoDBLocal extends BeforeAndAfterEach with DynamoDBLocalClients {
+  this: Suite =>
+
+  val tableName = "sierraItemsTable"
+
+  deleteTable()
+
+  private val table: CreateTableResult = createTable()
+
+  val streamArn: String =
```
alexwlchan: I don’t think this is every used, and then you can get rid of the `table` variable as well.

sierra_adapter/sierra_item_merger/src/test/scala/uk/ac/wellcome/platform/sierra_item_merger/locals/DynamoDBLocal.scala
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  private def clearTable(): List[DeleteItemResult] =
+    Scanamo.scan[MergedSierraRecord](dynamoDbClient)(tableName).map {
+      case Right(item) =>
+        dynamoDbClient.deleteItem(
+          tableName,
+          Map("id" -> new AttributeValue(item.id))
+        )
+      case a =>
```
alexwlchan: nit: can we use a better variable name than `a`?

sierra_adapter/sierra_item_merger/src/test/scala/uk/ac/wellcome/platform/sierra_item_merger/locals/DynamoDBLocal.scala
```diff
@@ -0,0 +1,63 @@
+package uk.ac.wellcome.platform.sierra_item_merger.locals
```
alexwlchan: This could live in the common lib and be shared with the bib merger, surely? They have the same table structure, yes?

sierra_adapter/sierra_item_merger/src/main/scala/uk/ac/wellcome/platform/sierra_item_merger/services/SierraItemMergerUpdaterService.scala
```diff
@@ -0,0 +1,37 @@
...
+            if (mergedRecord != record)
+              mergedSierraRecordDao.updateRecord(mergedRecord)
+            else Future.successful(())
+          case None =>
+            mergedSierraRecordDao.updateRecord(
+              MergedSierraRecord(bibId,
+                                 itemData = Map(itemRecord.id -> itemRecord)))
+        }
+    }
+    Future.sequence(updateFutures).map(_ => ())
```
kenoir: Because this update is not transactional, does it matter if it fails half way through?

alicefuzier: I think it should be fine. If it fails halfway through, the message will be resent, but the updates that succeeded won't be reactioned

sierra_adapter/sierra_item_merger/src/main/scala/uk/ac/wellcome/platform/sierra_item_merger/dynamo/MergedSierraRecordDao.scala
kenoir: Agreed. I think the work to use this in the bib merger is another PR though. Will move this into the sierra lib here though.

alicefuzier: I think we should definitely use it in the bib merger, but I would move it there as part of a refactoring patch on the bib merger, not on this patch

sierra_adapter/sierra_item_merger/src/main/scala/uk/ac/wellcome/platform/sierra_item_merger/modules/SierraDynamoConfigModule.scala
alicefuzier: Same as above: I agree but we have a lot of deducplication of code to do both on the mergers and on the sierra to dynamo ones. I think we should do that separately

sierra_adapter/sierra_item_merger/src/main/scala/uk/ac/wellcome/platform/sierra_item_merger/dynamo/MergedSierraRecordDao.scala
alexwlchan: I thought it might like to land there immediately because it was introduced in this patch, rather than being an edit to existing code, but I’m not massively fussed.

sierra_adapter/sierra_item_merger/src/main/scala/uk/ac/wellcome/platform/sierra_item_merger/modules/SierraItemMergerModule.scala
alicefuzier: No, it's the wrong behaviour. Copy paste error...

sierra_adapter/sierra_item_merger/src/test/scala/uk/ac/wellcome/platform/sierra_item_merger/dynamo/MergedSierraRecordDaoTest.scala
kenoir: I think this is definitely wrong, if the versions are the same it means that another process has updated the record while we were working.

sierra_adapter/sierra_item_merger/src/main/scala/uk/ac/wellcome/platform/sierra_item_merger/services/SierraItemMergerWorkerService.scala
alicefuzier: oops

sierra_adapter/sierra_item_merger/src/test/scala/uk/ac/wellcome/platform/sierra_item_merger/SierraItemMergerFeatureTest.scala
alicefuzier: I don't think there is much value in replicating test at the feature level. Those cases, unless I misunderstand their scope, are covered by the `SierraItemMergerUpdaterServiceTests`.

Feature tests should not cover corner cases but just assert that thing are wired in correclty and that everything works together.

sierra_adapter/sierra_item_merger/src/test/scala/uk/ac/wellcome/platform/sierra_item_merger/dynamo/MergedSierraRecordDaoTest.scala
alicefuzier: I've pushed the logic for incrementing the version in the dao, so that means this case should be correct. The name of the test is probably ambiguous though

sierra_adapter/sierra_item_merger/src/main/scala/uk/ac/wellcome/platform/sierra_item_merger/services/SierraItemMergerWorkerService.scala
```diff
@@ -21,9 +26,24 @@ class SierraItemMergerWorkerService @Inject()(
...
+
   override def processMessage(message: SQSMessage): Future[Unit] =
-    JsonUtil.fromJson[SierraItemRecord](message.body) match {
-      case Success(record) => sierraBibMergerUpdaterService.update(record)
-      case Failure(e) => Future(logger.warn(s"Failed processing $message", e))
+    decode[SierraItemRecord](message.body) match {
+      case Right(record) => sierraBibMergerUpdaterService.update(record)
+      case Left(e) => Future{
+        logger.warn(s"Failed processing $message", e)
+        throw e
```
kenoir: This will return a failed future will it not? The behaviour is different from using JsonUtil as before I think.

alicefuzier: Yes, it is, but I think the new behaviour is the correct one? I guess we don't want the message deleted from the queue if parsing it fails, do we? We should make sure we wrap it in a SqsReaderGracefulException (or whatever) though, so that we don't fail the TryBackoff

sierra_adapter/sierra_item_merger/src/test/scala/uk/ac/wellcome/platform/sierra_item_merger/SierraItemMergerFeatureTest.scala
alexwlchan: Got It – I’d missed the service tests when I did my original review.

kenoir: I'm inclined to agree that we don't need to cover these at multiple levels.

sierra_adapter/sierra_item_merger/src/test/scala/uk/ac/wellcome/platform/sierra_item_merger/dynamo/MergedSierraRecordDaoTest.scala
kenoir: I've been thoroughly persuaded this is the correct behaviour!

alexwlchan: Right, so I think I’ve convinced myself the test is correct, it’s the description that’s confusing. But I’m no longer convinced this test does anything interestingly different from the last one.

PR #1298, opened 2017-12-13
Use latest version of service module without S3 config

kenoir:

Simplify service modules!

### Who is this change for?

💻 

- [x] Deployed new versions


alexwlchan:
@kenoir I think this is ready for final review.

kenoir:
@alexwlchan all good.

PR #1309, opened 2017-12-15
Graceful try-backoff message parsing fails

kenoir:

The sierra mergers should not either trigger try backoff or eat incorrectly parsed message.

### Who is this change for?

📖 📈 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


kenoir:
Depends https://github.com/wellcometrust/platform/pull/1308

Issue #1281, opened 2017-12-12
Deleting a record from dynamo tables crashes our apps


alicefuzier:
Deleting a record from the SierraData_bibs and SierraData_Merged tables causes TryBackoffTerminalFailures downstream (sierra_bib_merger and sierra_transformer)

alicefuzier:
Still TODO

- [ ] Fix transformer
- [ ] Fix ingestor (in progress in #1318)

kenoir:
Closing as fixed.

PR #1314, opened 2017-12-18
Use circe in the bib merger

alicefuzier:
To solve #1281 for the bib merger (Still needs some work for the rest of our apps cause I think this affects anything that reads from SQS)
Basically a message like the following in our SQS queues causes NullPointerException in our applications
```json
{
  "Type" : "Notification",
  "MessageId" : "f589a60f-2fa1-5440-b416-f4a2f510f7e6",
  "TopicArn" : "arn:aws:sns:eu-west-1:123456789:sierra_bib_merger_events",
  "Subject" : "default-subject",
  "Message" : "null",
  "Timestamp" : "2017-12-15T14:13:05.242Z"
}
```
This is because, when we try to do `JsonUtil.fromJson[SierraBibRecord](message.body)` using Jackson, Jackson succeeds and returs a `null` `SierraItemRecord`. As we don't check for null downstream this eventually fails with NullPointerException.

I don't think we should be checking for null. This is a problem that comes from the fact that Jackson is a Java native library where sending `null` around is common, but in Scala this should be avoided. Circe, being a Scala native library, does not send nulls around, so let's use that

### DIFF
--- a/sierra_adapter/common/src/main/scala/uk/ac/wellcome/sierra_adapter/circe/package.scala
+++ b/sierra_adapter/common/src/main/scala/uk/ac/wellcome/sierra_adapter/circe/package.scala
@@ -1,0 +1,21 @@
+package uk.ac.wellcome.sierra_adapter
+
+import java.time.Instant
+
+import io.circe.{Decoder, HCursor}
+import cats.syntax.either._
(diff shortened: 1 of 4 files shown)


kenoir:
Some failing tests with `object circe is not a member of package`.

PR #1315, opened 2017-12-18
Add unlink logic to MergedSierraRecord

kenoir:

Adds the unlinking logic to MergedSierraRecord. 

### Who is this change for?

⛓ ‼️ 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.

### DIFF
--- a/common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
+++ b/common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
@@ -1,6 +1,7 @@
 package uk.ac.wellcome.models
 
 import java.time.Instant
+import java.time.Instant._
 
 import org.scalatest.{FunSpec, Matchers}
 import uk.ac.wellcome.utils.JsonUtil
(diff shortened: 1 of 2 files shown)


common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
```diff
@@ -195,42 +196,149 @@ class MergedSierraRecordTest extends FunSpec with Matchers {
...
           modifiedDate = "2001-01-01T01:01:01Z",
-          bibIds = List("i111"),
-          unlinkedBibIds = List(unlinkedBibId)
+          bibIds = List(bibId),
+          unlinkedBibIds = List()
+        )
+
+        val unlinkedItemRecord = record.copy(
+          bibIds = Nil,
+          modifiedDate = record.modifiedDate.plusSeconds(1),
```
alexwlchan: 👌

common/src/test/scala/uk/ac/wellcome/models/MergedSierraRecordTest.scala
```diff
@@ -195,42 +196,149 @@ class MergedSierraRecordTest extends FunSpec with Matchers {
...
+          title = "Only otters occupy the orange oval",
+          modifiedDate = "2001-01-01T01:01:01Z",
+          bibIds = List(bibId),
+          unlinkedBibIds = List()
+        )
+
+        val previouslyUnlinkedRecord = sierraItemRecord(
+          id = "i222",
+          title = "Only otters occupy the orange oval",
+          modifiedDate = "2001-01-01T01:01:01Z",
```
alexwlchan: Want to adjust the modified date so it’s definitely later than the previous record?

kenoir: I think all the dates can be cleaned up here - though i'm not sure this one should be changed, the test really should be "ignore an unrelated unlinking event".

PR #1319, opened 2017-12-18
Adds unlinking to SierraItemMergerUpdater

kenoir:

Resolves https://github.com/wellcometrust/platform/issues/1195

### Who is this change for?

🐼 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.

### DIFF
--- a/sierra_adapter/sierra_item_merger/src/main/scala/uk/ac/wellcome/platform/sierra_item_merger/services/SierraItemMergerUpdaterService.scala
+++ b/sierra_adapter/sierra_item_merger/src/main/scala/uk/ac/wellcome/platform/sierra_item_merger/services/SierraItemMergerUpdaterService.scala
@@ -44,13 +44,14 @@ class SierraItemMergerUpdaterService @Inject()(
             else Future.successful(())
           // In the case we cannot find the bib record
           // assume we're too early and put the message back
-          case None => Future.failed(SQSReaderGracefulException(
-            new RuntimeException("Missing Bib record to unlink: $unlinkedBibId.")
-          ))
+          case None =>
+            Future.failed(
+              SQSReaderGracefulException(
+                new RuntimeException(
+                  "Missing Bib record to unlink: $unlinkedBibId.")
+              ))
         }
     }
-
-
 
     Future.sequence(mergeUpdateFutures ++ unlinkUpdateFutures).map(_ => ())
   }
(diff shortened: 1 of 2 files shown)


sierra_adapter/sierra_item_merger/src/main/scala/uk/ac/wellcome/platform/sierra_item_merger/services/SierraItemMergerUpdaterService.scala
```diff
@@ -32,7 +33,26 @@ class SierraItemMergerUpdaterService @Inject()(
...
+            val mergedRecord = record.unlinkItemRecord(itemRecord)
+            if (mergedRecord != record)
+              mergedSierraRecordDao.updateRecord(mergedRecord)
+            else Future.successful(())
+          // In the case we cannot find the bib record
+          // assume we're too early and put the message back
+          case None => Future.failed(SQSReaderGracefulException(
+            new RuntimeException("Missing Bib record to unlink: $unlinkedBibId.")
+          ))
+        }
```
alicefuzier: I think there might be some way of generalise this block inside the map. The only difference with the block in lines 20-33 is the call to `record.unlinkItemRecord(itemRecord)` instead of `record.mergeItemRecord(itemRecord)`.

Something like 
```scala
private def combineMergedRecordWithItem(bibId: String, itemRecord: SierraItemRecord)(
    f: MergedSierraRecord => SierraItemRecord =>  MergedSierraRecord) = {
    mergedSierraRecordDao
      .getRecord(bibId)
      .flatMap {
        case Some(record) =>
          val mergedRecord = f(record)(itemRecord)
          if (mergedRecord != record)
            mergedSierraRecordDao.updateRecord(mergedRecord)
          else Future.successful(())
        case None =>
          mergedSierraRecordDao.updateRecord(
            MergedSierraRecord(bibId,
                               itemData = Map(itemRecord.id -> itemRecord)))
      }
  }
```

that then gets called as 
```scala
val mergeUpdateFutures = itemRecord.bibIds.map { bibId =>
      combineMergedRecordWithItem(bibId, itemRecord)(
        _.mergeItemRecord)
    }
val unlinkUpdateFutures = itemRecord.bibIds.map { bibId =>
      combineMergedRecordWithItem(bibId, itemRecord)(
        _. unlinkItemRecord)
    }
```

I'm not sure if it's more readable but it definitely reduces code duplication

kenoir: 👍  will clean up a bit in another PR

PR #1320, opened 2017-12-19
Consolidate DynamoConfig for sierra-x-to-dynamo services

kenoir:

Use the same dynamo config object for all sierra services, as per the catalogue pipeline.

### Who is this change for?

⚙️ 🙉 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


kenoir:
@alexwlchan I think this actually keeps the flags the same ... just checking.

PR #1322, opened 2017-12-19
Move transform function out of Transformable

alicefuzier:
Move transformer specific code into the Transformer and make the Transfomable classes pure representation of domain objects with no business logic


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/modules/TransformerModule.scala
```diff
@@ -0,0 +1,38 @@
...
+      case "CalmData" => new CalmTransformableTransformer
+      case "SierraData" => new SierraTransformableTransformer
+      case tableName =>
+        throw new RuntimeException(s"$tableName is not a recognised source")
+    }
+  }
+
+  @Singleton
+  @Provides
+  def providesTransformableParser(): TransformableParser[Transformable] = {
```
alicefuzier: Annoing duplicated logic but it going to go away once we do #1337

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/TransformableTransformer.scala
```diff
@@ -0,0 +1,17 @@
+package uk.ac.wellcome.transformer.transformers
+
+import uk.ac.wellcome.models.Work
+import uk.ac.wellcome.models.transformable.Transformable
+
+import scala.util.Try
+
+trait TransformableTransformer[+T <: Transformable] {
+  protected[this] def transformForType(t: T): Try[Option[Work]]
```
alicefuzier: I couldn't come up with a better name for this...

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/transformers/CalmTransformableTransformerTest.scala
```diff
@@ -0,0 +1,7 @@
+package uk.ac.wellcome.transformer.transformers
+
+import org.scalatest.FunSpec
+
+class CalmTransformableTransformerTest extends FunSpec {
+
```
alicefuzier: This is a placeholder, we don't have tests for calm transformation in common

PR #1328, opened 2017-12-20
Fix a null pointer bug in the API

alexwlchan:
If you add `?includes=items` to the API, and the works in Elasticsearch don’t have an items field, you get an InternalServerError. This is quite tedious. Make it not be so!

The underlying bug is a null pointer exception, thrown somewhere in Jackson:

```
java.lang.NullPointerException:
  at uk.ac.wellcome.platform.api.models.DisplayWork$.apply(DisplayWork.scala:85)
  at uk.ac.wellcome.platform.api.models.DisplayWork$.jsonToDisplayWork(DisplayWork.scala:104)
  at uk.ac.wellcome.platform.api.models.DisplayWorkTest$$anonfun$1.apply(DisplayItemTest.scala:26)
  at uk.ac.wellcome.platform.api.models.DisplayWorkTest$$anonfun$1.apply(DisplayItemTest.scala:8)
  at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
  at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
  at org.scalatest.Transformer.apply(Transformer.scala:22)
  at org.scalatest.Transformer.apply(Transformer.scala:20)
  at org.scalatest.FunSpecLike$$anon$1.apply(FunSpecLike.scala:454)
  at org.scalatest.TestSuite$class.withFixture(TestSuite.scala:196)
  ...
```

~Since I’m editing the JSON decoding, I‘m swapping it to use Circe at the same time.~

- [x] Reviewed by @jtweed
- [ ] Deployed new versions


alexwlchan:
There’s still a bug in the DisplayItem apply() method which is exposed in ApiWorksTest, but not in DisplayItemTest. Investigating.

PR #1341, opened 2017-12-22
Improve our Loris deployment

alexwlchan:

Several small improvements:

* Use a new Loris patch which should have a mild improvement in connection times
* Bake the config directly into the container (none of it is interpolated anyway), and use the new service module rather than S3 weirdness
* Simplify the image definition and Terraform accordingly, shaving a tidy 50MB off the size of our images

### Who is this change for?

Folks who want a fast and performant Loris.

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


alexwlchan:
(I’m not deploying this today. Even I’m not reckless enough to do a production deploy on the last working day of the year!)

PR #1344, opened 2018-01-02
Send Sierra bib updates to SNS instead of DynamoDB, part 1

alexwlchan:
First part of #1366 – this just moves the sbt/Travis/Make config to point to the new name.


alicefuzier:
Superseeded by #1375

Issue #1243, opened 2017-12-04
Regular ingest items from Sierra


alexwlchan:
- [x] Fetching from Sierra to DynamoDB for items
- [x] Lambdas to feed SQS for each task
- [x] Item merging services to update dynamo

Broken out as a separate ticket from #1005.

Issue #1005, opened 2017-09-29
Regularly ingest bibs from Sierra


jtweed:
- [x] Fetching from Sierra to DynamoDB for bibs
- [x] Lambdas to feed SQS for each task
- [x] Bib merging services to update dynamo

PR #1345, opened 2018-01-02
Add support for different stream types in the DynamoToSNS Lambda

alexwlchan:

Give a way for the user to decide what we emit – full events (the current behaviour, and the default), or a filtered version with the old/new image only.

Our transformers can only process the new image; we have filters in front of this stream, but when a pile of Sierra updates arrive all at once, we seem to be hitting concurrency limits in AWS Lambda. This lets us move the problem back upstream.

Plus a bunch of tests and test refactoring.

The idea is that, in the Sierra pipeline at least, we can cut out the filter Lambda.

Required for: #1005, #1243.

### Who is this change for?

Folks who want a reliable Sierra pipeline.

- [ ] Run `terraform apply`.


alexwlchan:
Eek! This sprawled a bit. I’m going to break this up into smaller, more manageable chunks shortly.

alexwlchan:
I got a bit carried away tidying up the sierra_adapter terraform, trying to give everything more sensible names and levels of encapsulation. I think the new layout is better, but might want to be broken up for an easier review…

alexwlchan:
This should be rebased and reviewed **after** #1348 is merged.

alexwlchan:
Closing, as this has been applied piecemeal in master.

PR #1348, opened 2018-01-03
Make the result type configurable in our Dynamo to SNS Lambda

alexwlchan:
Spun out as a smaller patch from #1345.

> Give a way for the user to decide what we emit – full events (the current behaviour, and the default), or a filtered version with the old/new image only.
>
> Our transformers can only process the new image; we have filters in front of this stream, but when a pile of Sierra updates arrive all at once, we seem to be hitting concurrency limits in AWS Lambda. This lets us to reduce the flood of SNS notifications.
>
> Plus a bunch of tests and test refactoring.
>
> The idea is that, in the Sierra pipeline at least, we can cut out the filter Lambda.

- [ ] Deployed new versions
- [ ] Run `terraform apply`.

### DIFF
--- a/shared_infra/dynamo_to_sns/main.tf
+++ b/shared_infra/dynamo_to_sns/main.tf
@@ -8,7 +8,8 @@ module "lambda_dynamo_to_sns" {
   memory_size = 1024
 
   environment_variables = {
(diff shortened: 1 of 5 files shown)


alexwlchan:
(Merging by fiat because this is non-critical code.)

PR #1353, opened 2018-01-05
Rejig autoscaling in the sierra_to_dynamo apps

alexwlchan:
This patch includes a couple of changes, which could be separate but currently aren’t:

* Don’t autoscale the sierra_to_dynamo apps. The number of records they get from Sierra is potentially unlimited, if we autoscale down we end up interrupting the flow and the window fails.
* Do scale the associated DynamoDB tables, or we end up hitting throughput limits when there are lots of updates.
* Tweak the SQS redrive, so it can be run as an ECS task or as a local script.


kenoir:
The scaling stuff is interesting here, invites a bit of thought to see if we can find a way around this issue. Not here though 👍

PR #1370, opened 2018-01-09
Delete a few unused IAM policy documents

alexwlchan:
These documents are leftover from when the API was our only stack; they're not referenced anywhere in the API Terraform.


alexwlchan:
I’ve run `terraform plan` locally with this patch, and there are no changes.

Issue #1304, opened 2017-12-15
After-action review actions from API flakiness


alexwlchan:
- [x] Fix the credit lines in the API (#1192)
- [x] Have alarms for healthy/unhealthy hosts (#1342)
- [x] Push non-critical alarms into a separate Slack channel (#1305)
- [ ] ~Create summaries of non-critical errors (#1367)~
- [ ] ~Automate Remus/Romulus deployments~ (#1316)
- [x] Terraform notifications in Slack (#1310)
- [x] Terraform warnings on production changes (#1306)

alexwlchan:
Closing because we've done everything we're going to do for now; the two remaining items have been spun out as separate issues.

PR #1371, opened 2018-01-09
Enable healthy/unhealthy hosts alarms for the Catalogue API and Loris

alexwlchan:
Part of #1304.

- [x] Run `terraform apply`.


alexwlchan:
Ah, this isn't quite good enough – the heuristic is “<= 50% of desired count is an alarm”, but we also allow “a minimum of 50% when doing blue/green deploys”.

alexwlchan:
Blocked on https://github.com/wellcometrust/terraform-modules/pull/40

PR #1375, opened 2018-01-09
Send Sierra bib updates to S3 instead of DynamoDB

alexwlchan:
The follow-up to #1344.

Still to do:

* [x] Group updates into “pages”
* [x] Add config options for bibs/items
* [x] Add a "done" marker file
* [ ] Add incomplete window trackers
* [x] Add support for resuming a window
* [x] Set the window prefixes correctly


alexwlchan:
🎉

PR #1377, opened 2018-01-10
Finally swap out jackson for circe

alicefuzier:
Use only circe as json library (except for the api)
### Who is this change for?
🐱


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/source/SierraBibData.scala
```diff
@@ -3,8 +3,8 @@ package uk.ac.wellcome.transformer.source
 case class SierraBibData(
   id: String,
   title: String,
-  deleted: Boolean,
-  suppressed: Boolean,
+  deleted: Boolean = false,
+  suppressed: Boolean = false,
```
alicefuzier: This is because circe will fail decoding json that doesn't contain those fields unless there is a default value specified

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/source/SierraItemData.scala
```diff
@@ -2,7 +2,7 @@ package uk.ac.wellcome.transformer.source
 
 case class SierraItemData(
   id: String,
-  deleted: Boolean,
+  deleted: Boolean = false,
```
alicefuzier: Ditto

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/MiroTransformableTransformer.scala
```diff
@@ -400,35 +400,19 @@ class MiroTransformableTransformer
     * of (key, value) tuples.  Note: we don't use a map because keys aren't
     * guaranteed to be unique.
     */
-  def zipMiroFields(keys: Option[List[String]],
-                    values: Option[List[String]]): List[(String, String)] = {
-    (keys, values) match {
-      case (Some(k), Some(v)) => {
-        if (k.length != v.length) {
+  def zipMiroFields(keys: List[Option[String]],
```
alicefuzier: This is related to the change on `MiroTransformableData`:`image_award` can contain null values ie `[null, "Biomedical award"]`. While Jackson simply ignores `null`, circe needs the list to explicitly be a `List[Option[String]]` to be able to decode `null`.
Nice side effect is that the logic here becomes much simpler.

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/receive/SQSMessageReceiverTest.scala
```diff
@@ -125,7 +124,7 @@ class SQSMessageReceiverTest
     val future = recordReceiver.receiveMessage(failingTransformCalmSqsMessage)
 
     whenReady(future.failed) { x =>
-      x shouldBe a[JsonParseException]
+      x shouldBe a[SQSReaderGracefulException]
```
alicefuzier: I don't think the original behaviour here was correct, so fixing it

common/src/main/scala/uk/ac/wellcome/models/Item.scala
```diff
@@ -7,7 +7,6 @@ case class Item(
   sourceIdentifier: SourceIdentifier,
   identifiers: List[SourceIdentifier] = Nil,
   locations: List[Location] = List(),
-  visible: Boolean = true
-) extends Identifiable {
-  @JsonProperty("type") val ontologyType: String = "Item"
-}
+  visible: Boolean = true,
+  @JsonProperty("type") ontologyType: String = "Item"
```
alicefuzier: Had to pull this into the parameter list cause circe relies on shapeless instead of reflection, which doesn't give visibility on members in the class body. 
I did prefer them being constant, but, meh 😐

kenoir:
retrigger travis

common/src/main/scala/uk/ac/wellcome/models/License.scala
```diff
@@ -18,16 +16,20 @@ sealed trait License {
   @JsonProperty("type") val ontologyType: String = "License"
 }
 
-class LicenseDeserialiser extends JsonDeserializer[License] with Logging {
+object License extends Logging {
+  implicit val licenseEncoder = Encoder.instance[License](license => Json.obj(
```
alicefuzier: Had to do this again cause shapeless doesn't give visibility on things other than case class parameters. Licenses are case objects so they can't have parameters so, meh 😐

common/src/test/scala/uk/ac/wellcome/models/ItemTest.scala
```diff
@@ -23,13 +24,13 @@ class ItemTest extends FunSpec with Matchers with JsonTestUtil {
       |  "locations": [
       |    {
       |      "locationType": "location",
+      |      "url" : null,
+      |      "credit" : null,
```
alicefuzier: circe is configured to output everything, even nulls. We can configure it differently but I prefer this 'cause it means all fields are always present

common/src/test/scala/uk/ac/wellcome/models/ItemTest.scala
```diff
@@ -23,13 +24,13 @@ class ItemTest extends FunSpec with Matchers with JsonTestUtil {
...
       |      "license": {
       |        "licenseType": "${License_CCBY.licenseType}",
       |        "label": "${License_CCBY.label}",
-      |        "url": "${License_CCBY.url}",
-      |        "type": "License"
-      |      },
-      |      "type": "Location"
+      |        "url": "${License_CCBY.url}"
+      |      }
```
alicefuzier: I only changed `Work` and `Item` to have an `ontlogyType` field that can be serialised by circe. That is because they are `Identifiable` and the id_minter needs the `ontologyType` to save in the database along with the `sourceIdentifier`. 
For other types, I don't actually think we need this. It's more relayed to how we display things in the API, so I would be in favour of removing all of these types out of our internal data model. What do you think?

common/src/test/scala/uk/ac/wellcome/models/LicenseTest.scala
```diff
@@ -1,24 +1,18 @@
...
+import uk.ac.wellcome.circe.jsonUtil._
 
 class LicenseTest extends FunSpec with Matchers {
 
   it("should serialise a License as JSON") {
-    val result = JsonUtil.toJson(License_CCBY)
+    val result = jsonUtil.toJson[License](License_CCBY)
     result.isSuccess shouldBe true
-    result.get shouldBe """{"licenseType":"CC-BY","label":"Attribution 4.0 International (CC BY 4.0)","url":"http://creativecommons.org/licenses/by/4.0/","type":"License"}"""
+    result.get shouldBe """{"licenseType":"CC-BY","label":"Attribution 4.0 International (CC BY 4.0)","url":"http://creativecommons.org/licenses/by/4.0/"}"""
```
alicefuzier: same as above

common/src/main/scala/uk/ac/wellcome/models/Identifiable.scala
```diff
@@ -8,4 +8,6 @@ trait Identifiable {
   def id: String = canonicalId.getOrElse(
     throw UnidentifiableException()
   )
+  val sourceIdentifier: SourceIdentifier
+  val ontologyType: String
```
alicefuzier: Id minter needs these to be able to mint ids so enforcing them in the base trait

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/source/MiroTransformableData.scala
```diff
@@ -1,37 +1,34 @@
...
+  @JsonKey("image_tech_file_size") techFileSize: Option[List[String]],
+  @JsonKey("image_use_restrictions") useRestrictions: Option[String],
+  @JsonKey("image_supp_lettering") suppLettering: Option[String],
+  @JsonKey("image_innopac_id") innopacID: Option[String],
+  @JsonKey("image_credit_line") creditLine: Option[String],
+  @JsonKey("image_source_code") sourceCode: Option[String],
+  @JsonKey("image_library_ref_department") libraryRefDepartment: List[Option[String]] = Nil,
+  @JsonKey("image_library_ref_id") libraryRefId: List[Option[String]] = Nil,
+  @JsonKey("image_award") award: List[Option[String]] = Nil,
+  @JsonKey("image_award_date") awardDate: List[Option[String]] = Nil
```
alicefuzier: These have become `List[Option[String]]` instead of `Option[List[String]]`. Check comment on `MiroTransformableTransformer` as to why

catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/models/DisplayItemTest.scala
```diff
@@ -63,7 +64,8 @@ class DisplayItemTest extends FunSpec with Matchers {
   }
 
   it("correctly parses an Item without any locations") {
-    val item = JsonUtil.fromJson[Item]("""
+    val item =
+      ApiJsonUtil.fromJson[Item]("""
```
alexwlchan: It's not especially relevant to this patch, but why is there any JSON decoding happening in this test?

catalogue_pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/IdMinterFeatureTest.scala
```diff
@@ -88,7 +87,7 @@ class IdMinterFeatureTest
     val miroId = "1234"
     val sqsMessage = generateSqsMessage(miroId)
 
-    sqsClient.sendMessage(idMinterQueue, JsonUtil.toJson(sqsMessage).get)
+    sqsClient.sendMessage(idMinterQueue, toJson(sqsMessage).get)
```
alexwlchan: Should we modify the signature of `sendMessage()` to take a queue and an instance of SQSMessage, and have it do the JSON encoding?

(Again, could be a separate patch!)

catalogue_pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/steps/IdEmbedderTests.scala
```diff
@@ -95,8 +95,8 @@ class IdEmbedderTests
         )
     ).thenReturn(Try(throw expectedException))
 
-    val newWorkFuture = idEmbedder.embedId(
-      json = parse(JsonUtil.toJson(originalWork).get).right.get)
+    val newWorkFuture =
+      idEmbedder.embedId(json = parse(toJson(originalWork).get).right.get)
```
alexwlchan: Sending a work to a JSON string, then unpacking it as a Circe Json object, seems moderately odd.

Maybe we need `toJsonString` and `toCirceJson` in the utils file (or similar)? Depends on how frequently it comes up elsewhere.

catalogue_pipeline/ingestor/src/main/scala/uk/ac/wellcome/platform/ingestor/services/WorkIndexer.scala
```diff
@@ -33,12 +31,11 @@ class WorkIndexer @Inject()(
       "ingestor-index-work",
       () => {
         Future
-          .fromTry(Try {
-            decode[Work](document) match {
-              case Right(work) => work
-              case Left(error) => throw SQSReaderGracefulException(error)
+          .fromTry(
+            fromJson[Work](document).recover {
+              case error: Throwable => throw GracefulFailureException(error)
```
alexwlchan: Note to self: this clashes with #1427, and that will need to be rebased atop this later.

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/source/SierraBibData.scala
alexwlchan: When would we not have a value?

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/MiroTransformerFeatureTest.scala
```diff
@@ -1,6 +1,7 @@
 package uk.ac.wellcome.transformer
 
 import org.scalatest.{FunSpec, Matchers}
+import uk.ac.wellcome.utils.JsonUtil._
```
alexwlchan: What do these imports add?

common/src/main/scala/uk/ac/wellcome/models/Identifiable.scala
alexwlchan: (I’d forgotten these weren’t already there!)

common/src/main/scala/uk/ac/wellcome/models/Item.scala
alexwlchan: Yeah, except they aren’t sometimes, so we'd have done this anyway 😕

common/src/test/scala/uk/ac/wellcome/models/ItemTest.scala
alexwlchan: As discussed, I think we do need to pass these types around internally.

sierra_adapter/sierra_item_merger/src/main/scala/uk/ac/wellcome/platform/sierra_item_merger/services/SierraItemMergerWorkerService.scala
```diff
@@ -24,12 +25,12 @@ class SierraItemMergerWorkerService @Inject()(
 
   override def processMessage(message: SQSMessage): Future[Unit] =
     // Using Circe here because Jackson creates nulls for empty lists
-    decode[SierraItemRecord](message.body) match {
-      case Right(record) => sierraItemMergerUpdaterService.update(record)
-      case Left(e) =>
+    JsonUtil.fromJson[SierraItemRecord](message.body) match {
```
alexwlchan: nit: let's use fromJson for consistency.

sierra_adapter/sierra_item_merger/src/main/scala/uk/ac/wellcome/platform/sierra_item_merger/services/SierraItemMergerWorkerService.scala
```diff
@@ -24,12 +25,12 @@ class SierraItemMergerWorkerService @Inject()(
 
   override def processMessage(message: SQSMessage): Future[Unit] =
     // Using Circe here because Jackson creates nulls for empty lists
```
alexwlchan: I think you can lose this comment now!

sierra_adapter/sierra_items_to_dynamo/src/main/scala/uk/ac/wellcome/platform/sierra_items_to_dynamo/services/SierraItemsToDynamoWorkerService.scala
```diff
@@ -21,13 +22,13 @@ class SierraItemsToDynamoWorkerService @Inject()(
 ) extends SQSWorker(reader, system, metrics) {
 
   def processMessage(message: SQSMessage): Future[Unit] =
-    decode[SierraRecord](message.body) match {
-      case Right(record) =>
+    JsonUtil.fromJson[SierraRecord](message.body) match {
```
alexwlchan: nit: ditto above

catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/models/DisplayItemTest.scala
alicefuzier: very good question

PR #1378, opened 2018-01-10
Version Sierra items

kenoir:

Solve a potential issue in the items to dynamo app where a race condition could cause data loss.

### Who is this change for?

🍔 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


kenoir:
Fixes https://github.com/wellcometrust/platform/issues/1333

kenoir:
I did think about drying this out a bit when looking at the `SierraTransformableDao` but decided it wasn't worth it (yet) given that it might be a bit awkward trying to tame Scanamo into being happy with an abstraction.

PR #1400, opened 2018-01-12
Add the S3 demultiplexer

alexwlchan:


kenoir:
I wonder if we can do this with https://docs.aws.amazon.com/sns/latest/dg/message-filtering.html

alexwlchan:
Blocked on https://github.com/wellcometrust/aws_utils/pull/16 being released.

PR #1415, opened 2018-01-15
Move WorksIndex to ingestor

kenoir:

Clarify and make index creation more reliable.

This change starts this process by moving WorksIndex to `ingestor`.

### Who is this change for?

🦀 🔫 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


kenoir:
Closing, as the API is a dependency here (through the tests, as they need the mapping in order to create a valid work).

PR #1417, opened 2018-01-15
Split model and index actions for ES mappings

kenoir:

Clarify and make index creation more reliable.

### Who is this change for?

🦀 🔫 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


common/src/main/scala/uk/ac/wellcome/elasticsearch/WorksIndex.scala
```diff
@@ -1,22 +1,24 @@
...
+                           @Flag("es.index") name: String,
                            @Flag("es.type") itemType: String)
-    extends Logging {
+    extends ElasticSearchIndex
+    with Logging {
+
+  val rootIndexType = "work"
+
+  val httpClient: HttpClient = client
+  val indexName = "works"
```
alexwlchan: Why aren't you using the `-es.index` flag with a default?

common/src/test/scala/uk/ac/wellcome/elasticsearch/ElasticSearchIndexTest.scala
```diff
@@ -0,0 +1,166 @@
...
+    eventually {
+      val hits = elasticClient
+        .execute(search(s"$testIndexName/$testType").matchAllQuery())
+        .map {
+          _.hits.hits
+        }
+        .await
+      hits should have size 1
+
+      JsonUtil.fromJson[TestObject](
```
alexwlchan: We have `assertJsonStringsAreEqual` in `JsonTestUtil`.

Issue #1419, opened 2018-01-15
Transform bibs and items, part 1


alexwlchan:
This is the first set of low-hanging fruit from Silver's notes.

- [x] **wwork:identifiers**. Already working, we need to break it into a separate trait. (#1421)
- [x] **wwork:title**.
- [ ] ~**wwork:genres** and **wwork:subjects**, but only the concept – qualified concepts come later.~
- [x] **wwork:publicationDate** – passing the date straight through. Parsing it into a consistent format comes later. (#1599)
- [x] **wwork:placeOfPublication** – only the non-identified places. (#1648)
- [x] **wwork:publishers** – subfield 260 is already transformed, but not 264. ~Open Q: what if both 260 and 264 are populated?~ (#1429)
- [x] **wwork:lettering** – just use 264 indicator 16; don't worry about whether it’s an archive/image/journal (#1467)
- [x] **wwork:description** (#1444)
- [x] **wwork:extent** (#1632)
- [x] **wwork:physicalDescription** (#1629)
- [ ] **wwork:items** – we need to break it into a separate trait, and add a few extra fields (#1462, #1608, #1609)

Part of #1361.

PR #1429, opened 2018-01-16
Take account of MARC field 264 when populating the "publishers" field

alexwlchan:

Take account of new notes from Silver/Branwen about using field 264 if 260 isn't populated.

Part of #1419.

### Who is this change for?

📖


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraPublishers.scala
```diff
@@ -8,25 +8,38 @@ trait SierraPublishers {
...
-      }
+    val matchingFields = bibData.varFields
+      .filter { _.marcTag.contains("260") }
+
+    val publisherFields = if (!matchingFields.isEmpty) {
+      matchingFields
+    } else {
+      bibData.varFields
+        .filter { _.marcTag.contains("264") }
+    }
```
alexwlchan: I’m not particularly happy with this chunk of code – get 260 fields, or 264 if there aren’t any. Is there a neater way to do this? 😕

PR #1432, opened 2018-01-17
SQSReaderGracefulException -> GracefulFailureException

kenoir:

Decouples the GracefulFailureException from SQSReader so it can be used elsewhere.

### Who is this change for?

🌁 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


alicefuzier:
Merged as part of #1377

PR #1444, opened 2018-01-18
Extract Work description in Sierra transformer

kenoir:

Filling out Works from Sierra. Adds a description.

### Who is this change for?

📖 🐳 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.

### DIFF
--- a/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraDescription.scala
+++ b/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraDescription.scala
@@ -4,7 +4,6 @@ import uk.ac.wellcome.transformer.source.{MarcSubfield, SierraBibData}
 
 trait SierraDescription {
 
-
   def getSubfields(
     bibData: SierraBibData,
     marcTag: String,
@@ -23,7 +22,6 @@ trait SierraDescription {
     })
   }
 
-
   // Populate wwork:description.
   //
   // We use MARC field "520"
@@ -40,15 +38,14 @@ trait SierraDescription {
     getSubfields(bibData, "520", List("a", "b"))
       .foldLeft[List[String]](Nil)((acc, subfields) => {
 
-      (subfields.get("a"), subfields.get("b")) match {
-        case (Some(a), Some(b)) => acc :+ s"${a.content} ${b.content}"
-        case (Some(a), None) => acc :+ a.content
-        case (None, None) => acc
-      }
-    }) match {
+        (subfields.get("a"), subfields.get("b")) match {
+          case (Some(a), Some(b)) => acc :+ s"${a.content} ${b.content}"
+          case (Some(a), None) => acc :+ a.content
+          case (None, None) => acc
+        }
+      }) match {
       case Nil => None
       case list => Some(list.mkString(" "))
     }
   }
 }
-
--- a/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraDescription.scala
+++ b/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraDescription.scala
@@ -4,34 +4,51 @@ import uk.ac.wellcome.transformer.source.{MarcSubfield, SierraBibData}
 
 trait SierraDescription {
 
-  def getSubfield(
+
+  def getSubfields(
     bibData: SierraBibData,
     marcTag: String,
-    marcSubfieldTag: String
-  ): Option[MarcSubfield] = {
+    marcSubfieldTags: List[String]
+  ): List[Map[String, MarcSubfield]] = {
     val matchingFields = bibData.varFields
       .filter {
         _.marcTag.contains(marcTag)
       }
 
-    val matchingSubfields  =
-      matchingFields.flatMap {
-        _.subfields
-      }.flatten
-
-    matchingSubfields.find(_.tag == marcSubfieldTag)
+    matchingFields.map(varField => {
+      varField.subfields
+        .filter(subfield => marcSubfieldTags.contains(subfield.tag))
+        .map(subfield => subfield.tag -> subfield)
+        .toMap
+    })
   }
 
+
+  // Populate wwork:description.
+  //
+  // We use MARC field "520"
+  //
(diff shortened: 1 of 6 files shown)


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraDescription.scala
```diff
@@ -0,0 +1,29 @@
+package uk.ac.wellcome.transformer.transformers.sierra
+
+import uk.ac.wellcome.models.Agent
+import uk.ac.wellcome.transformer.source.{MarcSubfield, SierraBibData}
+
+trait SierraDescription {
+
+  def getDescription(bibData: SierraBibData): Option[String] = {
```
alexwlchan: The big change I’ll want in the final version is a comment, similar to that on getPublisher and getTitle, that explains the _intent_ of the transformation below. A tidied-up version of Silver’s notes would be fine, noting anywhere you’ve made assumptions or deviations from the spec.

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraDescription.scala
```diff
@@ -0,0 +1,29 @@
...
+      }.flatten
+
+    matchingSubfields
+      .filter {
+        _.tag == "a"
+      }
+      .map { subfield =>
+        subfield.content
+      }
+      .headOption
```
alexwlchan: I haven't checked, but it sounds like you're taking the first instance subfield $a.

Some fields/subfields can be repeated; others can't. It would be worth double-checking with Branwen what happens here. Can you have multiple 250 fields? Or subfield $a's on a particular 250?

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraDescription.scala
```diff
@@ -0,0 +1,36 @@
...
+
+    val matchingSubfields =
+      matchingFields.flatMap {
+        _.subfields
+      }.flatten
+
+    matchingSubfields.find(_.tag == marcSubfieldTag)
+  }
+
+  def getDescription(bibData: SierraBibData): Option[String] = {
```
alexwlchan: This needs a comment explaining the intent (see previous comments).

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraDescription.scala
```diff
@@ -0,0 +1,36 @@
...
+    val descriptionField = getSubfield(bibData, "520", "a")
+    val summaryDescriptionField = getSubfield(bibData, "520", "b")
```
alexwlchan: Have you checked that we would only ever have a single instance of these subfields on the entire MARC record?

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraDescription.scala
```diff
@@ -0,0 +1,36 @@
...
+
+    (descriptionField, summaryDescriptionField) match {
+      case (Some(a), Some(b)) => Some(s"${a.content} ${b.content}")
+      case (Some(a), None) => Some(a.content)
+      case (None, None) => None
+    }
```
alexwlchan: Ooh, very neat. 👌

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraDescriptionTest.scala
```diff
@@ -0,0 +1,80 @@
...
+import org.scalatest.{FunSpec, Matchers}
+import uk.ac.wellcome.models.{Agent, IdentifierSchemes, SourceIdentifier}
+import uk.ac.wellcome.test.utils.SierraData
+import uk.ac.wellcome.transformer.source.{
+  MarcSubfield,
+  SierraBibData,
+  VarField
+}
+
+class SierraDescriptionTest extends FunSpec with Matchers with SierraData {
```
alexwlchan: I think you want a third test case – neither subfield a nor b are populated.

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraDescription.scala
kenoir: Yes, apparently there can be!

alexwlchan: err…?

Can be multiple, or can only be one? 😄

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraDescription.scala
alexwlchan: That said – you’re missing a case. What if we find nothing in $a but something in $b?

We don’t expect that to happen, but if it does, we’ll throw a CaseUnmatched error, which I think will be fatal to the transformer’s TryBackoff? We should probably throw a GracefulReaderException here instead, and send the bib to the DLQ.

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraDescription.scala
```diff
@@ -0,0 +1,51 @@
...
+
+  // Populate wwork:description.
+  //
+  // We use MARC field "520"
+  //
+  // The value comes from comes subfield $a concatenated with subfield $b
+  //
+  // Notes:
+  //  - A bib may have multiple 520 records
+  //  - If subfield $b is empty,
```
alexwlchan: "If subfield $b is empty" is an incomplete sentence.

PR #1446, opened 2018-01-18
Make organisation a subclass of Agent

alicefuzier:
Use the type system to enforce model consistency
### Who is this change for?
🦆 🚲


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraPublishers.scala
```diff
@@ -23,7 +22,7 @@ trait SierraPublishers {
   // https://www.loc.gov/marc/bibliographic/bd260.html
   // https://www.loc.gov/marc/bibliographic/bd264.html
   //
-  def getPublishers(bibData: SierraBibData): List[Agent] = {
+  def getPublishers(bibData: SierraBibData): List[Organisation] = {
```
alexwlchan: Double-checking: because Organisation is a subclass of Agent, we could use `List[Agent]` for a list that contained Agents or Organisations?

alicefuzier: yes. Actually `publishers` on `Work` is a `List[Agent]`. I changed the return type of this function cause it seems like in this case we always return `Organisation`

alicefuzier: I don't see `Event` as subclass of `Agent` in @silveroliver model.. But definitely we can for `Agent`, `Person` and `Organisation`

alexwlchan: It’s a funny one, and maybe we need to rethink creators and/or organisations? The relevant bit is here:

https://github.com/wellcometrust/platform/blob/master/ontologies/WIP/sierratransformable#L75-L77

and in fact, Silver was going to look at exactly how that works, because creators has flipped from being an Agent to being a QualifiedConcept, which can in turn specialise to Organisation.

Aren't types fun? 😄

alexwlchan:
So to cover us later, I'd quite like a test something like:

```scala
val agents = List[Agent](Agent("Henry Hippo"), Organisation("Wondrous Wellcome"))

val agentsJson = toJson(agents).get
val agentsFromJson = fromJson[List[Agent]](agentsJson).get

agentsFromJson shouldBe agents
```

to test that we deserialise the specialised types in JSON correctly.

(I don't know if that works or not; my home machine is still resolving dependencies, sigh.)

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraPublishers.scala
silveroliver: @alicefuzier no we dont have event as agent at the moment. In theory, in MARC, you can have a meeting as the creator of a work. I dont know if a) this actually happens in practice b) we think this is sensible.

Regarding QualifiedConcept I was going to suggest we introduce a QualifiedAgent. Bit yucky but bascially what is happening here. Think need some collective whiteboard time when I am back in.

alicefuzier: I see... I don't think this affects this PR 'cause if creators can be both `Agent` and `Event` and there is no common supertype for both, then we're a bit screwed anyway and we have to define creators as `List[Any]' (which I'd rather not do). But yeah, maybe we should take that into account

alicefuzier:
I think `WorkTest` and `ItemTest` pretty much cover all cases (In fact I have a failing test in `WorkTest`). I don't really think we need extra tests, it's just a refactoring that, as long as it doesn't change the current behaviour, then it's fine

alexwlchan:
KK. I managed to get Scala working, and indeed it mangles Organisation when it gets serialised via JSON.

My preference would be to add the extra test, so we're not relying on a side effect of another test to check this behaviour, and as a model test it's fairly cheap – but I won't press the issue.

alicefuzier:
Yep, that was why `WorkTest` and `ItemTest` were failing. Should be fine now

PR #1447, opened 2018-01-18
Don’t run the full test suite on every push to master

alexwlchan:
This is pulling what is hopefully big piece of low-hanging fruit from our Travis build times, without compromising too much on test coverage.

Our current policy is:

- PRs only run tests if they’re relevant
- Pushes to master run the full test suite, only publish if they’re relevant

This means we burn ~25m of build time for every push to master, even if it’s a one-character change to the README. This is patently silly.

This patch modifies master to only run tests if they’re relevant, so we should save a bunch of build time on master. Then I’ll enable nightly builds which run the full test suite – once a day, and outside office hours – so we still have the complete test suite running semi-regularly.

### DIFF
--- a/.travis/should_run_tests.py
+++ b/.travis/should_run_tests.py
@@ -12,11 +12,11 @@ def should_run_tests(task, travis_event_type):
     We skip doing the publish/deploy step when running a build on master that
     doesn't have any relevant changes since the last deploy.
     """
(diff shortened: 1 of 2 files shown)


alexwlchan:
I’m merging this straight to master to get the improvement immediately, but please yell on this PR if you disagree with what I’ve done.

PR #1448, opened 2018-01-19
Format Instants as ISO dates in dynamo

alicefuzier:
Write data that circe can parse into dynamo
### Who is this change for?
🔥 🔥 🔥


alexwlchan:
Tests?

Issue #1436, opened 2018-01-17
Make "title" optional on Work


alexwlchan:
So we aren't writing a title for deleted records just for the sake of writing a title.

PR #1455, opened 2018-01-23
Make title optional on work

alicefuzier:
🔝 
### Who is this change for?
🐬 🛵


alexwlchan:
Resolves #1436.

alicefuzier:
Well this change really removes any logic of creating a stub. The api results are filtered base on the visibility field on Work, so from an API perspective it doesn't matter if the Work is a stub or not.
So, for example for a suppressed Sierra recodrd, we might end up with a fully populated Work in elasticsearch that will never be displayed.

alexwlchan:
🎉

PR #1462, opened 2018-01-24
Extract items locations

alicefuzier:
### What is this PR trying to achieve?
Show locations on items
### Who is this change for?
🏰 🏯 🏡

PR #1463, opened 2018-01-25
Add publishers to the API

### LINKED ISSUE #1458: Staging API is returning 500s when trying to inspect the sierra-bibs-title index
From just now:

```
18:25:11.691 [ForkJoinPool-1-worker-69] ERROR u.a.w.p.a.f.e.GeneralExceptionMapper - Sending HTTP 500 from GeneralExceptionMapper
java.lang.RuntimeException: Unable to parse JSON as Work (com.fasterxml.jackson.databind.JsonMappingException: Can not construct instance of uk.ac.wellcome.models.AbstractAgent: abstract types either need to be mapped to concrete types, have custom deserializer, or contain additional type information
at [Source:
{
    "subjects": [],
    "visible": true,
    "description": null,
    "items": [],
    "identifiers": [
        {
            "identifierScheme": "sierra-system-number",
            "value": "b1030546"
        }
    ],
    "thumbnail": null,
    "canonicalId": "a24apyjm",
    "genres": [],
    "lettering": null,
    "publishers": [
        {
            "label": "Burgess and Hill,",
            "type": "Organisation"
        }
    ],
    "title": "An essay on the epidemic cholera of India",
    "type": "Work",
    "creators": [],
    "sourceIdentifier": {
        "identifierScheme": "sierra-system-number",
        "value": "b1030546"
    },
    "createdDate": null
}
; line: 1, column: 225] (through reference chain: uk.ac.wellcome.models.Work["publishers"]->com.fasterxml.jackson.module.scala.deser.BuilderWrapper[0])):
{
    "subjects": [],
    "visible": true,
    "description": null,
    "items": [],
    "identifiers": [
        {
            "identifierScheme": "sierra-system-number",
            "value": "b1030546"
        }
    ],
    "thumbnail": null,
    "canonicalId": "a24apyjm",
    "genres": [],
    "lettering": null,
    "publishers": [
        {
            "label": "Burgess and Hill,",
            "type": "Organisation"
        }
    ],
    "title": "An essay on the epidemic cholera of India",
    "type": "Work",
    "creators": [],
    "sourceIdentifier": {
        "identifierScheme": "sierra-system-number",
        "value": "b1030546"
    },
    "createdDate": null
}
```

alexwlchan:
* Add a publishers field to DisplayWork
* Create a new DisplayAgent class for the API, which is currently used only for publishers
* Use Circe to decode works that we read from Elasticsearch, so Circe is the only decoder of our internal models

Resolves #1458.

### Who is this change for?

📦 📖 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/models/DisplayAgent.scala
```diff
@@ -0,0 +1,23 @@
...
+  value = "Agent"
+)
+case class DisplayAgent(
+  @ApiModelProperty(
+    value = "The name of the agent"
+  ) label: String,
+  @JsonProperty("type") ontologyType: String = "Agent"
+)
+
+case object DisplayAgent {
```
alexwlchan: I tried to do regular classes with a distinction between DisplayAgent and DisplayOrganisation, but got bogged down in the tests. I'm proposing we do it like this for now, and if we want to define more specific types, we can do that in a separate patch.

alicefuzier: 👍

alicefuzier: BTW `Location` is being split into `DigitalLocation` and `PhysicalLocation` in #1462 and I ran into probably similar issues so, if we decide to do the same for `Agent`, we can adopt a similar solution hopefully

PR #1466, opened 2018-01-25
Suppress items with visible=false from the API

### LINKED ISSUE #1456: Make sure we don't show invisible items in the API

alexwlchan:
Resolves #1456.

### Who is this change for?

🃏


alexwlchan:
It occurs to me the API is maybe the wrong place to be filtering deleted items.

PR #1483, opened 2018-01-29
Adds SourceDataDao

kenoir:

Creates a DAO for the new unified source data table.

### Who is this change for?

🍥


common/src/main/scala/uk/ac/wellcome/dynamo/SourceDataDao.scala
```diff
@@ -0,0 +1,66 @@
...
+import uk.ac.wellcome.utils.GlobalExecutionContext.context
+
+import scala.concurrent.Future
+
+case class SourceData(
+  id: String,
+  version: Int,
+  reindexShard: String,
+  reindexVersion: Int,
+  ref: String
```
alexwlchan: What’s “ref” a reference to? A URL? An S3 object?

kenoir: Will rename in further branch.

PR #1486, opened 2018-01-30
Rename SourceDataDao to VersionedDao

alicefuzier:
More consistent model
### Who is this change for?
🐩


common/src/main/scala/uk/ac/wellcome/models/Versioned.scala
```diff
@@ -0,0 +1,11 @@
+package uk.ac.wellcome.models
+
+trait Versioned {
+  val version: Int
+  val sourceId: String
+  val sourceName: String
+}
+
+trait VersionUpdater[T <: Versioned] {
+  def updateVersion(versioned: T, newVersion: Int): T
```
alicefuzier: I tried putting this method on the trait, but there is no way to express that the return type should be the runtime type of the class. Also, I plan to read a bit about shapeless and figure out how we can generate this implicitly so that we don't have to reimplement the same thing for each case class

PR #1491, opened 2018-01-31
Use MurmurHash3

kenoir:

Better file hashing algorithm!

### Who is this change for?

:hash: 🌵 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


jamesgorrie:
:+1: to Murmur.

PR #1493, opened 2018-01-31
Make the transformer read from VHS 📼

alicefuzier:
Modify the transformer to receive `HybridRecord`s from SQS and read the contents from S3 pointers
### Who is this change for?
📼


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/receive/SQSMessageReceiver.scala
```diff
@@ -61,6 +73,19 @@ class SQSMessageReceiver(
     }
   }
 
-  def publishMessage(work: Work): Future[PublishAttempt] =
-    snsWriter.writeMessage(JsonUtil.toJson(work).get, Some("Foo"))
+  private def chooseTransformer(transformable: Transformable) = {
+    transformable match {
+      case _: CalmTransformable => new CalmTransformableTransformer
+      case _: MiroTransformable => new MiroTransformableTransformer
+      case _: SierraTransformable => new SierraTransformableTransformer
```
alicefuzier: This is the switch that we would like to avoid having. To be honest I don't know how we could not have this because this can only be a runtime check.

common/src/main/scala/uk/ac/wellcome/models/transformable/Transformable.scala
```diff
@@ -6,52 +6,46 @@ import uk.ac.wellcome.models.transformable.sierra.{
 }
 import io.circe.Decoder
 import cats.syntax.functor._
+import uk.ac.wellcome.models.Versioned
 import uk.ac.wellcome.utils.JsonUtil._
 
-sealed trait Transformable
-object Transformable {
```
alicefuzier: Don't need this IF and only if we ensure that the bib and item merger serialize stuff in s3 with the type discriminator. Should be as simple as using`versionedHybridStore.put(sierraTransformable: Transformable)` instead of `versionedHybridStore.put(sierraTransformable)`

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/receive/SQSMessageReceiver.scala
kenoir: How else could one perform an operation at runtime 🤔 

giphy 5

alicefuzier:
Sorry, most of the changes were renaming the various id fields to `sourceId` though

PR #1499, opened 2018-02-05
Write Terraform for the ReindexShardTracker table

alexwlchan:
📑


alexwlchan:
The reindex-job-creator is tested and working. \o/

PR #1501, opened 2018-02-05
Add the reindex job creator

alexwlchan:

Get updates from the ReindexShardTracker table, and push them to an SNS topic.

### Who is this change for?

~Economic stimulus.~ The reindexer rewrite.

🕴


catalogue_pipeline/reindex_job_creator/src/reindex_job_creator.py
```diff
@@ -0,0 +1,36 @@
...
+from wellcome_aws_utils.sns_utils import publish_sns_message
+
+
+def main(event, _ctxt):
+    print(f'event={event!r}')
+
+    sns_client = boto3.client('sns')
+    topic_arn = os.environ['TOPIC_ARN']
+
+    for record in event['Records']:
```
kenoir: Might be nice to push this into the utils so we get a generator for dynamo events (like with SNS).

alexwlchan: It already is, but I couldn't get it to work!

PR #1509, opened 2018-02-06
Refactor the reindexer to read from SQS and reindex the hybrid store

alexwlchan:
The reindexer was rather complicated. Modifying it to use SQS lets us delete a _bunch_ of code, which makes it better immediately. Modifying it to use the versionedDao has a similar effect. This is a big chunk of the work to get the reindexer working against the SourceData table.

I created a new `ReindexerWorkerService` to look similar to our existing SQSWorker-backed services, then made some small tweaks to `ReindexTargetService` – but since it only had a single public method, hooking into it was quite easy. Then I just changed things until it compiled cleanly.

What about the tests, I hear you say? Where we’re going, WE DON'T NEED TESTS.

(This is a lie. We do need tests, but probably not all the tests we currently have – a bunch of functionality is being stripped out of the reindexer, so we can start by deleting the ones we don’t need.)

### Who is this change for?

🐺


catalogue_pipeline/reindexer/src/main/scala/uk/ac/wellcome/platform/reindexer/services/ReindexTargetService.scala
```diff
@@ -12,91 +12,113 @@ import com.gu.scanamo.syntax._
...
-        scanamoUpdate(reindexItem.hashKey and reindexItem.rangeKey,
-                      set('ReindexVersion -> requestedVersion))
+      case Right(hybridRecord) => {
+        val existingRecord =
+          versionedDao.getRecord[HybridRecord](id = hybridRecord.id)
+        existingRecord.map { possibleRecord =>
+          // getRecord() returns an Option[HybridRecord] because you may be
+          // looking up a non-existent ID; since the ID came from the table we
+          // can assume the record exists!
+          val record = possibleRecord.get
```
kenoir: Ideally we might wrap this in something that logs a clear exception in the case of failure here.

catalogue_pipeline/reindexer_v2/reindex_worker/src/test/scala/uk/ac/wellcome/platform/reindexer/services/ReindexServiceTest.scala
```diff
@@ -81,16 +82,13 @@ class ReindexServiceTest
...
-        val reindexVersions = Scanamo
-          .scan[HybridRecord](dynamoDbClient)(tableName)
-          .map {
-            case Right(record) => record.reindexVersion
-          }
+    whenReady(reindexService.runReindex(reindexJob)) { _ =>
+
+        val hybridRecords = Scanamo.scan[HybridRecord](dynamoDbClient)(tableName).map (_.right.get)
+
+        hybridRecords.filter(_.reindexShard != shardName) should contain theSameElementsAs notInShardRecords
```
kenoir: 👍

alicefuzier:
`StartupTest` is failing. I think we can just delete those everywhere. They're just a leftover from early days but don't add anything now given that we have feature tests. What do you think?

PR #1513, opened 2018-02-07
VHS update function takes a transform function

kenoir:

A step towards separating metadata and asset data when using the VHS.

### Who is this change for?

🖼 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/models/Versioned.scala
+++ b/common/src/main/scala/uk/ac/wellcome/models/Versioned.scala
@@ -11,10 +11,6 @@ trait Sourced {
 
 trait Versioned extends Sourced {
   val version: Int
-
-  val sourceId: String
-  val sourceName: String
-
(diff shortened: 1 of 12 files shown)


common/src/main/scala/uk/ac/wellcome/models/Versioned.scala
```diff
@@ -4,18 +4,26 @@ import com.amazonaws.services.dynamodbv2.model.AttributeValue
...
+trait Sourced {
+  val sourceId: String
+  val sourceName: String
+}
+
+trait Versioned extends Sourced {
   val version: Int
+
   val sourceId: String
   val sourceName: String
```
alexwlchan: Do we need to duplicate sourceId and sourceName if we're extended Sourced?

alicefuzier: nope, you're right

PR #1514, opened 2018-02-07
Monitor overspends on our AWS budgets in Slack

alexwlchan:
This adds a small ECS task that monitors our AWS budgets, and pops an alarm in Slack if we're in overspend. This is meant to give us a little nudge to fix things if a budget starts drifting in the wrong direction.

We're adding this after the slight CloudWatch overspend – hopefully this would have caught it a little earlier, when the overall platform budget started to drift.


monitoring/main.tf
```diff
@@ -71,6 +71,16 @@ module "post_to_slack" {
...
 }
 
+module "slack_budget_bot" {
+  source = "slack_budget_bot"
+
+  slack_webhook           = "${var.non_critical_slack_webhook}"
+  release_ids             = "${var.release_ids}"
+  dashboard_bucket_id     = "${var.dash_bucket}"
+  account_id              = "${data.aws_caller_identity.current.account_id}"
+  ecs_services_cluster_id = "${local.ecs_services_cluster_id}"
```
kenoir: Should this be in the monitoring cluster?

PR #1507, opened 2018-02-06
[WIP] Compose Wrapper

kenoir:
### What is this PR trying to achieve?

Works with https://github.com/wellcometrust/dockerfiles/pull/15 to allow us to test our lambdas against containers.

### Who is this change for?

🐩 

### Have the following been considered/are they needed?

- [ ] Deployed new versions
- [ ] Run `terraform apply`.

PR #1518, opened 2018-02-08
Add machinery for running docker-compose in pytest

alexwlchan:
This picks up the idea in #1507, integrates with https://github.com/wellcometrust/dockerfiles/pull/16, and allows us to run containers alongside our Lambda tests (if you so wish).

You write a docker-compose.yml in your `src` directory, bring in a fixture from `lambda_conftest.py`, and you get a boto3 resource/client that connects to your container.


builds/build_lambda_test_image.sh
```diff
@@ -39,26 +35,25 @@ fi
 
 # If a requirements.txt file exists, we need to check if it's more up-to-date
 # than the existing Lambda image, and rebuild if so.
```
alexwlchan: This comment is out-of-date.

alexwlchan:
I think this is ready for review; in particular I’d like @kenoir and @alicefuzier to look and check they think they could write new docker-compose tests with this setup.

Issue #1511, opened 2018-02-07
Improve S3 bucket naming conventions


jtweed:
S3 bucket naming is a bit all over the place.

I think we need to come up with a consistently applied set of conventions and rename existing buckets, e.g. (names for discussion):

* `wellcomecollection-adapters-sierra` not `wellcomecollection-sierra-reader-data`
* `wellcomecollection-adapters-calm` not `calm-data`
* `wellcomecollection-infra-config` not `platform-infra`
* `wellcomecollection-infra-logs-alb` not `wellcomecollection-alb-logs`
* `wellcomecollection-infra-logs-cloudfront` not `wellcome-platform-cloudfront-logs`
* `wellcomecollection-collections-data` not `wellcomecollection-source-data`

We should also remove any buckets that are empty and not being used.

The rational for this is that as the number of buckets increases, I'm getting increasingly concerned that we begin to lose sight over what they are for. This is obviously a real issue when they start to contain master assets and data for the collections.

Those irreplaceable buckets need to be very obvious and very clear in the context of everything else.

alexwlchan:
To record the outcome of this afternoon’s discussion:

In a new account, we have two buckets that store our irreplaceable assets/data:

* wellcomecollection<strong>-assets-primary</strong>
* wellcomecollection<strong>-assets-secondary</strong>

These buckets are versioned, replicated, kept in separate geographical regions, use multi-factor auth deletes, and protected by a fleet of nuclear submarines.

In the current account, we have another bucket that stores the asset derivatives, which could be replaced but it would be expensive to do so:

* wellcomecollection<strong>-assets-derivatives</strong>

Our other buckets gain a **-platform** or **-public** prefix – although they’re associated with the wellcomedigitalplatform AWS account, bucket names are globally unique. e.g.:

* wellcomecollection<strong>-platform-adapters-miro</strong>
* wellcomecollection<strong>-platform-infra-*</strong>
* wellcomecollection<strong>-platform-logs-*</strong>

* wellcomecollection<strong>-public-*</strong>

alexwlchan:
As of just now, these are the buckets I think we still want to rename or remove:

* miro-images-sync
* wellcomecollection-client-transfer, maybe?
* wellcomecollection-editorial-photography, maybe?
* wellcomecollection-images
* wellcomecollection-miro-images-public
* wellcomecollection-source-data

alexwlchan:
How about **wellcomecollection-vhs-sourcedata**?

[...]

PR #1521, opened 2018-02-08
Rename the bucket that holds our ALB logs

alexwlchan:
This name was less bad, but now it's consistent with CloudFront: **wellcomecollection-platform-logs-alb**. I’m migrating the existing logs to the new bucket (even though they expire after 30 days).

Part of #1511.

### Who is this change for?

📛


alexwlchan:
The old logs have been migrated successfully, and the ALB is writing logs to the new bucket. Merging.

PR #1522, opened 2018-02-08
Rename the bucket that has the sierra_adapter data

alexwlchan:
Another step in the direction of #1511.

### Who is this change for?

📛


alexwlchan:
> We will need to manually sync the data over too.

Doing it now!

alexwlchan:
Done!

PR #1524, opened 2018-02-09
Rename the bucket that holds our dashboard

alexwlchan:
Another slice of #1511.


alexwlchan:
This change is deployed, data synced, and old buckets cleaned up.

PR #1526, opened 2018-02-09
Fix flaky bib merger tests

kenoir:

Ensure tests are more reliable in master.

### Who is this change for?

🌐 

- [ ] Deployed new versions


sierra_adapter/sierra_bib_merger/src/test/scala/uk/ac/wellcome/platform/sierra_bib_merger/SierraBibMergerFeatureTest.scala
```diff
@@ -232,19 +210,10 @@ class SierraBibMergerFeatureTest
       ),
       modifiedDate = oldUpdatedDate
     )
-    sendBibRecordToSQS(record)
 
-    // Blocking in Scala is generally a bad idea; we do it here so there's
-    // enough time for this update to have gone through (if it was going to).
-    Thread.sleep(5000)
```
alexwlchan: So while I agree this is un-Scala like, I think we need it here – or the test could spuriously pass because we check DynamoDB *before* it applies the bad update.

Unless this was the cause of the fault?

PR #1531, opened 2018-02-13
Provide more useful descriptions in our SNS subjects

alexwlchan:
We have an SQS message on the transformer DLQ with

```
  "Subject" : "default-subject",
  "Message" : "null",
```

and it's really not obvious where it came from!

This patch modifies the subject in our SNS writers to provide slightly more helpful descriptions, so hopefully we'll be able to trace this sort of message in future!

### DIFF
--- a/catalogue_pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/services/IdMinterWorkerService.scala
+++ b/catalogue_pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/services/IdMinterWorkerService.scala
@@ -30,7 +30,7 @@ class IdMinterWorkerService @Inject()(
       workWithCanonicalId <- idEmbedder.embedId(json)
       _ <- writer.writeMessage(
         message = workWithCanonicalId.toString(),
-        subject = "source: IdMinterWorkerService.processMessage"
+        subject = s"source: ${this.getClass.getSimpleName}.processMessage"
       )
     } yield ()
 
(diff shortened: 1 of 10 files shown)


shared_infra/dynamo_to_sns/src/dynamo_to_sns.py
```diff
@@ -53,5 +54,6 @@ def main(event, _ctxt=None, sns_client=None):
         publish_sns_message(
             sns_client=sns_client,
             topic_arn=topic_arn,
-            message=message
+            message=message,
+            subject=f'source: dynamo_to_sns ({topic_name})'
```
kenoir: This is python! 

`s" source: dynamo_to_sns ($topic_name)`

catalogue_pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/services/IdMinterWorkerService.scala
```diff
@@ -28,8 +28,10 @@ class IdMinterWorkerService @Inject()(
     for {
       json <- Future.fromTry(parseMessageIntoJson(message))
       workWithCanonicalId <- idEmbedder.embedId(json)
-      _ <- writer.writeMessage(workWithCanonicalId.toString(),
-                               Some(snsSubject))
+      _ <- writer.writeMessage(
+        message = workWithCanonicalId.toString(),
+        subject = "source: IdMinterWorkerService.processMessage"
```
alicefuzier: I'd rather not hardcode the name of the service in a string. It's easy to forget renaming it if we refactor. You could use `this.getClass.getSimpleName` instead

alexwlchan: I didn’t realise you could do that, thanks!

Issue #1512, opened 2018-02-07
Shard data in source data bucket


jtweed:
- [x] New records written by the VHS are sharded (#1540)
- [x] Existing records are re-sharded (#1545)

jtweed:
Presumably makes sense to rename this bucket to the new scheme as part of doing this.

alexwlchan:
@jtweed I‘m not opposed to renaming the bucket, but let's do one change at a time. 😄

alexwlchan:
The existing records have been copied into new shards; I still have to clean up the old shards.

alexwlchan:
All the existing records have been resharded correctly.

👌

PR #1545, opened 2018-02-14
Write a resharder

alexwlchan:
This is a temporary resharder for #1512.

It gets attached to the VHS event stream, and then we'll reindex the entire table. It should reshard everything in the S3 bucket, then we can:

1. Delete the Lambda
2. Delete the unsharded files


kenoir:
This is specifically to reshard the S3 bucket for the VHS yes?

alexwlchan:
Yes. We’ll attach it to the VHS event stream, reindex the table, then delete the “resharded” flag and all of this code.

PR #1547, opened 2018-02-15
Move the reindexer code into a separate folder

alexwlchan:
Travis is [failing on master](https://travis-ci.org/wellcometrust/platform/jobs/341814879) because it's looking in `catalogue_pipeline/reindex_worker` for the reindexer code to build a Docker image, and it can't find anything.

Rather than delve too heavily into those build scripts, this patch just moves all the reindexer into its own directory, so there's only one level of nesting. Should get the build working again.


alexwlchan:
So after the first reindex “completed”, here’s what we‘ve learnt:

* Not all the reindex jobs got created, because the reindex_job_creator kept timing out.
* The complete_reindex Lambda is looking for a version field that doesn't exist
* We are hitting the upper limit of DynamoDB autoscaling quite badly

PR #1548, opened 2018-02-15
Janusgraph prototype

alicefuzier:
See how the relater would look like if we decided to use jenusgraph
### Who is this change for
⛺️ 🍔


kenoir:
Closing as done.

PR #1553, opened 2018-02-16
Allow complete_reindex lambda to handle missing version

kenoir:

Allow complete_reindex lambda to handle missing version!

### Who is this change for?

📴 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


kenoir:
Supersedes: https://github.com/wellcometrust/platform/pull/1552

reindexer/complete_reindex/src/complete_reindex.py
```diff
@@ -2,31 +2,80 @@
...
-def main(event, _ctxt=None, dynamodb_client=None):
-    print(f'event = {event!r}')
-    dynamodb_client = dynamodb_client or boto3.client('dynamodb')
+def _get_item_version(dynamo_item):
+    dynamo_item_version = 1
+
+    if 'version' in dynamo_item:
+        dynamo_item_version = dynamo_item['version']
+
+    return dynamo_item_version
```
alexwlchan: Easier would be to use the `get` method on Python dicts: `dict.get(key[, default]`

i.e.

```python
dynamo_item.get('version', 1)
```

reindexer/complete_reindex/src/complete_reindex.py
```diff
@@ -2,31 +2,80 @@
...
+    dynamodb_response = table.get_item(Key={'shardId': shard_id})
 
-    table_name = os.environ['TABLE_NAME']
+    dynamo_item = dynamodb_response['Item']
 
+    print(f'Retrieved {dynamo_item}')
+
+    dynamo_current_version = dynamo_item['currentVersion']
+
+    if (dynamo_current_version >= completed_reindex_version):
```
alexwlchan: nit: drop the parens

reindexer/complete_reindex/src/complete_reindex.py
```diff
@@ -2,31 +2,80 @@
...
-                }
-            )
-        except ClientError as err:
-            if err.response['Error']['Code'] == 'ConditionalCheckFailedException':
-                print(f'{row["shardId"]} already has a newer currentVersion')
-            else:
-                raise
+        job = _process_reindex_tracker_update_job(table, record.message)
+
+    if job is not None:
```
alexwlchan: Should this be in the body of the loop? – you construct a job for every update, but only process the last one. Possibly needs extra tests?

reindexer/complete_reindex/src/complete_reindex.py
```diff
@@ -2,31 +2,80 @@
...
+
+    return dynamo_item_version
+
+
+def _update_versioned_item(table, dynamo_item, desired_item):
+    desired_dynamo_conditional_update_version = (
+            _get_item_version(dynamo_item) + 1
+    )
+
+    new_dynamo_item = {**dynamo_item, **desired_item}
```
alexwlchan: This construction is quite unusual – could we just pass in the item we want to PUT?

PR #1568, opened 2018-02-19
Transformable: inherit from Sourced, not Versioned/Reindexable

### LINKED ISSUE #1562: Clean up Transformable
Remove version, reindexVersion & reindexShard from Transformable as they are not needed on that model and their presence could cause confusion.

alexwlchan:
This lets us delete several fields on this model that we weren’t using.

Along the way, I had to clean up the Sourced/Versioned models, as there were several things defined on Versioned that really belong on Sourced.

I’m still working on fixing the problems caused by the Sourced/Versioned changes.

Resolves #1562.

### Who is this change for?

☘️


alexwlchan:
Moving this back to WIP while I continue to fix test failures.

alexwlchan:
All the state for this change has been migrated successfully. 🎉

PR #1570, opened 2018-02-19
Add version to work

alicefuzier:
that
### Who is this change for?
🍸

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/models/Work.scala
+++ b/common/src/main/scala/uk/ac/wellcome/models/Work.scala
@@ -6,6 +6,7 @@ import uk.ac.wellcome.utils.JsonUtil._
 /** A representation of a work in our ontology */
 case class Work(title: Option[String],
                 sourceIdentifier: SourceIdentifier,
+                version: Int,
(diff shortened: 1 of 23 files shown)


common/src/main/scala/uk/ac/wellcome/models/Work.scala
```diff
@@ -6,6 +6,7 @@ import uk.ac.wellcome.utils.JsonUtil._
 /** A representation of a work in our ontology */
 case class Work(title: Option[String],
                 sourceIdentifier: SourceIdentifier,
+                version: Int,
```
kenoir: Maybe this should make use of the `Versioned` trait from @alexwlchan's PR?

alicefuzier: 👍

alicefuzier: Actually, I agree, but since this is mostly ready to go and making `Work` extend `Versioned` should not cause any other change, I think we can do that once @alexwlchan pR is merged. In the meantime this can go as is

Issue #1544, opened 2018-02-14
Move and/or disambiguate the platform-infra bucket


alexwlchan:
Related: #1511.

alexwlchan:
This gets its own ticket because it's potentially disruptive. This might be a good opportunity to break up what's in the bucket, if we think it's a good idea (e.g. Lambdas and releases go to a different bucket from terraform state/vars).

PR #1572, opened 2018-02-20
reindexer: move the associated Terraform into a separate stack

alexwlchan:
The reindexer isn't really part of the catalogue pipeline – it's a separate set of services.

To keep our Terraform stacks small and fast, this pulls the reindexer pieces into a separate stack before we add the relater and merger.

This also creates the new infra bucket in line with #1511, #1544.


alexwlchan:
The only failing tests are in sbt-common, and that's because we're downstream of master.

PR #1577, opened 2018-02-20
reindex_worker: simplify the implementation to avoid leaning so heavily on Scanamo

alexwlchan:
Previously in the reindexer:

We'd cribbed some fairly complicated QueryStream code from Scanamo so we didn't load an entire result list into memory at once and have an OutOfMemory error. Now that:

* Reindex shards contain fewer records
* Individual records are also smaller

it's worth revisiting this code. Do we still hit OOM errors, or can we load an entire reindex shard into memory at once? This patch does exactly that – and then processes the entire reindex.

I'm hoping this will also clear up the weirdness we're seeing where the reindexer isn't always completing jobs correctly.

I'm going to deploy this patch immediately and trigger a reindex, to see how it behaves.


reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex_worker/services/ReindexService.scala
```diff
@@ -31,130 +30,61 @@ class ReindexService @Inject()(dynamoDBClient: AmazonDynamoDB,
...
-      .map(result =>
-        if (result.contains(false)) {
-          throw GracefulFailureException(
-            new RuntimeException(
-              "Not all records were successfully processed!"
-            ))
-        } else {
-          info(s"Successfully processed reindex job $reindexJob")
-      })
+    Future.sequence(updates)
```
alexwlchan: This returns `Future[List[Unit]]`, which may not behave correctly with the for comprehension in the worker service.

alexwlchan: Also, I don't log anything here. 🙈

alicefuzier:
Uhm, I assume you've discussed this with @kenoir so I may be outnumbered, but I think it not loading everything into memory is a nice feature and I'd be in favour of keeping it. Shard sizes are managed elsewhere and it's not easy to spot that if/when we change the shard size we should also increment the memory.

alexwlchan:
@alicefuzier Yep, I discussed with Robert – we thought it would be a benefit to get rid of the opaque Scanamo code if we could, where “if we could” is mostly determined by memory concerns. My gut feeling is that it’s less of an issue than we think it is, but we’ll find out if I’m right!

We had a suspicion that something in that code was catching and ignoring throughout errors – it seems suspicious that the reindeer never hits/logs them – and thought getting rid of it might help us spot the underlying issue.

alexwlchan:
On the topic of shard size: what determines “the best shard size”, if not the reindexer? What else affects that choice?

We tried to average it at 1500 (although it’s become clear they’re not very evenly spread!) precisely to keep in doable within the life of one SQS message. I don’t think it’s that bad for the reindexer to have slightly stronger requirements on that choice.

kenoir:
@alexwlchan @alicefuzier this ain't a democracy!  :) We'll chat about the various solutions on their merits tomorrow.

alicefuzier:
Actually I completely changed my mind on this. It would be nice to have this a reindexer feature, but it is very obscure code and if we can make it simpler removing something we don't explicitly need, that's good. And yeah, shard sizes will never be massive, 'cause anyway the reindexer reads from a queue and it needs to finish within the visibility timeout.
Making it use less memory would only be useful to reduce costs in scaling it up, which we don't care that much about that.

alexwlchan:
FWIW, we ran a complete reindex of Sierra last night using this code. It’s still not running to completion, but:

* There's nothing in CloudWatch to suggest it ran out of memory
* The RST table is correctly reporting that those shards didn't finish

PR #1585, opened 2018-02-22
.travis: load our AWS credentials from a file, not from environment variables

alexwlchan:
Previously Travis loaded AWS credentials from the environment, whereas we use the ~/.aws directory for all local operations. This brings Travis into line with local workflows, so we can simplify some of the `docker run` workflows in a separate PR.


kenoir:
Travis is sad because" `botocore.exceptions.NoCredentialsError: Unable to locate credentials`

alexwlchan:
Ah, it wants AWS credentials to use during tests. I think that's fixable (and really we probably shouldn't be using real credentials in tests!).

PR #1591, opened 2018-02-22
Split Work into Unidentifiedwork and IdentifiedWork

alicefuzier:
Remove the id field from `Work` that throws an exception if `canonicalId` is `None`
### Who is this change for?
🐻 🏖

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/models/Work.scala
+++ b/common/src/main/scala/uk/ac/wellcome/models/Work.scala
@@ -4,55 +4,57 @@ import com.sksamuel.elastic4s.Indexable
 import uk.ac.wellcome.utils.JsonUtil._
 
 /** A representation of a work in our ontology */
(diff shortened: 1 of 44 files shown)


common/src/main/scala/uk/ac/wellcome/models/Work.scala
```diff
@@ -4,26 +4,57 @@ import com.sksamuel.elastic4s.Indexable
 import uk.ac.wellcome.utils.JsonUtil._
 
 /** A representation of a work in our ontology */
-case class Work(title: Option[String],
+trait Work extends Versioned with Identifiable{
+  val title: Option[String]
+val  sourceIdentifier: SourceIdentifier
+val                version: Int
```
kenoir: Wat is going on with this whitespace?

alicefuzier: For some reason travis is not formatting...

PR #1594, opened 2018-02-22
Use tenacity in complete_reindex lambda

kenoir:

Implement some of our own retry/backoff logic in the complete_reindex lambda to allow time for dynamo autoscaling.

### Who is this change for?

🍡 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.

### DIFF
--- a/reindexer/complete_reindex/src/complete_reindex.py
+++ b/reindexer/complete_reindex/src/complete_reindex.py
@@ -4,10 +4,23 @@ import os
 
 from boto3.dynamodb.conditions import Attr
 import boto3
-
+from botocore.exceptions import ClientError
+from tenacity import *
 from wellcome_aws_utils import sns_utils
 
 
+def should_retry(err):
+    if isinstance(err, ClientError):
+        if err.response['Error']['Code'] == 'ProvisionedThroughputExceededException':
+            return True
+
+    return False
+
+
+@retry(
+    wait=wait_exponential(multiplier=1, max=10),
+    retry=retry_if_exception(should_retry)
+)
 def _update_versioned_item(table, item):
     print(f'Attempting conditional update to {item}')
 
@@ -21,6 +34,10 @@ def _update_versioned_item(table, item):
     print(f'Successful update to {item}')
 
 
+@retry(
+    wait=wait_exponential(multiplier=1, max=10),
+    retry=retry_if_exception(should_retry)
+)
 def _process_reindex_tracker_update_job(table, message):
     shard_id = message['shardId']
     completed_reindex_version = message['completedReindexVersion']
--- a/reindexer/complete_reindex/src/requirements.txt
+++ b/reindexer/complete_reindex/src/requirements.txt
@@ -1,1 +1,0 @@
-wellcome_aws_utils==2.0.1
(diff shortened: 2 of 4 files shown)


reindexer/complete_reindex/src/complete_reindex.py
```diff
@@ -4,10 +4,23 @@
...
 import boto3
-
+from botocore.exceptions import ClientError
+from tenacity import *
 from wellcome_aws_utils import sns_utils
 
 
+def should_retry(err):
+    if isinstance(err, ClientError):
+        if err.response['Error']['Code'] == 'ProvisionedThroughputExceededException':
```
alexwlchan: minor: you could simplify this even further:

```python
return err.response['Error']['Code'] == 'ProvisionedThroughputExceededException'
```

reindexer/complete_reindex/src/requirements.txt
```diff
@@ -1 +0,0 @@
-wellcome_aws_utils==2.0.1
```
alexwlchan: You need to compile the `requirements.txt` (I think there's a Make task for it), or the deps won't end up in the image.

reindexer/complete_reindex/src/test_complete_reindex.py
```diff
@@ -70,6 +70,25 @@ def _create_event(shard_id, current_version, desired_version):
...
+
+    err_bad = ClientError({
+        'Error': {
+            'Code': 'Bangarang!',
+            'Message': 'oops'
+        }
+    },'testing')
+
+    assert should_retry(err_good) == True
+    assert should_retry(err_bad) == False
```
alexwlchan: And a test that `should_retry(AssertionError("foo"))` is False?

reindexer/complete_reindex/src/complete_reindex.py
kenoir: Hmm, what if it's not a ClientError?

reindexer/complete_reindex/src/requirements.txt
kenoir: Might have to show me where this is, sorry.

reindexer/complete_reindex/src/complete_reindex.py
alexwlchan: Then you won't enter this branch.

To be clear, I meant 14–15, so you’d have

```python
if isinstance(err, ClientError):
    return err.response['Error']['Code'] == 'ProvisionedThroughputExceededException'

return False
```

reindexer/complete_reindex/src/test_complete_reindex.py
kenoir: Yup, good plan.

reindexer/complete_reindex/src/requirements.txt
alexwlchan:
```console
$ cd ~/repos/platform
$ make $(pwd)/reindexer/complete_reindex/src/requirements.txt
```

reindexer/complete_reindex/src/complete_reindex.py
kenoir: Ah yeah. Good.

PR #1597, opened 2018-02-22
Store our Lambdas in the new infra bucket

alexwlchan:
Another piece of #1544 – Lambdas now live in wellcomecollection-platform-infra.

I’m going to get this migrated, and merge it as soon as tests go green, or we'll have conflicting Terraform state.


alexwlchan:
releases have come along for the ride too.

Issue #1565, opened 2018-02-19
Implement service for storing transformed works


kenoir:
We will need to be able to look up transformed unidentified works by their source identifier, in order to do this we will implement a service which use the VHS to store these works after transformation.

PR #1600, opened 2018-02-23
SourcedObjectStore -> ObjectStore

kenoir:

Simplify SourcedObjectStore to remove the necessity for stored objects to be stored.

### Who is this change for?

Developers who would like to make the object store as simple as possible in order to use it in the recorder service.

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


kenoir:
Part of https://github.com/wellcometrust/platform/issues/1565

common/src/main/scala/uk/ac/wellcome/s3/S3ObjectStore.scala
```diff
@@ -15,39 +15,55 @@ import scala.concurrent.Future
...
+    val keyPrefix = keyPrefixGenerator.generate(sourcedObject)
+    S3ObjectStore.put[T](s3Client, bucketName)(keyPrefix)(sourcedObject)
+  }
+
+  def get(key: String)(implicit decoder: Decoder[T]): Future[T] = {
+    S3ObjectStore.get[T](s3Client, bucketName)(key)
+  }
+}
+
+object S3ObjectStore extends Logging {
```
alexwlchan: Having both a class and an object of the same name feels like it's asking for trouble.

Maybe the class could be `KeyPrefixedS3ObjectStore`?

common/src/main/scala/uk/ac/wellcome/s3/S3ObjectStore.scala
```diff
@@ -15,39 +15,55 @@ import scala.concurrent.Future
...
-      // characters of the ID and reverse them.  This ensures that:
-      //
-      //  1.  It's easy for a person to find the S3 data corresponding to
-      //      a given source ID.
-      //  2.  Adjacent objects are stored in shards that are far apart,
-      //      e.g. b0001 and b0002 are separated by nine shards.
-      //
-      val s3Shard = sourcedObject.sourceId.reverse
-        .slice(0, 2)
+      val prefixCleaningRegex = "^/|/$".r
```
alexwlchan: I have no idea what this regex is meant to accomplish, or whether it's doing it correctly. Comment please!

(More thoughts: https://alexwlchan.net/2016/04/regexes-are-code/)

alexwlchan: Also, this feels like an odd boundary violation – both the class and the companion object have some logic for picking the key prefix. Surely better to move this up into the class, and have all the key prefix logic there?

alexwlchan: Okay, so Alice has explained to me that this is trimming slashes from the string. Maybe use `stripPrefix` and `stripSuffix` for readability? i.e.

```
scala> "/foo/".stripPrefix("/").stripSuffix("/")
res1: String = foo
```

alicefuzier: I think this should live in the object because you want to protect yourself from double slashes in both cases butt 👍 for the `stripPrefix` and `stripSuffix`. Didn't know you could do that!

common/src/main/scala/uk/ac/wellcome/s3/S3ObjectStore.scala
kenoir: This is  a "companion object" which is not unusual in Scala:  http://tutorials.jenkov.com/scala/singleton-and-companion-objects.html

common/src/main/scala/uk/ac/wellcome/s3/S3ObjectStore.scala
```diff
@@ -15,39 +15,55 @@ import scala.concurrent.Future
 import scala.io.Source
 import scala.util.hashing.MurmurHash3
 
-class SourcedObjectStore @Inject()(
+trait KeyPrefixGenerator[-T] {
```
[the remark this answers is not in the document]
kenoir: 👍 ... or at least a pointer to some documentation explaining it.

common/src/main/scala/uk/ac/wellcome/s3/S3ObjectStore.scala
```diff
@@ -15,39 +15,55 @@ import scala.concurrent.Future
...
       val contentHash = MurmurHash3.stringHash(content, MurmurHash3.stringSeed)
 
-      // To spread objects evenly in our S3 bucket, we take the last two
-      // characters of the ID and reverse them.  This ensures that:
-      //
-      //  1.  It's easy for a person to find the S3 data corresponding to
-      //      a given source ID.
-      //  2.  Adjacent objects are stored in shards that are far apart,
-      //      e.g. b0001 and b0002 are separated by nine shards.
-      //
```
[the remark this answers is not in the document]
kenoir: This logic is now in the `KeyPrefixGenerator` - this comment should have followed it there. Will re-add in the appropriate place.

common/src/main/scala/uk/ac/wellcome/s3/S3ObjectStore.scala
```diff
@@ -0,0 +1,100 @@
...
+import scala.io.Source
+import scala.util.hashing.MurmurHash3
+
+// KeyPrefixGenerator is contravariant for type T
+//
+// Practically this means `KeyPrefixGenerator[Sourced]` is a
+// subclass of `KeyPrefixGenerator[SierraTransformable]`
+// so it will be accepted in the S3ObjectStore for all types
+// extending Sourced. This prevents us from having to write
+// SierraTransformableKeyPrefixGenerator etc. etc.
```
alexwlchan: Yay comments! Does this still make sense to you today?

I might go further, and be explicit about the relationship between `Sourced` and `SierraTransformable` (which may not be obvious if you're not familiar with our codebase.

Something like (spoilered if you'd rather have a go at writing it yourself):

<details>
<summary>Spoiler block</summary>

```scala
// '-T' means KeyPrefixGenerator is contravariant for type T
//
// This means that if S is a subclass of T, then KeyPrefixGenerator[T]
// is a subclass of KeyPrefixGenerator[S] (the class hierarchy is reversed).
//
// For example,
//
//      Sourced < SierraTransformable
//
// but
//
//      KeyPrefixGenerator[SierraTransformable] < KeyPrefixGenerator[Sourced]
//
// In S3ObjectStore, we need a KeyPrefixGenerator[T], where T is usually some
// subclass of Sourced (this is the way we use it, not a strict requirement).
// This contravariance allows us to define a single KeyPrefixGenerator[Sourced]
// and use it on all instances, even though T is some subclass of Sourced.
```

</details>

common/src/main/scala/uk/ac/wellcome/s3/S3ObjectStore.scala
```diff
@@ -0,0 +1,100 @@
...
+  }
+}
+
+class S3ObjectStore[T] @Inject()(
+  s3Client: AmazonS3,
+  @Flag("aws.s3.bucketName") bucketName: String,
+  keyPrefixGenerator: KeyPrefixGenerator[T]
+) extends Logging {
+  def put(sourcedObject: T)(implicit encoder: Encoder[T]): Future[String] = {
+
```
alexwlchan: nit: lose this empty line

common/src/main/scala/uk/ac/wellcome/storage/VersionedHybridStore.scala
```diff
@@ -26,8 +26,8 @@ case class HybridRecord(
     with Sourced
     with Versioned
 
-class VersionedHybridStore @Inject()(
-  sourcedObjectStore: SourcedObjectStore,
+class VersionedHybridStore[T <: Sourced] @Inject()(
```
alexwlchan: Do we need to subclass from Sourced? I don't think it's integral to the VHS, which just needs an "id". Maybe at some point we want to remove this restriction?

common/src/main/scala/uk/ac/wellcome/s3/S3ObjectStore.scala
kenoir: Still makes sense, but i'm going to steal your spoiler!

common/src/main/scala/uk/ac/wellcome/storage/VersionedHybridStore.scala
kenoir: That's where we're going: https://github.com/wellcometrust/platform/pull/1604

PR #1604, opened 2018-02-26
Remove Sourced and Reindexable from VHS

alicefuzier:
Remove dependency on `Sourced` and `Reindexable` in the VHS while retaining the ability to pass arbitrary fields in the HybridRecord
### Who is this change for?
🐐


common/src/main/scala/uk/ac/wellcome/dynamo/VersionedDao.scala
```diff
@@ -2,57 +2,164 @@ package uk.ac.wellcome.dynamo
...
 import uk.ac.wellcome.utils.GlobalExecutionContext.context
 
+import scala.collection.generic.CanBuildFrom
 import scala.concurrent.Future
 
+trait UpdateExpressionGenerator[T] {
+  def generateUpdateExpression(t: T): Option[UpdateExpression]
+}
+
+object UpdateExpressionGenerator {
```
alexwlchan: This is what creates the update expressions that we can throw the type checker at, yes?

So the theory roughly goes: the "Witness" allows us to find objects who have an attribute "id" of any type, and then … waves hands … an update expression pops out?

common/src/main/scala/uk/ac/wellcome/dynamo/VersionedDao.scala
```diff
@@ -2,57 +2,164 @@ package uk.ac.wellcome.dynamo
...
+      at[FieldType[K, V]] { fieldtype =>
+        {
+          val fieldValue: V = fieldtype
+          set(witness.value -> fieldValue)
+        }
+      }
+    }
+  }
+
+  object Folder extends Poly2 {
```
alexwlchan: And this does a left fold over update expressions? So we can combine them into one?

Why is this extending Poly2 instead of Poly1? Is there a difference?

common/src/main/scala/uk/ac/wellcome/dynamo/VersionedDao.scala
```diff
@@ -2,57 +2,164 @@ package uk.ac.wellcome.dynamo
...
+
+  def create[T](f: (T) => Option[UpdateExpression]) =
+    new UpdateExpressionGenerator[T] {
+      override def generateUpdateExpression(record: T) = f(record)
+    }
+
+  implicit def elementUpdateExpressionGenerator[L <: HList,
+                                                A,
+                                                B <: HList,
+                                                C <: HList](
```
alexwlchan: I have no idea what this does.

common/src/main/scala/uk/ac/wellcome/dynamo/VersionedDao.scala
```diff
@@ -2,57 +2,164 @@ package uk.ac.wellcome.dynamo
...
+    create { t: L =>
+      {
+        val recordAsHlist = t - 'id
+        val nonIdTaggedFields = recordAsHlist.map(toUpdateExpressions)
+
+        nonIdTaggedFields.foldLeft[Option[UpdateExpression]](None)(Folder)
+      }
+    }
+
+  implicit def productUpdateExpressionGenerator[C, T](
```
alexwlchan: Is "product" a case class, or something else?

common/src/main/scala/uk/ac/wellcome/dynamo/package.scala
```diff
@@ -11,4 +14,13 @@ package object dynamo {
     )(
       _.toString
     )
+
+  // DynamoFormat for tagged HLists
```
alexwlchan: I think I know what an HList is (although a reminder would be helpful, as it seems to be a shapeless concept) – what’s a tagged HList?

common/src/main/scala/uk/ac/wellcome/storage/HybridRecordEnricher.scala
```diff
@@ -0,0 +1,38 @@
+package uk.ac.wellcome.storage
+
+import shapeless.ops.hlist.{Collect, Prepend, Zip}
+import shapeless.{Id => ShapelessId, _}
+import uk.ac.wellcome.models.Id
+
+trait HybridRecordEnricher[T] {
```
alexwlchan: What does it mean to enrich a record?

common/src/main/scala/uk/ac/wellcome/storage/VersionedHybridStore.scala
```diff
@@ -1,54 +1,45 @@
...
 ) {
 
-  implicit val hybridRecordVersionUpdater = new VersionUpdater[HybridRecord] {
-    override def updateVersion(testVersioned: HybridRecord,
-                               newVersion: Int): HybridRecord = {
-      testVersioned.copy(version = newVersion)
-    }
-  }
-
+  case class EmptyCC()
```
alexwlchan: CC = "case class"?

common/src/main/scala/uk/ac/wellcome/type_classes/IdGetter.scala
```diff
@@ -0,0 +1,45 @@
...
+  type IdField = FieldType[IdKey, String]
+
+  def apply[A](implicit enc: IdGetter[A]): IdGetter[A] =
+    enc
+
+  def createIdGetter[T](f: T => String): IdGetter[T] = new IdGetter[T] {
+    def id(t: T) = f(t)
+  }
+
+  // Generates an IdGetter for an HList of arbitrary length with the id field as the first field
```
alexwlchan: Why does it matter whether the id field is the first or last field?

common/src/main/scala/uk/ac/wellcome/dynamo/VersionedDao.scala
alicefuzier: Yes, it's a case class in shapeless terminology

common/src/main/scala/uk/ac/wellcome/storage/HybridRecordEnricher.scala
kenoir: In this case it's appending the fields from the metadata type handed in.

common/src/main/scala/uk/ac/wellcome/storage/VersionedHybridStore.scala
kenoir: Yup.

common/src/main/scala/uk/ac/wellcome/dynamo/package.scala
kenoir: I think it's a field key and value pair.

common/src/test/scala/uk/ac/wellcome/type_classes/VersionUpdaterTest.scala
```diff
@@ -0,0 +1,35 @@
...
+import shapeless.LabelledGeneric
+
+class VersionUpdaterTest extends FunSpec with Matchers {
+
+  case class TestVersion(something: String,
+                         version: Int,
+                         somethingElse: String)
+
+  it("creates a VersionUpdater instance for an hlist") {
+    val gen = LabelledGeneric[TestVersion]
```
alexwlchan: this is the hlist generator

kenoir:
Glory be! It is merged!

PR #1608, opened 2018-02-27
Refactor logic from SierraTransformableTransformer onto SierraItems

alexwlchan:
This is a pure refactoring change in anticipation of #1466. It should have no changes in functionality; it just moves some logic to a new SierraItems trait.

### DIFF
--- a/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraItems.scala
+++ b/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraItems.scala
@@ -1,3 +1,27 @@
 package uk.ac.wellcome.transformer.transformers.sierra
 
-trait SierraItems { }
+import com.twitter.inject.Logging
+import scala.util.{Failure, Success}
+import uk.ac.wellcome.models.transformable.SierraTransformable
+import uk.ac.wellcome.transformer.source.SierraItemData
+import uk.ac.wellcome.utils.JsonUtil._
+
+trait SierraItems extends Logging {
+  def extractItemData(sierraTransformable: SierraTransformable): List[SierraItemData] = {
+    sierraTransformable
+      .itemData
+      .values
+      .map { _.data }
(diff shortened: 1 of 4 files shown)


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraItems.scala
```diff
@@ -0,0 +1,55 @@
...
+  SourceIdentifier,
+  UnidentifiedItem
+}
+import uk.ac.wellcome.models.transformable.SierraTransformable
+import uk.ac.wellcome.transformer.source.SierraItemData
+import uk.ac.wellcome.utils.JsonUtil._
+
+trait SierraItems extends Logging with SierraLocation {
+  def extractItemData(
+    sierraTransformable: SierraTransformable): List[SierraItemData] = {
```
kenoir: I might put some of this logic on a companion object for `SierraItemData` in an `apply` method.

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraItems.scala
```diff
@@ -0,0 +1,55 @@
...
+import uk.ac.wellcome.models.{
+  IdentifierSchemes,
+  SourceIdentifier,
+  UnidentifiedItem
+}
+import uk.ac.wellcome.models.transformable.SierraTransformable
+import uk.ac.wellcome.transformer.source.SierraItemData
+import uk.ac.wellcome.utils.JsonUtil._
+
+trait SierraItems extends Logging with SierraLocation {
```
kenoir: Does this need to be a trait? Could it simply be an object?

alexwlchan: It’s a trait because that’s the current pattern used by SierraTransformableTransformer. They could all be objects instead.

PR #1609, opened 2018-02-27
Remove deleted items from works in the transformer

### LINKED ISSUE #1456: Make sure we don't show invisible items in the API

alexwlchan:
Per discussion yesterday, when the transformer sees an Item with deleted=true, rather than passing it along with visibility=false, it removes it entirely from the Work.

Resolves #1456.

### DIFF
--- a/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraItems.scala
+++ b/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraItems.scala
@@ -29,31 +29,27 @@ trait SierraItems extends Logging with SierraLocation {
       .flatten
   }
 
-  def transformItemData(sierraItemData: SierraItemData): Option[UnidentifiedItem] = {
+  def transformItemData(sierraItemData: SierraItemData): UnidentifiedItem = {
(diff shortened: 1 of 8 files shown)


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraItems.scala
```diff
@@ -29,27 +29,31 @@ trait SierraItems extends Logging with SierraLocation {
...
+          )
+        ),
+        locations = getLocation(sierraItemData).toList
+      ))
+    }
   }
 
   def getItems(
     sierraTransformable: SierraTransformable): List[UnidentifiedItem] = {
     extractItemData(sierraTransformable)
```
kenoir: What if we `.filter` on `deleted` here?

alexwlchan: Yeah, actually, that's simpler, isn't it?

kenoir: 👍

alexwlchan:
Closing and reopening to try to trigger Travis.

Issue #1606, opened 2018-02-26
Mutability of cancellable in TryBackoff could be source of bugs


rgladwell:
Mutability of the `maybeCancellable` private field in `TryBackoff` could be source of bugs:

https://github.com/wellcometrust/platform/blob/fc0566b7f4b466db76b703b9991945dad1805e96/common/src/main/scala/uk/ac/wellcome/utils/TryBackoff.scala#L56

For instance, if the `run` function is called more than once on the same instance of `TryBackoff` the previous `maybeCancellable` will be overwritten and the previously started scheduler will become uncancellable.

This maybe related to issues I'm having running the tests from within a SBT REPL session.

rgladwell:
@kenoir Taking a look at the code, it looks as though a temporary fix would look something like this:

```scala
      if (shouldReschedule) {
        val waitTime = timeToWaitOnAttempt(attempt)
        val cancellable = system.scheduler.scheduleOnce(waitTime milliseconds)(
          run(f, system, attempt = numberOfAttempts))
        cancelRun()
        maybeCancellable = Some(cancellable)
      }
```

Although this above fix maybe susceptible to threading errors, for examnple if two threads call `run` at the same time and `maybeCancellable` is overwritten by one thread before it can be cancelled by the other.

We're actually starting multiple schedulers when we use `TryBackoff` as it recursively calls its own `run` method, so it lookas as if we might be overwriting `maybeCancellable` all the time without cancelling it.

kenoir:
> So it look as as if we might be overwriting maybeCancellable all the time without cancelling it.

Eep!

I was thinking something like:

```scala
    if(!maybeCancellable.isEmpty) {
      throw new RuntimeException(
        "Cannot run, process already running."
      )
    }
```

In the `run` method might highlight the issue further.

rgladwell:
So after reviewing the `TryBackoff` code with @kenoir it looks as though the the actor/thread reading messages off the SQS queue is not shutting down even after `cancelRun`, **if a `GracefulFailureException` is thrown.**

After discussion, writing your own retry/backoff code is really hard and we are consdering removing the custom code for this from `platform` and replacing it with a more established library like [retry](https://github.com/softwaremill/retry) or [scala-retry](https://github.com/hipjim/scala-retry).

Before we proceed we wanted @alexwlchan and @alicefuzier's input, please? Thoughts?

alexwlchan:
[...]

PR #1612, opened 2018-02-27
Remove TryBackoff

rgladwell:
https://github.com/wellcometrust/platform/issues/1606

Simplify the code and remove a source of bugs by deleting `TryBackoff` and rely on the SQS queue to handle retries.

### Who is this change for?

🧐

- [ ] Deployed new versions


common/src/main/scala/uk/ac/wellcome/sqs/SQSWorker.scala
```diff
@@ -26,22 +26,20 @@ abstract class SQSWorker(sqsReader: SQSReader,
...
-        .fromTry(extractMessage(message))
-        .flatMap(m => {
-          debug(s"Processing message: $m")
-
-          metricsSender.timeAndCount(
-            s"${workerName}_ProcessMessage",
-            () => processMessage(m)
-          )
-        })
+      for {
```
kenoir: 🎉

common/src/main/scala/uk/ac/wellcome/sqs/SQSReader.scala
```diff
@@ -24,75 +24,77 @@ class SQSReader @Inject()(sqsClient: AmazonSQS, sqsConfig: SQSConfig)
...
-      if (messages.nonEmpty)
-        info(
-          s"Received messages ${messages.map { _.getMessageId }} from queue ${sqsConfig.queueUrl}")
-      else
-        debug(
-          s"Received messages ${messages.map { _.getMessageId }} from queue ${sqsConfig.queueUrl}")
-      processAndDeleteMessages(messages, process).map { _ =>
+    process: Message => Future[Unit]): Future[Unit] = {
+    val eventuallyProcessedMessages =
+      for {
```
kenoir: 🎉

alicefuzier:
So much nicer with all of these for comprehensions! ➿

catalogue_pipeline/recorder/src/main/scala/uk/ac/wellcome/platform/recorder/modules/RecorderModule.scala
```diff
@@ -3,27 +3,16 @@ package uk.ac.wellcome.platform.recorder.modules
 import akka.actor.ActorSystem
 import com.twitter.inject.{Injector, TwitterModule}
 import uk.ac.wellcome.platform.recorder.services.RecorderWorkerService
-import uk.ac.wellcome.utils.TryBackoff
 
-object RecorderModule extends TwitterModule with TryBackoff {
-
-  override lazy val continuous: Boolean = false
-
-  override def singletonStartup(injector: Injector) {
```
kenoir: Why is this no longer needed?

kenoir: Ignore this ^ I just figured it out.

common/src/main/scala/uk/ac/wellcome/sqs/SQSReader.scala
```diff
@@ -24,75 +24,77 @@ class SQSReader @Inject()(sqsClient: AmazonSQS, sqsConfig: SQSConfig)
...
+        _ <- Future.sequence { ms.map { processAndDeleteMessage(_, process) } }
+      } yield {
+        val messageIds = ms.map { _.getMessageId }
+
+        if (ms.nonEmpty) {
+          info(
+            s"Received messages $messageIds from queue ${sqsConfig.queueUrl}")
+        } else {
+          debug(
+            s"Received messages $messageIds from queue ${sqsConfig.queueUrl}")
```
alexwlchan: Would it be slightly cheaper to do

```scala
debug(s"Received no messages from queue ${sqsConfig.queueUrl}")
```

Save us the cost of a string interpolation, and makes it stand out more from the previous message.

common/src/test/scala/uk/ac/wellcome/sqs/SQSReaderTest.scala
```diff
@@ -16,139 +17,149 @@ class SQSReaderTest
...
     with Matchers
     with ScalaFutures
-    with SQSLocal {
+    with Eventually
+    with SqsFixtures {
 
-  val queueUrl = createQueueAndReturnUrl("test_queue")
-  // Setting 1 second timeout for tests, so that test don't have to wait too long to test message deletion
-  sqsClient.setQueueAttributes(queueUrl, Map("VisibilityTimeout" -> "1"))
+  override def withFixture(test: OneArgTest) = withLocalSqsQueue { queueUrl =>
```
kenoir: Can you talk us through the advantages of this pattern over the previous method of creating test resources?

rgladwell: Basically, the `with*` fixture methods allow us to safely create and destroy stateful resources like SQS queues without relying on thread-unsafe techniques like using mutable variables. It also allows us to create resources with randomized names to improve test isolation. 

Lots of the tests create stateful stuff like SQS queues without tearing them down afterwards.

rgladwell: You can read more about the Scalatest "official" techinques for sharing fixtures here:

http://www.scalatest.org/user_guide/sharing_fixtures

kenoir: Good enough for me.

common/src/test/scala/uk/ac/wellcome/sqs/SQSWorkerToDynamoTest.scala
```diff
@@ -74,6 +60,7 @@ class SQSWorkerToDynamoTest
     }
 
     override def terminalFailureHook(): Unit = {
+      println(s"$this.terminalFailureHook()")
```
kenoir: Get rid of this, Ricardo

common/src/main/scala/uk/ac/wellcome/sqs/SQSWorker.scala
```diff
@@ -1,47 +1,45 @@
...
+        m <- Future.fromTry { fromJson[SQSMessage](message.getBody) }
+        _ <- Future.successful { debug(s"Processing message: $m") }
+        metricName = s"${workerName}_ProcessMessage"
+        _ <- metricsSender.timeAndCount(metricName, () => processMessage(m))
+      } yield ()
+    } recover {
+      case _: Throwable => terminalFailureHook()
```
alicefuzier: Is this correct? If, say Json parsing of the message fails, it should throw a `GracefulFailureException` which is not a terminal failure. I know it will only affect alarms we get and it's not an actual terminal failure (so maybe we should also rename that method)

common/src/test/scala/uk/ac/wellcome/sqs/SQSReaderTest.scala
alicefuzier: Oh really nice! So much better than before

common/src/main/scala/uk/ac/wellcome/sqs/SQSReader.scala
```diff
@@ -24,75 +24,77 @@ class SQSReader @Inject()(sqsClient: AmazonSQS, sqsConfig: SQSConfig)
...
-        debug(
-          s"Received messages ${messages.map { _.getMessageId }} from queue ${sqsConfig.queueUrl}")
-      processAndDeleteMessages(messages, process).map { _ =>
+    process: Message => Future[Unit]): Future[Unit] = {
+    val eventuallyProcessMessages =
+      for {
+        _ <- Future.successful {
+          debug(s"Looking for new messages at ${sqsConfig.queueUrl}")
+        }
+        ms <- receiveMessages()
```
alicefuzier: Not really important but I think I preferred the old variable name `messages`

rgladwell: I agree, but I think in this case I used a shorter name because otherwise the proceeding code breaks the 80 character limit and the re-formatter makes it look ugly. Perhaps we could shorten the name of the `processAndDeleteMessage` method instead?

common/src/main/scala/uk/ac/wellcome/sqs/SQSWorker.scala
rgladwell: I think the `GracefulFailureException` is handled by the [`SQSReader`](https://github.com/wellcometrust/platform/blob/1d79970cf8084b926b5165405b13e4ba4f4f1781/common/src/main/scala/uk/ac/wellcome/sqs/SQSReader.scala#L72). I think by the time execution to this code we no longer need to worry about it.

Is this correct?

common/src/main/scala/uk/ac/wellcome/sqs/SQSReader.scala
alicefuzier: What if we made the 80 character limit be a 100 character limit instead?

common/src/main/scala/uk/ac/wellcome/sqs/SQSWorker.scala
alicefuzier: oh right, yep you're correct

common/src/main/scala/uk/ac/wellcome/sqs/SQSReader.scala
rgladwell: +1

kenoir: pffft yeah, damn the 80 characters if it means we get better variable names.

PR #1613, opened 2018-02-27
[WIP] Write tests for the recorder

alexwlchan:
Part of #1565. I realise I’m blocked on #1604, because what tests to write depends on exactly how the new VHS works.


kenoir:
Closing for now ... I'm going to link this from the merger ticket so we can pull the work back when we get there.

PR #1615, opened 2018-02-27
Run with warnings as errors

alexwlchan:
As a way to enforce good code quality and avoid bugs the compiler could tell us about.

This is at least partially experimental, I want to see how many warnings we currently have.

### DIFF
--- a/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/TransformableTransformer.scala
+++ b/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/TransformableTransformer.scala
@@ -3,6 +3,7 @@ package uk.ac.wellcome.transformer.transformers
 import uk.ac.wellcome.models.UnidentifiedWork
 import uk.ac.wellcome.models.transformable.Transformable
 
+import scala.reflect.ClassTag
 import scala.util.Try
 
 trait TransformableTransformer[+T <: Transformable] {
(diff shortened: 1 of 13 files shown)


common/src/test/scala/uk/ac/wellcome/utils/TryBackoffTest.scala
```diff
@@ -64,17 +64,17 @@ class TryBackoffTest
     f.calls.length shouldBe finalLength
   }
 
-  it(
```
rgladwell: Probably better to mark this test as [quarantined](http://www.scalatest.org/user_guide/tagging_your_tests) rather than comment it out.

rgladwell:
I think I'd be more in favor of this change if it just applied to Travis/CI.

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/TransformableTransformer.scala
```diff
@@ -3,15 +3,16 @@ package uk.ac.wellcome.transformer.transformers
...
-trait TransformableTransformer[+T <: Transformable] {
+trait TransformableTransformer[T <: Transformable] {
   protected[this] def transformForType(
     t: T,
     version: Int): Try[Option[UnidentifiedWork]]
 
-  def transform(transformable: Transformable,
-                version: Int): Try[Option[UnidentifiedWork]] =
+  def transform(transformable: Transformable, version: Int)(
+    implicit tag: ClassTag[T]): Try[Option[UnidentifiedWork]] =
```
alicefuzier: I'd really avoid reflection stuff here. I have a fix for this using partial functions that I have meant to do for a while

PR #1616, opened 2018-02-28
Fix for the ECS dashboard

kenoir:

Visibility of the ECS dashboard.

### Who is this change for?

Those interested in visibility of the infrastructure state.

- [X] Deployed new versions

### DIFF
--- a/monitoring/ecs_dashboard/Makefile
+++ b/monitoring/ecs_dashboard/Makefile
@@ -16,12 +16,15 @@ $(val $(call stack_setup))
 
 ECS_DASHBOARD = $(ROOT)/monitoring/ecs_dashboard
 
-build_deployment_container:
+$(ROOT)/.docker/ecs_dashboard:
 	docker build \
 		-f client/Dockerfile client/ \
 		--tag ecs_dashboard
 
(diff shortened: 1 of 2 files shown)


monitoring/ecs_dashboard/Makefile
```diff
@@ -16,15 +16,14 @@ $(val $(call stack_setup))
...
-	$(ROOT)/builds/build_ci_docker_image.py \
-		--project=ecs_dashboard \
-		--dir=$(ECS_DASHBOARD)/client \
-		--file=$(ECS_DASHBOARD)/client/Dockerfile
+build_deployment_container:
+	docker build \
+		-f client/Dockerfile client/ \
+		--tag ecs_dashboard
 
-ecs_dashboard_client-publish: $(ROOT)/.docker/ecs_dashboard
```
alexwlchan: We use this dotfile as the target so we only have to build it once. After the target completes, you run `mkdir -p $(ROOT)/.docker; touch $(ROOT)/.docker/ecs_dashboard`.

It's not massively important because we don't run this very often, but might be worth changing.

PR #1624, opened 2018-03-01
Don't have a "visible" field on DisplayWork

### LINKED ISSUE #1619: Remove the "visible" field from DisplayWork
If something gets as far as being a DisplayWork, it surely must be visible!

alexwlchan:
This refactors WorkService to deal exclusively in the sort of Work that lives internal to Elasticsearch, and pushes out the display logic.

Resolves #1619.

### DIFF
--- a/catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
+++ b/catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
@@ -114,9 +114,12 @@ class WorksController @Inject()(@Flag("api.prefix") apiPrefix: String,
       works
         .map { resultList =>
           DisplayResultList(
-            results = resultList.results.map { DisplayWork(_, includes) }.toArray,
+            results =
+              resultList.results.map { DisplayWork(_, includes) }.toArray,
             pageSize = pageSize,
-            totalPages = Math.ceil(resultList.totalResults.toDouble / pageSize.toDouble).toInt,
+            totalPages = Math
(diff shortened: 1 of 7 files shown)


catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
```diff
@@ -93,26 +93,36 @@ class WorksController @Inject()(@Flag("api.prefix") apiPrefix: String,
...
             pageSize = pageSize,
             pageNumber = request.page,
-            includes = request.includes.getOrElse(WorksIncludes()),
             index = request._index
           )
       }
 
       works
+        .map { resultList =>
+          DisplayResultList(
```
kenoir: It might be worth moving some of the work done here to an apply method on `DisplayResultList` to make this less verbose.

catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
```diff
@@ -93,26 +93,36 @@ class WorksController @Inject()(@Flag("api.prefix") apiPrefix: String,
...
           worksService.listWorks(
             pageSize = pageSize,
             pageNumber = request.page,
-            includes = request.includes.getOrElse(WorksIncludes()),
             index = request._index
           )
       }
 
       works
+        .map { resultList =>
```
kenoir: You could probably end up using a `for` comprehension here.

PR #1629, opened 2018-03-02
Add physicalDescription to Works

alexwlchan:
Implement another transformation!

Transformation checklist:

- [x] Added any new fields to Work/Item (and the unidentified/identified variants)
- [x] Update the JSON serialisation tests in [Un]IdentifiedWorkTest
- [x] Implemented the new transformation
- [x] Added any new fields to the Elasticsearch mapping
- [x] Updated the Display models in the API

Related: #1419


alexwlchan:
Pausing because it's suddenly struck me that I'm not sure if this lives on Works or Items.

alexwlchan:
See Slack discussion starting here: https://wellcome.slack.com/archives/C8X9YKM5X/p1519989270000444

The data is only available on bib-level records, so we're putting it on Work for now. Silver's going to discuss it with the librarians next week.

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/MarcUtils.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import uk.ac.wellcome.transformer.source.{MarcSubfield, SierraBibData}
+
+trait MarcUtils {
+  def getMatchingSubfields(
+    bibData: SierraBibData,
+    marcTag: String,
+    marcSubfieldTags: List[String]): List[MarcSubfield] = {
+    bibData.varFields
+      .filter { _.marcTag == Some(marcTag) }
```
kenoir: You can `collect` whereever you `filter` and `map`.

PR #1633, opened 2018-03-05
Various improvements to our Loris image

### LINKED ISSUE #924: Thread of unexplained Loris 500 errors
This is a ticket for tracking any 500s we’ve seen from Loris that we haven’t fixed or explained.

We can split them off into their own tickets if/when we start working on them – for now, this is a place to dump errors that are transient/one-off.

### LINKED ISSUE #1627: Upgrade our Loris container to use Python 3
One of the new features in Loris 2.3.0 is support for Python 3.

If we jump straight to Python 3.6, I'd expect to see a moderate improvement in CPU usage and memory utilisation (dicts are more memory-efficient in 3.6 than 2.7, and there are lots of dicts in Loris).

### LINKED ISSUE #1628: Stop using a patched version of Loris
Pun intended.

Currently we use a version of Loris with a custom patch. Specifically, we're patching the TemplateHTTPResolver to support tenacity in requests from S3.

It would be better if we wrote our own resolver, then loaded that directly into the container – so we could use the original Loris source code without modifications.

This would be a good opportunity to learn more about how Loris works.

alexwlchan:
This patch:

* Upgrades us to the latest version of Loris, plus a few extra fixes (see https://github.com/loris-imageserver/loris/pull/410)

* Upgrades Loris to use Python 3, which should bring a tidy perf improvement (resolves #1627)

* Moves our customisations into this repo, rather than squirrelling them away in a fork (resolves #1628, resolves #924).

  This includes the tenacity patch for fetching from S3, which was the source of all our flakey S3 errors. We still have some flakiness in the ALB, but that's not appearing in Loris logs. Now we copy that file into our Docker image, which:

  - Isolates any Wellcome-specific code from core Loris
  - Opens the door for easy customisation later (e.g. loading some images from boto3, others over HTTP)
  - Makes it easy to see exactly what we're doing that's unusual

  In particular, I'm hoping this makes further customisations to our image more accessible to other devs.

### DIFF
--- a/loris/loris/Dockerfile
+++ b/loris/loris/Dockerfile
@@ -1,15 +1,28 @@
 FROM ubuntu:16.04
 
+
+# Install base Python and uWSGI (web server) dependencies
 RUN apt-get update && \
-    apt-get install --yes python uwsgi uwsgi-plugin-python && \
+    apt-get install --yes python3 python3-pip uwsgi uwsgi-plugin-python3 && \
     apt-get clean
 
-ENV LORIS_COMMIT a349d24b63bf1ae1af97e376304c4b1acc983639
+RUN pip3 install --upgrade pip
+
+# Imaging dependencies.  These libraries don't change much between versions
+# of Loris, so we can cache their installation.
+RUN apt-get update && \
+    apt-get install --yes libffi-dev libjpeg-turbo8-dev libfreetype6-dev zlib1g-dev \
+        liblcms2-dev liblcms2-utils libssl-dev libtiff5-dev libwebp-dev && \
+    apt-get clean
+
+
 ENV LORIS_GITHUB_USER loris-imageserver
+ENV LORIS_COMMIT 841383810cdd62ae591f532fd89330bb0be06565
 
 COPY requirements.txt /
 COPY install_loris.sh /install_loris.sh
 RUN /install_loris.sh
+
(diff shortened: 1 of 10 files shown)


loris/loris/Dockerfile
```diff
@@ -1,16 +1,29 @@
 FROM ubuntu:16.04
```
kenoir: You could perhaps build this from https://github.com/wellcometrust/dockerfiles/tree/master/python3?

loris/loris/wellcome_loris.py
```diff
@@ -0,0 +1,29 @@
...
+an upstream patch.
+
+"""
+
+from loris.resolvers import TemplateHTTPResolver
+from requests.exceptions import RequestException
+from tenacity import retry, retry_if_exception_type, stop_after_attempt
+
+
+class WellcomeTemplateHTTPResolver(TemplateHTTPResolver):
```
kenoir: 👍

loris/loris/Dockerfile
alexwlchan: We ran Loris on Alpine originally, and saw a small improvement in response times when we switched to Ubuntu (I'm not sure if it was statistically significant tho!).

Might be worth another try at some point, with proper testing.

Issue #1639, opened 2018-03-07
Migrate the internal Sierra IDs to the correct scheme


alexwlchan:
- [x] Modify the Sierra readers to write records with the 7-digit, no-prefix, no check-digit scheme (#1640)
- [x] Move all the rows in the SierraItems DynamoDB table to the 7-digit, no-prefix, no check digit scheme (#1642).
- [ ] Move all the rows in the SourceData DynamoDB table to the 7-digit, no-prefix, no check digit scheme (#1642, #1645)
- [ ] Move all the keys in the S3 bucket to match (#1645)
- [ ] Fix the contents of everything in the S3 bucket (#1645)
- [x] Modify the Sierra transformer to use two versions of Sierra IDs:
     - [x] `sierra-identifier` which is the raw id
     - [x] `sierra-system-number` which has the record type and check digit
- [x] Modify the Miro transformer to use two versions of Sierra IDs:
     - [ ] ~`sierra-identifier` which is the raw id~
     - [ ] `sierra-system-number` which has the record type and check digit

jtweed:
Probably don't need `sierra-identifier` on Miro works.

alexwlchan:
So, change of plan. We're going to go back to Sierra and reread everything, rather than trying to do any more database surgery.

pollecuttn:
Can whoever goes back to Sierra let Louise Simon know when it's happening so she can be aware if there's any knock on effects to Sierra.

alexwlchan:
I didn't manage to clear all the Sierra IDs from the SourceData table or the VHS buckets, so somebody else will need to do that first.

alexwlchan:
Now we’ve reharvested all the Sierra data, I think this is done, modulo reindexes.

PR #1642, opened 2018-03-07
migrate_dynamodb_items.py: add script for migrating the SierraData_items table

alexwlchan:
Part 2 of #1639.

Transforms are tested; DynamoDB logic is not. Should be fairly fast, as there are only 850k records in the table.

### DIFF
--- a/migrate_dynamodb_items.py
+++ b/migrate_dynamodb_items.py
@@ -1,0 +1,53 @@
+#!/usr/bin/env python
+# -*- encoding: utf-8 -*-
+
+import json
+import re
+
+import boto3
+
+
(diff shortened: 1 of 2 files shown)


migrate_dynamodb_items.py
```diff
@@ -0,0 +1,53 @@
...
+
+    return item
+
+
+for item in items():
+    old_item = item.copy()
+    new_item = transform_item(item)
+    if old_item == new_item:
+        continue
+    table.put_item(Item=new_item)
```
alexwlchan: Note: we can get away with a regular PUT rather than a conditional update because these keys don't exist yet!

PR #1645, opened 2018-03-08
Script for migrating the SourceData table

alexwlchan:
Another chunk of #1639.

You can see the two test cases (with and without itemData) in the table:

* sierra/b1919592
* sierra/b2097560

I've created the new keys, but not deleted the old keys yet – when I run the script proper, they'll gradually be cleaned up.


kenoir:
Closing as not required.

Issue #1397, opened 2018-01-11
Scheduled snapshots from Elasticsearch to S3


jtweed:
We also need to make sure we suppress visible records using the same query as the API.

Depends:
- #1668 
- #1663
- #1659
- #1661
- #1660

alexwlchan:
The current code is here: https://github.com/wellcometrust/dockerfiles/tree/master/elasticdump

It was meant to run as a standalone task; we'll probably want to change it to be passed config directly if it's being run on a schedule.

jtweed:
When someone picks this up, I'd like to pair on it to learn a bit a how Terraform and ECS services work.

jtweed:
I would like us to set an approach here that we can use for all downloadable datasets, not just the Catalogue. We already have examples of other datasets we would like to make available both as an API and a downloadable dataset, eg. the Medical Officer of Health reports.

My proposal is therefore that we create a new subdomain for all downloadable datasets, that mirrors the structure for public APIs. The URL pattern should be:

`https://data.wellcomecollection.org/{api}/{version}/{type}.{format}`

This would be mapped to a public S3 bucket called `s3://wellcomecollection-public-data`.

If we use this to make data dumps available that don't have an associated API, the pattern would be:

`https://data.wellcomecollection.org/{dataset}/{type}.{format}`

Files should be gzipped, so the Catalogue API works would be available at:

`https://data.wellcomecollection.org/catalogue/v1/works.json.gz`

Outstanding questions:

- Do we need to date version the dumps? eg. `/catalogue/20180306-works.json.gz` or `/catalogue/20180306/works.json.gz`
- If so, should we always additionally provide the latest without a date in the filename?
- Is it weird using the API version like this? It provides a contract about the format of the data.

kenoir:
The API version is useful as it implies major format changes. 

We need to decide if we are happy to make date snapshots available, as we are likely to want to remove things from old snapshots (based on say author requests) and I don't think we want to deal with having to redact old snapshots.

alexwlchan:
> Do we need to date version the dumps? eg. /catalogue/20180306-works.json.gz or /catalogue/20180306/works.json.gz

I think so – and then dated data sets are immutable. If you download the 20180306 data set, you always get either:

* A 200 OK and the same data you would have got the previous day; or
* A 410 Gone if we've had to redact the data

Since the API doesn't make the guarantee of immutability, and it feels like that's one of the features/reasons we'd have the snapshots.

[...]

PR #1647, opened 2018-03-08
Add a data_api stack

kenoir:

Adds a new terraform stack for the mooted `data.wellcomecollection.org` services.

Part of: 

- https://github.com/wellcometrust/platform/issues/1397

### Who is this change for?

Folk who want to access (and make available) large datasets.

- [ ] Reviewed by @jtweed
- [ ] Deployed new versions
- [ ] Run `terraform apply`.

### DIFF
--- a/data_api/main.tf
+++ b/data_api/main.tf
@@ -1,0 +1,11 @@
+module "elasticdump" {
+  source        = "git::https://github.com/wellcometrust/terraform.git//ecs_script_task?ref=v1.0.0"
+  task_name     = "elasticdump"
+  app_uri       = "wellcome/elasticdump:latest"
+  task_role_arn = "${module.ecs_elasticdump_iam.task_role_arn}"
+
+  env_vars = [
+    "{\"name\": \"BUCKET\", \"value\": \"${var.elasticdump_config_bucket}\"}",
+    "{\"name\": \"S3_KEY\", \"value\": \"${var.elasticdump_config_s3_key}\"}",
+  ]
+}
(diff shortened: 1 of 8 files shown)


data_api/main.tf
```diff
@@ -0,0 +1,11 @@
+module "elasticdump" {
+  source        = "git::https://github.com/wellcometrust/terraform.git//ecs_script_task?ref=v1.0.0"
+  task_name     = "gatling_loris"
```
alexwlchan: ???

data_api/main.tf
```diff
@@ -0,0 +1,11 @@
+module "elasticdump" {
+  source        = "git::https://github.com/wellcometrust/terraform.git//ecs_script_task?ref=v1.0.0"
+  task_name     = "elasticdump"
+  app_uri       = "wellcome/elasticdump:latest"
+  task_role_arn = "${module.ecs_elasticdump_iam.task_role_arn}"
+
+  env_vars = [
+    "{\"name\": \"BUCKET\", \"value\": \"${var.elasticdump_config_bucket}\"}",
```
alexwlchan: Ick. We should make the module not be this way!

alexwlchan:
We should teach Travis that anything in the data_api folder should be ignored for non-data tasks.

data_api/main.tf
kenoir: I'm not sure it needs to at a more current version. 

This is really just a placeholder to add _some script task_ in the future.

kenoir:
@jtweed you mentioned you wanted to work on deploying a new service. This might be a good opportunity to do so.

PR #1648, opened 2018-03-08
Transform place of publication

alicefuzier:
✂️


catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/transformers/SierraTransformableTransformerTest.scala
```diff
@@ -432,4 +432,44 @@ class SierraTransformableTransformerTest
...
+    val sierraTransformable = SierraTransformable(
+      sourceId = id,
+      maybeBibData =
+        Some(SierraBibRecord(id = id, data = data, modifiedDate = now())))
+
+    val transformedSierraRecord =
+      transformer.transform(sierraTransformable, version = 1)
+    transformedSierraRecord.isSuccess shouldBe true
+
+    transformedSierraRecord.get.get.placesOfPublication shouldBe List(
```
rgladwell: Might be clearer to assert something like `should contain only` here.

PR #1652, opened 2018-03-08
sierra_ids.md: add some notes on Sierra IDs, and how we store them

### LINKED ISSUE #1651: Do we need to support searching on variants of the Sierra system number?
The proposal in #1639 is that we store and present two forms of the Sierra ID:

* Without a prefix or check digit, e.g. `1234567`
* With a prefix and check digit, e.g. `b1234567x`

The Sierra manual says:

> When searching for a record number, if you don't know the check digit, you can substitute the
> character **a** for the check digit. For example, you could enter [.b1024364] as .b1024364a.

Elasticsearch won't know about this variant of the ID. Should we include it as an invisible identifier – indexed for search, but not visible to the user? Is this wildcard ever exposed to users? Is it supported in the current library search?

alexwlchan:
Notes some of the discussion from yesterday and today.

Resolves #1651.

### DIFF
--- a/docs/sierra_ids.md
+++ b/docs/sierra_ids.md
@@ -1,0 +1,115 @@
+# Sierra IDs
+
+The Sierra IDs returned by the Sierra API are seven digits long,
+for example `1234567`.
+
+However, there are several optional features of a Sierra ID:
+
+*   A leading period (`.`)
+
+*   A one-character tag to identify the type -- for example, `b` for a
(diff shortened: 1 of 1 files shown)


alexwlchan:
Probably best reviewed in the Markdown preview: https://github.com/wellcometrust/platform/blob/sierra-ids-docs/docs/sierra_ids.md

PR #1657, opened 2018-03-09
Make the metrics flushInterval configurable

alicefuzier:
We seem to be still spending ~10£ a day on cloudwatch which means ~300£ a month and it's a bit more than I expected to spend with the MetricSender changes we did last month. 
The way `MetricSender` works is that it collects metrics to be sent to cloudwatch until they reach the maximum number metric that we can send in one PutMetric request to AWS. These metrics are collected in an in memory buffer in windows of 10 seconds, so if 10 seconds pass without having reached the maximum number of metrics, `MetricsSender` will send a PutMetric data request with fewer metrics.
I think 10 seconds might be a bit too short and causing a lot of these requests to not use the maximum number of metrics, so this is to make the interval configurable with the default value 5 minutes instead of 10 seconds

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/finatra/modules/AmazonCloudWatchModule.scala
+++ b/common/src/main/scala/uk/ac/wellcome/finatra/modules/AmazonCloudWatchModule.scala
@@ -1,31 +1,31 @@
 package uk.ac.wellcome.finatra.modules
 
-import akka.actor.ActorSystem
 import com.amazonaws.client.builder.AwsClientBuilder.EndpointConfiguration
-import com.amazonaws.services.cloudwatch.{
-  AmazonCloudWatch,
-  AmazonCloudWatchClientBuilder
-}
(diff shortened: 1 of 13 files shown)


common/src/main/scala/uk/ac/wellcome/finatra/modules/AmazonCloudWatchModule.scala
```diff
@@ -25,7 +39,11 @@ object AmazonCloudWatchModule extends TwitterModule {
   @Singleton
   def providesMetricsSender(amazonCloudWatch: AmazonCloudWatch,
                             actorSystem: ActorSystem) =
-    new MetricsSender(awsNamespace(), amazonCloudWatch, actorSystem)
+    new MetricsSender(
+      awsNamespace(),
+      flushInterval(),
+      amazonCloudWatch,
+      actorSystem)
```
alexwlchan: Could we get some keyword arguments on this for clarity?

PR #1661, opened 2018-03-12
Move the Display models to the sbt-common lib

alexwlchan:
~Some refactoring work for #1397. We’ll need these models for the task that converts raw Elasticsearch results into our Work type, so let's get them into the common lib!~

This PR has expanded somewhat. Now:

* It pulls the Display models out of the API and into an sbt-display library
* Sets up the new library with tooling in Travis and Make
* Establishes a pattern for further breaking up the common lib (which we'll want to keep class paths tidy)


common/src/main/scala/uk/ac/wellcome/models/display/DisplayAgent.scala
```diff
@@ -1,4 +1,4 @@
-package uk.ac.wellcome.platform.api.models
+package uk.ac.wellcome.models.display
```
kenoir: 👍 good use of namespacing.

project/Dependencies.scala
```diff
@@ -83,15 +87,13 @@ object Dependencies {
     "com.novocode" % "junit-interface" % versions.junitInterface % "test",
     "org.scalacheck" %% "scalacheck" % versions.scalaCheckVersion % "test",
     "com.github.alexarchambault" %% "scalacheck-shapeless_1.13" % versions.scalaCheckShapelessVersion % "test"
-  ) ++ esDependencies ++ awsDependencies ++ akkaDependencies ++ dynamoDependencies ++ jacksonDependencies ++ circeDependencies
+  ) ++ esDependencies ++ awsDependencies ++ akkaDependencies ++ dynamoDependencies ++ jacksonDependencies ++ circeDependencies ++ swaggerDependencies
```
kenoir: Might be an idea to split out display into it's own package.

common/src/main/scala/uk/ac/wellcome/models/display/DisplayAgent.scala
kenoir: This also means we can drop `Display` from the name of the classes.

alexwlchan: Having thought about it, I think I’ll keep the Display prefixes. Yes it’s double-namespacing, but there are contexts where it makes a difference – e.g. in the API, where at various points you’re dealing with internal/display Works.

For human-readability, I think I’ll leave it (else I’d change all the imports to `import uk.ac.wellcome.models.display.{Work => DisplayWork}` anyway!)

project/Dependencies.scala
alexwlchan: So I think we should do this; next question, how.

How would you feel about an `sbt_common` directory in the top-level, which contains sub-projects for these separate packages. For example:

* sbt_common/display_models
* sbt_common/elasticsearch
* sbt_common/sierra_models

which serves as a common subdirectory for our sbt projects (and then lets us throw them into our Make ecosystem).

Issue #1643, opened 2018-03-07
Improve test isolation


rgladwell:
A [recent ticket](https://github.com/wellcometrust/platform/pull/1612) revealed that many of our tests aren't properly isolated and share stateful objects and services like mocks and SQS queues.

We need to go through all the tests in each service and:

 1. Replace all occurrences of mutable variables with immutable ones.
 2. In cases where mutable variables are required, ensure interractions are thread-safe.
 3. Ensure instances of mocks are not shared between tests.
 4. Ensure all stateful, persistent resources (like SQS queues) are safely created and destroyed using the [loan-fixture](https://github.com/wellcometrust/platform/blob/2d5bb62f78715ff3ca7767482696f420aec911a5/common/src/test/scala/uk/ac/wellcome/test/fixtures/AkkaFixtures.scala) design pattern

We should go through each Scala sub-module and verify the above:

 - [x] api (#1714)
 - [x] common (#1656)
 - [x] id_minter (#1654)
 - [x] ingestor (#1665)
 - [x] recorder
 - [x] reindex_worker (#1712)
 - [x] sierra_adapter_common
 - [x] sierra_bib_merger (#1764)
 - [x] sierra_item_merger (#1764)
 - [x] sierra_items_to_dynamo (#1690)
 - [x] sierra_reader (#1708)
 - [x] transformer (#1667)

alexwlchan:
The sierra-common tests don’t have any tests which need fixtures. That package does contain the VHSLocal, but we need to remove that in other tests first!

alexwlchan:
The one outstanding thorn here is the API, which has a bunch of test classes that use a `server` instance a la ApiWorksTest.

In many of these tests, we're testing the public-facing behaviour – I don’t think we could do away with the server instance. If we had to start/stop a test every time, the tests could take noticeably longer (and tests would become spammier).

I had a go at this once (#1714), and I didn't manage to finish it – the tests were hanging, but Travis couldn't tell me why.

PR #1667, opened 2018-03-13
Increase Transformer test isolation

kenoir:

Increase the isolation of Transformer tests.

### Who is this change for?

🐷 🐛 

Part of:

https://github.com/wellcometrust/platform/issues/1643


catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/CalmTransformerFeatureTest.scala
```diff
@@ -1,80 +1,75 @@
...
+    withLocalSnsTopic { topicArn =>
+      withLocalSqsQueue { queueUrl =>
+        withLocalS3Bucket { bucketName =>
+          val flags: Map[String, String] = Map(
+            "aws.sqs.queue.url" -> queueUrl,
+            "aws.sns.topic.arn" -> topicArn,
+            "aws.s3.bucketName" -> bucketName,
+            "aws.sqs.waitTime" -> "1",
+            "aws.metrics.namespace" -> "sierra-transformer"
+          ) ++ s3LocalFlags ++ snsLocalFlags ++ sqsLocalFlags
```
alexwlchan: Q: should we perhaps make these functions that take the bucket name, topic ARN, queue URL, etc., and adds those too?

So the type system forces you to add these flags.

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/CalmTransformerFeatureTest.scala
```diff
@@ -1,80 +1,75 @@
...
 
-    eventually {
-      val snsMessages = listMessagesReceivedFromSNS()
-      snsMessages should have size 1
-      assertSNSMessageContainsCalmDataWith(snsMessages.head, Some("public"))
-    }
+          withServer(flags) { _ =>
+            val calmTransformable =
```
alexwlchan: This, and calmHybridRecordMessage, should be moved out of the with block.

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/MiroTransformerFeatureTest.scala
```diff
@@ -1,56 +1,74 @@
...
+            "aws.sqs.waitTime" -> "1",
+            "aws.metrics.namespace" -> "sierra-transformer"
+          ) ++ s3LocalFlags ++ snsLocalFlags ++ sqsLocalFlags
+
+          withServer(flags) { _ =>
+            val miroID = "M0000001"
+            val title = "A guide for a giraffe"
+
+            val secondMiroID = "M0000002"
+            val secondTitle = "A song about a snake"
```
alexwlchan: Move these constants out of the "with" block.

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/SierraTransformerFeatureTest.scala
```diff
@@ -2,73 +2,87 @@ package uk.ac.wellcome.transformer
...
-  it("should transform sierra records, and publish them to the given topic") {
+          withServer(flags) { _ =>
+            val id = "b001"
+            val title = "A pot of possums"
+            val lastModifiedDate = Instant.now()
 
-    val id = "b001"
-    val title = "A pot of possums"
-    val lastModifiedDate = Instant.now()
+            val sierraHybridRecordMessage =
```
alexwlchan: Move this out of the "with" block (I'll stop saying this now, but the general pattern applies).

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/CalmTransformerFeatureTest.scala
alexwlchan: Now implemented as #1670.

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/SierraTransformerFeatureTest.scala
kenoir: Generally why do you think this is a good idea?

I've decided I like it because it gives you the opportunity to define values as soon as possible so that non dependent processes can run concurrently (for example creating buckets/queues and starting the server).

PR #1670, opened 2018-03-14
Minor improvements to our fixtures

alexwlchan:
Rather than defining a set of static flags, then having to remember to add flags for queue URL/bucket name/topic ARN, make the flags a method that takes the dynamic values. This avoids any issues of you forgetting to define those flags (which I’ve done a few times!).

Plus moving a bit of boilerplate code down into the SQS queue fixture.


kenoir:
LGTM, but going to leave this for @rgladwell to review.

common/src/test/scala/uk/ac/wellcome/test/fixtures/SqsFixtures.scala
```diff
@@ -23,11 +24,13 @@ trait SqsFixtures {
   private val accessKey = "access"
   private val secretKey = "secret"
 
-  val sqsLocalFlags = Map(
+  def sqsLocalFlags(queueUrl: String) = Map(
```
rgladwell: +1 much simpler, less code

PR #1689, opened 2018-03-16
Clean up IdentifierSchemes.scala

alexwlchan:
While staring at IdentifierSchemes.scala right now, I spotted some things we could simplify.

* Get rid of the Jackson-related code – Jackson never uses this internal model
* Add some tests for this model
* Simplify the horrible switch statement used for JSON deserialisation
* Reduce the number of places where you need to define a new identifier

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/models/IdentifierSchemes.scala
+++ b/common/src/main/scala/uk/ac/wellcome/models/IdentifierSchemes.scala
@@ -8,15 +8,6 @@ import cats.syntax.either._
   *  the strings that will be presented to users of the API.
   */
 object IdentifierSchemes extends Logging {
-  private final val knownIdentifierSchemes = Seq(
-    miroImageNumber,
-    miroLibraryReference,
-    calmPlaceholder,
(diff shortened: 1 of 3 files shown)


common/src/main/scala/uk/ac/wellcome/models/IdentifierSchemes.scala
```diff
@@ -87,28 +51,25 @@ object IdentifierSchemes extends Logging {
...
-        IdentifierSchemes.calmAltRefNo
-      case s: String if s == IdentifierSchemes.calmPlaceholder.toString =>
-        IdentifierSchemes.calmPlaceholder
-      case s: String if s == IdentifierSchemes.miroLibraryReference.toString =>
-        IdentifierSchemes.miroLibraryReference
-      case s: String if s == IdentifierSchemes.marcCountries.toString =>
-        IdentifierSchemes.marcCountries
-      case identifierScheme =>
+    knownIdentifierSchemes
+      .filter { _.toString == identifierScheme }
```
rgladwell: 👍 much simpler

PR #1691, opened 2018-03-17
Account for at least once message delivery

kenoir:

I've seen this transformer test behave unreliably even though it is isolated. It appears that elasticmq (the container providing FakeSQS) adheres to the same "at least once" delivery guarantee that SQS does. As a consequences messages may be delivered more than once (and this behaviour is _expected_) so the tests need to account for that.

This could also explain reliability issues elsewhere!

### Who is this change for?

🎲 👾 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


alexwlchan:
Rather than fixing this one test (I've seen this issue elsewhere too), could we add another method to our SNS fixture that solves the problem at its core?

`listUniqueMessagesReceivedFromSNS`?

kenoir:
@alexwlchan I think it might be difficult to differentiate between situations where messages are the same because they should be or because of the multiple delivery.

I think I prefer it being explicitly accounted for in the tests.

Issue #1682, opened 2018-03-14
Transform bibs and items, part 2


alexwlchan:
This is a tracking ticket for the new transformations we agreed in backlog refinement today. In priority order (most important first):

* [x] workType (see also: #787, #1688)
* [x] creators (see also: #1724 , #1698 , #1528)
* [x] contributors (#1917, #1920, #1925)
* [x] genres and subjects (#1938)
* [x] lettering (see also: #1673)
* [x] language (#1699)
* [x] dimensions (#1733)

---

* [x] Reindex Miro and Sierra source data into new index.
* [ ] Reindex source data into new indices for the contributor changes

PR #1698, opened 2018-03-19
Extract identifiers for sierra creators

alicefuzier:
that 🔝 . Part of #1682 
### Who is this change for?
⛸ 🏔


catalogue_pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
```diff
@@ -24,23 +24,22 @@ class IdentifiersDao @Inject()(db: DB, identifiers: IdentifiersTable)
    * source identifiers.
    */
   def lookupId(
-    sourceIdentifier: SourceIdentifier,
-    ontologyType: String
+    sourceIdentifier: SourceIdentifier
```
alexwlchan: Just from a reviewing standpoint, it would be easier if the ID minter changes were in a separate PR. :-)

sbt_common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayWork.scala
```diff
@@ -45,7 +45,7 @@ case class DisplayWork(
     dataType = "List[uk.ac.wellcome.display.models.DisplayAgent]",
     value =
       "Relates a work to its author, compiler, editor, artist or other entity responsible for its coming into existence in the form that it has."
-  ) creators: List[DisplayAgent] = List(),
+  ) creators: List[DisplayAbstractAgent] = List(),
```
alexwlchan: Is this naming mismatch going to cause issues in the Swagger?

catalogue_pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
alicefuzier: Not really possible. The id minter changes where needed because we introduced the `Identified` `Identifiable` and `Unidentifiable` wrapper classes. If we didn't do this in the same PR the id minter would crap out for every work with creators or publishers with identifiers

alexwlchan: I've just had the same conversation with Ricardo – model/ID minter changes are fine together, would be nice if we could get them separate from the MARC-parsing creators stuff, but understand that might be tricky.

alicefuzier: Yeah, I think that's a bit tricky at this point

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraCreatorsTest.scala
```diff
@@ -83,9 +84,151 @@ class SierraCreatorsTest extends FunSpec with Matchers {
...
+      IdentifierSchemes.libraryOfCongressNames,
+      "Person",
+      code.trim)
+    creators should contain only Identifiable(
+      Person(label = name),
+      sourceIdentifier = sourceIdentifier,
+      List(sourceIdentifier))
+  }
+
+//  it("extracts the creator identifier removing trailing and leading from marcTag 100 0") {
```
rgladwell: We should comment on why this test is commented out. I would also suggest we quarantine it rather than comment it out using the [ignore tag](http://www.scalatest.org/user_guide/tagging_your_tests).

common/src/main/scala/uk/ac/wellcome/models/IdentityState.scala
```diff
@@ -0,0 +1,23 @@
+package uk.ac.wellcome.models
+
+sealed trait IdentityState[+T]
```
rgladwell: I'm biased but I really like this.

common/src/main/scala/uk/ac/wellcome/models/IdentityState.scala
```diff
@@ -0,0 +1,23 @@
...
+
+sealed trait IdentifieableOrUnidentifiable[+T] extends IdentityState[T]
```
rgladwell: As per our discussion last week, I think this trait should definiately be renamed to something like `Displayable`.

common/src/main/scala/uk/ac/wellcome/models/IdentityState.scala
```diff
@@ -0,0 +1,23 @@
...
+sealed trait IdentifiedOrUnidentifiable[+T] extends IdentityState[T]
```
rgladwell: Maybe this could be `MaybeDisplayable` instead?

sbt_common/display/src/test/scala/uk/ac/wellcome/display/models/DisplayWorkTest.scala
```diff
@@ -88,15 +89,23 @@ class DisplayWorkTest extends FunSpec with Matchers {
         identifiers = Nil,
         canonicalId = "hz2hrba9",
         publishers = List(
-          Agent("Henry Hare"),
-          Agent("Harriet Heron")
+          Unidentifiable(Agent("Henry Hare")),
```
rgladwell: I notice we use `Unidentifiable` in a lot here. We should also test `Indentifiable` and `Indentified` values in this test suite.

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraCreatorsTest.scala
alicefuzier: oops 👍

sbt_common/display/src/test/scala/uk/ac/wellcome/display/models/DisplayWorkTest.scala
alicefuzier: Makes sense. I added a more end to end one to `CreatorsTest` but yeah, We should have one here to.

common/src/main/scala/uk/ac/wellcome/models/IdentityState.scala
alicefuzier: I think you mean the other way round? I feel that I'd prefer a name that somehow mentions identifiers, but I agree that the current names are rubbish and can cause confusion, so I'll go for `Displayable` and `MaybeDisplayable` for now

common/src/main/scala/uk/ac/wellcome/models/IdentityState.scala
alicefuzier: Yep, it's very neat 👏

catalogue_pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/IdEmbedder.scala
```diff
@@ -46,15 +47,21 @@ class IdEmbedder @Inject()(metricsSender: MetricsSender,
...
         .retrieveOrGenerateCanonicalId(
-          identifier = sourceIdentifier,
-          ontologyType = ontologyType
+          identifier = sourceIdentifier
         )
         .get
-      ("canonicalId", Json.fromString(canonicalId)) +: obj
+
+      if (obj.contains("type")) {
+        (("type", Json.fromString("Identified")) +:
```
alicefuzier: This makes me winge a bit but, with a model agnostic id_minter, I don't know how to avoid it

rgladwell: It might help to override the discriminator defaults that Circe uses for sealed traits (class names) so that at least changing a class name won't break our JSON.

alicefuzier: I don't think you can...it uses shapeless under the hood and that's what shapeless gives you in a coproduct

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraCreators.scala
```diff
@@ -1,14 +1,25 @@
...
+  /* Populate wwork:creators. Rules:
+   *
+   * . For all bibliographic records use "marcTag": "100", "110"
+   * . Platform "label" populate from subfield "a"
+   * . For "100" type as "Person" populate "prefix" from 100 subfield c and "numeration" from b
+   * . For "110" type as "Organisation"
+   * . If subfield 0 contains a value use it to populate "identifiers". The "identifierScheme" will be lc-names.
+   * For now we are only removing leading and trailing spaces in identifiers.
+   * TODO: Figure out if we have to normalise further.
+   */
```
alexwlchan: 👍

PR #1711, opened 2018-03-21
Further reduce the autoscaling max on the ID minter

alexwlchan:
Because we're still getting terminal failure warnings!


kenoir:
Good idea, I think if we hit 3 and still see errors then we might have to look at another approach.

PR #1712, opened 2018-03-21
Isolate the reindex worker test

alexwlchan:
This started as another piece of #1643, but the reindex is rather fiddly. So this patch *also*:

* Makes the GSI name a configurable flag to the reindexer (as our fixture generates it randomly)
* Makes the DynamoDB config flags a method on the fixture, so you *have* to supply a table name

WIP because I want to double-check everything before sending it round for review, but it should be passing all tests.


alexwlchan:
Okay, I think I’m happy for somebody else to review this.

PR #1714, opened 2018-03-21
Isolate the API tests

alexwlchan:
Another chunk of #1643.

While doing this, I think it might be nice to make itemType an attribute on the ESFixtures, then we can just use it further down – it would simplify some of this code considerably. But this patch is already large enough!


catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchServiceTest.scala
```diff
@@ -1,121 +1,182 @@
...
-      expectedWorks = List(work3, work2, work1).map { DisplayWork(_) }
-    )
+    testWith(searchService)
+  }
 
-    // TODO: canonicalID is the only user-defined field that we can sort on.
-    // When we have other fields we can sort on, we should extend this test
-    // for different sort orders.
+  it("should sort results from Elasticsearch in the correct order") {
+    withLocalElasticsearchIndex(itemType = itemType) { indexName =>
```
rgladwell: I noticed some test instability using this version of the method, vs the async `withLocalElasticsearchIndexAsync` version. When re-factoring the `common` tests.

alexwlchan: Okay, maybe when I come back and do the itemType stuff, I’ll get rid of all the non-async methods at the same time.

rgladwell: Cheers, Alex.

alexwlchan:
Hmm. The WorkIndexer tests are failing in Travis, but not locally. I smell async shenanigans!

alexwlchan:
Ah, so the spam comes because we start a server for every instances of ApiWorksTest – we should maybe think about rewriting those?

PR #1715, opened 2018-03-21
Implement the internal models for Concept and QualifiedConcept

alexwlchan:
A stepping stone to #1682. It’s based on the pattern established for AbstractAgent/Agent/Organisation.

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/models/Concept.scala
+++ b/common/src/main/scala/uk/ac/wellcome/models/Concept.scala
@@ -1,6 +1,70 @@
 package uk.ac.wellcome.models
 
+sealed trait AbstractConcept {
+  val label: String
+  val ontologyType: String
+}
+
+// This is an example of a QualifiedConcept and a Concept from Silver's notes:
+//
+//   "subjects": [
+//     {
+//       "label": "Tuberculosis - prevention & control.",
+//       "type": "QualifiedConcept",
+//       "concept": {
+//         "id": "hd2891f",
+//         "label": "Tuberculosis",
+//         "type": "Concept",
+//         "identifiers": [
+//           {
+//             "type": "Identifier",
+//             "identifierScheme": "mesh",
+//             "value": "D014376Q000517"
+//           }
+//         ],
+//         "qualifiers": [
+//           {
+//             "qualifierType": "general-subdivision" ,
+//             "label": "prevention & control",
+//             "type": "Concept"
+//           }
+//         ]
+//       }
+//     },
+//   ]
+//
+// In particular, these models assume:
+//
+//  - A Work can have a mixture of Concept or a QualifiedConcept in the subject
+//    and genre fields
+//  - The "concept" on a QualifiedConcept is never another QualifiedConcept
+//  - The "qualifiers" field can be nested -- so you could have something like:
+//
+//        "qualifiers": [
+//          {
+//            "qualifierType": "subdivision",
+//            "label": "fizzbuzz",
+//            "type": "Concept",
+//            "qualifiers": [
+//              {
+//                "qualifierType": "sub-subdivision",
+//                "label": "fizz",
+//                "type": "Concept",
+//              }
+//            ]
+//          }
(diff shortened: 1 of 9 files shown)


sbt_common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayConcept.scala
```diff
@@ -17,7 +17,7 @@ case class DisplayConcept(
 }
 
 case object DisplayConcept {
-  def apply(concept: Concept): DisplayConcept = DisplayConcept(
+  def apply(concept: AbstractConcept): DisplayConcept = DisplayConcept(
```
alexwlchan: This is just enough to get the internal models working – I’ll add proper support for QualifiedConcepts in the display models in a separate patch.

alexwlchan: See #1717.

common/src/main/scala/uk/ac/wellcome/models/Concept.scala
```diff
@@ -1,6 +1,88 @@
 package uk.ac.wellcome.models
 
+sealed trait AbstractConcept {
```
rgladwell: As discusssed, we should probably override Circe's default discrimator value (the class name) to isolate our JSON format changing when we change class names. If this is safely or easily possible in Circe.

alexwlchan:
That would be a good thing to do, but it's out-of-scope for this patch (and would need us to change Agent as well). Merging.

common/src/main/scala/uk/ac/wellcome/models/Concept.scala
alicefuzier: Uhm you can configure the name of the field (as in it can be called anything, not necessarily `type`) but you can't configure the value stored in there, I don't think

common/src/main/scala/uk/ac/wellcome/models/Concept.scala
```diff
@@ -1,6 +1,88 @@
...
+//              }
+//            ]
+//          }
+//        ]
+//
+
 case class Concept(
   label: String,
+  qualifierType: Option[String] = None,
+  qualifiers: List[Concept] = List(),
```
alicefuzier: why is not the `QualifiedConcept` the one having `qualifiers`?

alexwlchan: If you look at @silveroliver’s JSON example, that’s where it falls.

common/src/test/scala/uk/ac/wellcome/models/ConceptTest.scala
```diff
@@ -0,0 +1,69 @@
...
+              "qualifiers": []
+            }
+          ]
+        }
+      }"""
+    )
+  )
+
+  it("serialises Concepts to JSON") {
+    forAll(testCases) { (concept: AbstractConcept, expectedJson: String) =>
```
alicefuzier: why is this a property test? If the two rows in the table test different things, they should be separate tests with different names to clarify what it is that we are testing

common/src/test/scala/uk/ac/wellcome/models/ConceptTest.scala
```diff
@@ -0,0 +1,69 @@
...
+      val actualJson = toJson(concept).get
+      assertJsonStringsAreEqual(actualJson, expectedJson)
+    }
+  }
+
+  it("deserialises JSON as Concepts") {
+    forAll(testCases) { (concept: AbstractConcept, expectedJson: String) =>
```
alicefuzier: same as above

common/src/test/scala/uk/ac/wellcome/models/ConceptTest.scala
alexwlchan: I usually do parametrised tests like this for reduced boilerplate.

PR #1717, opened 2018-03-21
Implement Display models for QualifiedConcept

alexwlchan:
A stepping stone to #1682. Builds on #1715.

### DIFF
--- a/sbt_common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayConcept.scala
+++ b/sbt_common/display/src/main/scala/uk/ac/wellcome/display/models/DisplayConcept.scala
@@ -2,7 +2,7 @@ package uk.ac.wellcome.display.models
 
 import com.fasterxml.jackson.annotation.JsonProperty
 import io.swagger.annotations.{ApiModel, ApiModelProperty}
-import uk.ac.wellcome.models.AbstractConcept
+import uk.ac.wellcome.models.{AbstractConcept, Concept, QualifiedConcept}
 
 @ApiModel(
   value = "Concept",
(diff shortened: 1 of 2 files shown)


common/src/test/scala/uk/ac/wellcome/models/ConceptTest.scala
```diff
@@ -0,0 +1,69 @@
+package uk.ac.wellcome.models
+
+import org.scalatest.{FunSpec, Matchers}
+import org.scalatest.prop.TableDrivenPropertyChecks._
+import uk.ac.wellcome.test.utils.JsonTestUtil
+import uk.ac.wellcome.utils.JsonUtil._
+
+class ConceptTest extends FunSpec with Matchers with JsonTestUtil {
+
+  val testCases = Table(
```
alicefuzier: Same as in the other PR. I don't really like this usage of property testing to shrink test cases. If all of the elements in this table are relevant tests cases then that should be made explicit by separating them in different tests

alicefuzier:
Looks good to me except the test thing and I'd rather get to a common agreement on how we write these tests before adopting this pattern everywhere

common/src/test/scala/uk/ac/wellcome/models/ConceptTest.scala
rgladwell: There are a few tests similar to this and I think I might be missing something: they seem to verify that the case classes serialize and deserialize. Is this testing/verifying anything else?

alicefuzier: No really. to be honest I wonder whether they even need to exist. I don't think all of the model classes have them (I don't think I added them for Agent). It's also tested again in `IdentifiedWorkTest` and `UnidentifiedWorkTest`

PR #1722, opened 2018-03-22
Fix WorksIndexTest to actually run

alicefuzier:
Make WorksIndexTest fail correctly

### Who is this change for?
🐚 🐟


sbt_common/elasticsearch/src/test/scala/uk/ac/wellcome/elasticsearch/WorksIndexTest.scala
```diff
@@ -1,68 +1,68 @@
...
-    with ExtendedPatience
     with Matchers
     with JsonTestUtil
     with BeforeAndAfterEach
+    with ExtendedPatience
     with PropertyChecks {
 
-  implicitly[Arbitrary[IdentifiedWork]]
+  // On failure, scalacheck tries to shrink to the smallest input that causes a failure.
+  // With IdentifiedWork, that means that it never actually completes.
```
alexwlchan: ???

Feels to me there’s a fairly natural order of shrinking on case classes, so I’d be interested to understand why ScalaCheck is deficient here. I’ll do some reading.

👍 for comment

PR #1724, opened 2018-03-22
Extract sierra creators 

alicefuzier:
Extract sierra creators but no identifiers for creators
### Who is this change for?
📟 🇲🇰


rgladwell:
Just need to fix the tests and this is ready to merge.

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraCreators.scala
```diff
@@ -0,0 +1,39 @@
+package uk.ac.wellcome.transformer.transformers.sierra
+
+import uk.ac.wellcome.models.{AbstractAgent, Organisation, Person}
+import uk.ac.wellcome.transformer.source.{MarcSubfield, SierraBibData}
+
+trait SierraCreators extends MarcUtils {
+
+  def getCreators(bibData: SierraBibData): List[AbstractAgent] = {
```
alexwlchan: A bit late now, but could you write a comment (or just copy Silver’s!) that explains the intent of the transformation you’re doing.

It makes it easier to review whether the transformation is correct, and means it’s easier to see if we need to make new changes compared to sierratransformable.

PR #1732, opened 2018-03-26
Remove the 'visible' field from DisplayWork

alexwlchan:
This is an internal-only field; it doesn't make sense to have it on the display model.


alicefuzier:
??? I didn't think we expose it... Anyway, good spot

PR #1679, opened 2018-03-14
Adding snapshot_convertor functionality

kenoir:
### What is this PR trying to achieve?

Add the necessary conversion functionality.

This application should stream files `elasticdump` (line separated JSON objects) format files into the same format where each line is transformed into a `DisplayWork` and then serialised as in the API.

### Who is this change for?

🎲 ⌨️ 

### Have the following been considered/are they needed?

- [ ] Deployed new versions
- [ ] Run `terraform apply`.

PR #1742, opened 2018-03-28
Implement the snapshot convertor logic as a flow

alexwlchan:
An attempt to break up #1679.

This implements some of the logic for #1679, but as its own flow so we can test it separately, and reduce the size of that PR as we try to debug it.


alexwlchan:
Yep, this is all I could get done without a compiler. 😄

alicefuzier:
Sorry I thought it was ready for review

PR #1746, opened 2018-03-28
Break out the S3 reader as a separate component in the snapshot_convertor

alexwlchan:
I discussed with @alicefuzier today; we both think this is another useful piece spun out of #1679. Tests are still failing at one point:

```
The future returned an exception of type: akka.stream.impl.io.ByteStringParser$ParsingException, with message: Parsing failed in step ReadHeaders. (S3SourceTest.scala:46)
```

I also want to dig into the gzipping in the tests, as I suspect there’s a more efficient approach than going via another Akka stream and dealing with futures for what should be an in-memory operation.

### DIFF
--- a/data_api/snapshot_convertor/src/test/scala/uk/ac/wellcome/platform/snapshot_convertor/source/S3SourceTest.scala
+++ b/data_api/snapshot_convertor/src/test/scala/uk/ac/wellcome/platform/snapshot_convertor/source/S3SourceTest.scala
@@ -1,0 +1,60 @@
+package uk.ac.wellcome.platform.snapshot_convertor.source
+
+import akka.stream.ActorMaterializer
+import akka.stream.scaladsl.{Compression, Sink, Source}
+import akka.util.ByteString
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.{FunSpec, Matchers}
+import uk.ac.wellcome.platform.snapshot_convertor.fixtures.AkkaS3
+import uk.ac.wellcome.test.fixtures.Akka
+
+import scala.concurrent.Future
+
+class S3SourceTest
+    extends FunSpec
+    with Matchers
+    with ScalaFutures
+    with Akka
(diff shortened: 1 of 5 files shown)


project/Dependencies.scala
```diff
@@ -27,6 +28,11 @@ object Dependencies {
     "com.typesafe.akka" %% "akka-stream" % versions.akka
   )
 
+  val akkaStreamDependencies = Seq(
+    "com.typesafe.akka" %% "akka-stream" % versions.akka,
```
alicefuzier: I think this should be removed from line 28 if it's moving here?

alicefuzier: Also `akkaStreamDependencies` should maybe something like ```akkaDependencies ++ Seq(...)``` since you can't have akka streams without akka

data_api/snapshot_convertor/src/test/scala/uk/ac/wellcome/platform/snapshot_convertor/source/S3SourceTest.scala
```diff
@@ -0,0 +1,76 @@
...
+    // the most "Scala-like" way, it's to create a file that closely
+    // matches what we'd get from an elasticdump in S3.
+    val bw = new BufferedWriter(new FileWriter(tmpfile))
+    bw.write(content)
+    bw.close()
+
+    // This performs "gzip foo.txt > foo.txt.gz" in a roundabout way.
+    // Because we're using the busybox version of gzip, it actually cleans
+    // up the old file and appends the .gz suffix for us.
+    s"gzip ${tmpfile.getPath}" !!
```
alicefuzier: I didn't know you could do that. Nice

project/Dependencies.scala
alexwlchan: Ah, I hadn’t noticed that was duplicated. And 👍 on making it `akkaDependencies ++`.

PR #1748, opened 2018-03-29
Add a materialiser fixture, and use that throughout

alexwlchan:
Another piece spun out of #1679, and some cleanup in our fixtures which use Akka.

I’m working on a slow laptop this morning, so relying on Travis for compilations for now!

### DIFF
--- a/common/src/test/scala/uk/ac/wellcome/test/fixtures/Akka.scala
+++ b/common/src/test/scala/uk/ac/wellcome/test/fixtures/Akka.scala
@@ -1,6 +1,7 @@
 package uk.ac.wellcome.test.fixtures
 
 import akka.actor.ActorSystem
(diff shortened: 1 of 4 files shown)


alexwlchan:
Somebody with more Scala-fu than me might know how to get `implicit` working with the materializers; in the meantime, I think this is still a step forward from where we were.

alicefuzier:
What do you mean by getting the implicit to work? Passing the `actorSystem` implicitly to `ActorMatrieliser`? If that's what you mean I think in this instance is better to pass it explicitly

alexwlchan:
It’s not clear from the current diff – previously I was trying to do something like `implicit val materializer = materializer` so it would be picked up by the sinks, but now I pass it explicitly and I think that’s probably fine.

PR #1749, opened 2018-03-29
snapshot_convertor: Add a flow for converting Strings into gzip-compressed bundles

alexwlchan:
Another flow pulled out of #1679. I believe this is the final flow we need, and then I’ll go back and fit these into the previous PR.

Note this found a bug in the original code – in particular, we weren’t adding newlines in the final output, so every document would be squished onto a single line!

### DIFF
--- a/data_api/snapshot_convertor/src/test/scala/uk/ac/wellcome/platform/snapshot_convertor/flow/StringToGzipFlowTest.scala
+++ b/data_api/snapshot_convertor/src/test/scala/uk/ac/wellcome/platform/snapshot_convertor/flow/StringToGzipFlowTest.scala
@@ -54,13 +54,13 @@ class StringToGzipFlowTest
         .runWith(fileSink)
 
       whenReady(future) { _ =>
-
         // Unzip the file, then open it and check it contains the strings
         // we'd expect.  This is the Busybox version of gzip, which strips
         // the .gz from the filename.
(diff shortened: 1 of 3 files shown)


data_api/snapshot_convertor/src/test/scala/uk/ac/wellcome/platform/snapshot_convertor/flow/StringToGzipFlowTest.scala
```diff
@@ -0,0 +1,72 @@
...
+    with Akka
+    with ScalaFutures {
+
+  it("produces a gzip-compressed file from the lines") {
+    withActorSystem { actorSystem =>
+      implicit val executionContext: ExecutionContextExecutor =
+        actorSystem.dispatcher
+      implicit val system: ActorSystem = actorSystem
+      implicit val materializer: Materializer =
+        ActorMaterializer()(actorSystem)
```
alicefuzier: I think I saw a `withActorMaterialiser` fixture while reviewing another PR (can't remember which PR...)

alexwlchan: Yep, that’s in at least two other PRs. I’ll retrofit it to #1748. 😛

PR #1755, opened 2018-03-29
[WIP] Implement the ConvertorService

alexwlchan:
This removes the ConvertorService from #1679, uses all the new flows and sources, and adds a bunch of tests around it. Plus moving some of the gzip-related code into a utility trait.

Not quite working yet (I think), but I think the mainline function is okay.


alexwlchan:
Like #1679, this needs breaking up into smaller pieces.

PR #1756, opened 2018-03-29
Compose fixtures

rgladwell:

Reduce the boilerplate of composing fixtures together, using type classes.


alexwlchan:
I’m very much 👍 on the concept, but I’d like to sit down and walk through the FixtureComposers together to make sure I’ve really understood it before we merge.

PR #1759, opened 2018-04-03
Implement the conversion service

alexwlchan:
Breaking up #1755 and #1679 even further for easier review.

This patch adds the ConvertorService, which implements a single method:

```scala
class ConvertorService {
    def runConversion(conversionJob: ConversionJob): Future[CompletedConversionJob]
}
```

and runs the body of the conversion work.

The WorkerService (reading from SQS, extracting the message, pushing to SNS afterward) will be a separate patch.

### DIFF
--- a/data_api/snapshot_convertor/src/main/scala/uk/ac/wellcome/platform/snapshot_convertor/services/ConvertorService.scala
+++ b/data_api/snapshot_convertor/src/main/scala/uk/ac/wellcome/platform/snapshot_convertor/services/ConvertorService.scala
@@ -5,6 +5,7 @@ import scala.concurrent.Future
 import scala.util.{Success, Failure}
 
 import akka.actor.ActorSystem
+import akka.http.scaladsl.model.Uri
 import akka.stream.ActorMaterializer
 import akka.stream.alpakka.s3.scaladsl.{MultipartUploadResult, S3Client}
 import akka.stream.scaladsl.Sink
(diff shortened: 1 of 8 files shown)


data_api/snapshot_convertor/src/main/scala/uk/ac/wellcome/platform/snapshot_convertor/services/ConvertorService.scala
```diff
@@ -1,18 +1,85 @@
...
+      }
+      .map {
+        case Success(jsonString) => jsonString
+        case Failure(encodeError) => {
+          warn("Failed to convert $displayWork to string!", encodeError)
+          throw encodeError
+        }
+      }
+
+    val gzipContent = jsonStrings
```
alicefuzier: This is a `Flow` (if I'm reading correctly). It could be useful to say that in the variable name. It would make things easier to follow at least for me. Something like `gzipContentFlow`. Same for `displayWorks` and `jsonStrings` above

alexwlchan: Yep. Sensible suggestion, will change.

alexwlchan: Change of plan – I’m going to add type annotations instead. Slightly clearer, and clarity enforced by the compiler! 😄

alicefuzier: 👍

PR #1772, opened 2018-04-03
Id minter terminal failure

alicefuzier:
The id minter is currently sending alarms of TerminalFailures. These seems to be due mostly to different processes trying to mint identifiers for the same object. In that case, only one process succeeds while the others fail with exceptions like```java.sql.SQLIntegrityConstraintViolationException: Duplicate entry 'Item-sierra-system-number-i10824133' for key 'UniqueFromSource'```, which trigger the TerminalFailure alarm. 
It is an expected behaviour that sometimes more than one process try to identify the same message and it is expected behaviour that only one of those succeeds, so the failing ones should not be sending alarms
### Who is this change for?
⭐️ 🔌

### DIFF
--- a/catalogue_pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
+++ b/catalogue_pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
@@ -1,11 +1,13 @@
 package uk.ac.wellcome.platform.idminter.database
 
-import javax.inject.Singleton
+import java.sql.SQLIntegrityConstraintViolationException
 
 import com.google.inject.Inject
 import com.twitter.inject.Logging
+import javax.inject.Singleton
 import scalikejdbc._
-import uk.ac.wellcome.models.{IdentifierSchemes, SourceIdentifier}
+import uk.ac.wellcome.exceptions.GracefulFailureException
+import uk.ac.wellcome.models.SourceIdentifier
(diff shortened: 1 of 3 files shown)


catalogue_pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
```diff
@@ -71,11 +73,13 @@ class IdentifiersDao @Inject()(db: DB, identifiers: IdentifiersTable)
             )
         }.update().apply()
       }
+    } recover {
+      case e: SQLIntegrityConstraintViolationException =>
+        warn(s"Failed inserting identifier $identifier", e)
```
alexwlchan: Could we get a more descriptive log message please? Maybe something like:

```scala
warn(s"Unable to insert $identifier because of integrity constraints", e)
```

to distinguish it from the message below.

catalogue_pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/database/IdentifiersDao.scala
```diff
@@ -71,11 +73,15 @@ class IdentifiersDao @Inject()(db: DB, identifiers: IdentifiersTable)
             )
         }.update().apply()
       }
+    } recover {
+      case e: SQLIntegrityConstraintViolationException =>
+        warn(
+          s"Unable to insert $identifier because of integrity constraints",
```
alexwlchan: 👍

PR #1773, opened 2018-04-03
Increase the default number of messages read for all applications

alicefuzier:
Apparently, the id minter is not the only service being a bit slow. All of them read 1 message per second so they sit idle most of the time
🏷 ◾️


alicefuzier:
This is pretty, so i'm just going to merge it. But please do shout if you have any comments

PR #1775, opened 2018-04-04
Remove spaces from code identifiers

alicefuzier:
This should help clear the transformer DLQ. I see a lot of failures like this one ```java.lang.RuntimeException: Multiple identifiers in subfield $0: List(n 82270463 , n 82270463)```. This is due to the transformer finding more than one distinct identifier code for a creator, but these aren't actually distinct after normalisation, so they should pass transformation once this change is in
### Who is this change for?
🛋 ⛓


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraCreators.scala
```diff
@@ -49,16 +47,20 @@ trait SierraCreators extends MarcUtils {
...
         val organisation = Organisation(name.get)
         identify(codes, organisation, "Organisation")
       }
 
     persons ++ organisations
   }
 
+  private def getIdentifierCodes(subfields: List[MarcSubfield]) = {
+    subfields.collect {
+      case MarcSubfield("0", content) => content.replaceAll("\\s", "")
```
alexwlchan: Can we get a comment and example explaining why we're stripping spaces? The one from the PR description would be enough.

alicefuzier: 👍

Issue #1778, opened 2018-04-04
Washup tasks for snapshots


alexwlchan:
aka everything left over after #1771 

* [x] Fix the elasticdump container so it doesn't get terminated by healthchecks (#1787)
* [x] Serialise to DisplayWork with Jackson (#1788)
* ~Bump the scaling time on elasticdump~
* [x] elasticdump should be able to get other indexes from SQS (#1787)
* [x] snapshot_convertor should read target bucket from the SQS message (#1847)
* [x] snapshot_convertor should read target key from the SQS message (#1847)
* [x] Nice domain (#1851)
* [x] Expire old dumps in the private bucket (#1812)
* [x] Reindex the Miro data so we can snapshot the /v1/ API

---

Part 2:

- [x] Upload dumps to the correct key (`/catalogue/v1/works.json.gz`)(#1866 )
- ~Upload to S3 as Standard-IA~
- [x] Redirect the / root to the docs (#1854)
- ~Proper error pages~ (#1855)
- ~Allow sending custom ES config into elasticdump, needs config for:~ (#1868)
  - ~es index~
  - ~upload bucket~
  - ~version~

alexwlchan:
I’ve bumped “upload as Standard-IA” because there’s a minimum 30-day charge, and all our snapshot materials are either deleted after 30 days or replaced daily, so that’s less useful!

I’m also going to push out proper error pages for now, as I think that’s very much a bit of polish rather than core functionality.

PR #1787, opened 2018-04-05
elasticdump: use a proper Python web server, with a healthcheck

alexwlchan:
This rewrites the elasticdump image to be a Python application with a proper web server and healthcheck endpoint, rather than doing it all in bash.

This:

* means the elasticdump container can run arbitrarily long, and shouldn't be timed out by the ALB healthchecks
* opens the possibility of testing it properly, rather than relying on it “just working” as a shell script

Part of #1778.


alexwlchan:
A few tests are failing, but I think they’re the ones on master, and I fail to see how this PR could have affected them. Merging anyway!

PR #1788, opened 2018-04-05
snapshot_convertor: Serialise to DisplayWork with Jackson

alexwlchan:
Part of #1778.

Now DisplayWork snapshots created by the snapshot convertor more closely resemble the output of the Catalogue API.


alexwlchan:
Merging as @alicefuzier paired on this, and there’s nobody around for another review!

PR #1847, opened 2018-04-06
[WIP] Snapshot converter target info

alicefuzier:
Make snapshot convertor read target bucket and target key from the SQS message
### Who is this change for?
🍔 🎃


alexwlchan:
💃

PR #1850, opened 2018-04-09
Fix for Java 9

rgladwell:
Add JAXB depedency missing from Java 9 classpath.

Adds dependency for Java 9. Although this project uses Java 8, this prospective fix will make an upgrade easier later.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -20,6 +20,7 @@ object Dependencies {
     val scalaCheckVersion = "1.13.4"
     val scalaCheckShapelessVersion = "1.1.6"
     val sierraStreamsSourceVersion = "0.2"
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -87,6 +87,7 @@ object Dependencies {
     "org.scalatest" %% "scalatest" % versions.scalatest % "test",
     "com.novocode" % "junit-interface" % versions.junitInterface % "test",
     "org.scalacheck" %% "scalacheck" % versions.scalaCheckVersion % "test",
+    "javax.xml.bind" % "jaxb-api" % "2.2.11",
```
alexwlchan: Can we specify the javax version in the versions block at the top, like our other pinned versions?

PR #1856, opened 2018-04-10
Move API JSON serialisation tests to display-common

alicefuzier:
So that 
- we can test snapshot and api serialisation in one place
- we can test serialisation independently from elasticsearch ingestion
- we can lay the foundation for testing v1 and v2 display work serialisation separately

### Who is this change for?
📦 🇱🇦


sbt_common/display/src/test/scala/uk/ac/wellcome/display/models/DisplayJacksonModuleTestBase.scala
```diff
@@ -0,0 +1,139 @@
...
+    }"""
+
+  def identifier(identifier: SourceIdentifier) =
+    s"""{
+      "type": "Identifier",
+      "identifierScheme": "${identifier.identifierScheme}",
+      "value": "${identifier.value}"
+    }"""
+
+  def identifiedOrUnidentifiable[T](displayableAgent: Displayable[T],
```
alexwlchan: What does this do? A comment would be appreciated, because it’s not obvious to me.

alexwlchan:
Rebased against current master (was 0fed95a).

alexwlchan:
I think I’ve moved the remaining tests from ApiWorksTest that we might want to keep, and even some of those we might be able to drop. Ready for review @wellcometrust/platform-devs.

PR #1863, opened 2018-04-11
Move WorksUtil and the JSON test helpers into the display library

alexwlchan:
Pulled out as a standalone piece from #1856. Mostly unmodified from the original patch, aside from the additional comment and new variable names in `identifiedOrUnidentifiable`.


catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/services/WorksServiceTest.scala
```diff
@@ -3,15 +3,15 @@ package uk.ac.wellcome.platform.api.services
...
 import uk.ac.wellcome.platform.api.fixtures.{
   ElasticsearchServiceFixture,
   WorksServiceFixture
 }
 import uk.ac.wellcome.display.models.{
   DisplayIdentifier,
   DisplayWork,
-  WorksIncludes
+  WorksIncludes,
+  WorksUtil
```
kenoir: Is it worth putting this in a test namespace?

alexwlchan: Maybe, but I don’t care enough right now to wait another CI cycle. 🙃

PR #1865, opened 2018-04-11
Replace DynamoInserter logic in sierra_items_to_dynamo

kenoir:

The `DynamoInserter` was removed in a previous change but is required to properly ingest item changes. This PR replaces the `DynamoInserter` and updates tests to ensure it cannot be removed by accident in the future.

### Who is this change for?

🌾 🍑 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


alexwlchan:
Resolves #1860.

PR #1869, opened 2018-04-11
Freeze the V1 DisplayWork model

alexwlchan:
Initial approach at freezing the V1 DisplayWork with its own namespace and tests.

~Probably not for actual use, just testing the waters.~

### DIFF
--- a/sbt_common/display/src/main/scala/uk/ac/wellcome/display/models/v1/DisplayWorkV1.scala
+++ b/sbt_common/display/src/main/scala/uk/ac/wellcome/display/models/v1/DisplayWorkV1.scala
@@ -1,0 +1,145 @@
+package uk.ac.wellcome.display.models.v1
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import io.swagger.annotations.{ApiModel, ApiModelProperty}
+import uk.ac.wellcome.display.models._
+import uk.ac.wellcome.models.IdentifiedWork
+
+@ApiModel(
+  value = "Work",
+  description =
+    "An individual work such as a text, archive item or picture; or a grouping of individual works (so, for instance, an archive collection counts as a work, as do all the series and individual files within it).  Each work may exist in multiple instances (e.g. copies of the same book).  N.B. this is not synonymous with \\\"work\\\" as that is understood in the International Federation of Library Associations and Institutions' Functional Requirements for Bibliographic Records model (FRBR) but represents something lower down the FRBR hierarchy, namely manifestation. Groups of related items are also included as works because they have similar properties to the individual ones."
+)
+case class DisplayWorkV1(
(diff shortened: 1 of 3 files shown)


alexwlchan:
Thoughts on how we’d actually do this: add a DisplayWorksService to the API which gets a flag `api.displaywork.model.version` or something like that. That flag has a fixed set of values it can take (cf. the bibs/items stuff in the Sierra readers).

This service presents a single function `toDisplayWork(work: IdentifiedWork, includes: WorksIncludes)`.

It gets injected into ManagementController at startup, and picks which DisplayWork model to use based on the value of this flag. Thoughts?

sbt_common/display/src/main/scala/uk/ac/wellcome/display/models/v1/DisplayWorkV1.scala
```diff
@@ -0,0 +1,145 @@
...
+import io.swagger.annotations.{ApiModel, ApiModelProperty}
+import uk.ac.wellcome.display.models._
+import uk.ac.wellcome.models.IdentifiedWork
+
+@ApiModel(
+  value = "Work",
+  description =
+    "An individual work such as a text, archive item or picture; or a grouping of individual works (so, for instance, an archive collection counts as a work, as do all the series and individual files within it).  Each work may exist in multiple instances (e.g. copies of the same book).  N.B. this is not synonymous with \\\"work\\\" as that is understood in the International Federation of Library Associations and Institutions' Functional Requirements for Bibliographic Records model (FRBR) but represents something lower down the FRBR hierarchy, namely manifestation. Groups of related items are also included as works because they have similar properties to the individual ones."
+)
+case class DisplayWorkV1(
```
kenoir: No need to dupe the version identifier in the name when namespaced?

alicefuzier: True, but it might be helpful with IDE autocompletion and in general to avoid errors. It's more visible in the class name

rgladwell: I must admit, I'd prefer not to dupe the version in the class name as well. You can always alias the type name locally (`import *.v1.DisplayWork => V1DisplayWork`).

alexwlchan:
So I see the argument for not duplicating information in the namespace/class name, but I would lean towards putting it in the name and dropping the extra directory. You can alias the type name locally, but that causes IDE issues and makes it more likely that we'll use the name inconsistently.

alexwlchan:
It’s also easier to see which version you’re looking at when you have a single file open.

alexwlchan:
Closing and reopening to see if Travis notices.

PR #1870, opened 2018-04-11
Per message config for elasticdump

### LINKED ISSUE #1868: Modify the elastic dump container to support saving snapshots from different indexes
Pulled out of #1778 as a nice-to-have but not essential feature of snapshots.

(Especially if the snapshot convertor can do V1 and V2 from the same index.)

kenoir:

Configure elasticdump on a per message basis (fixes #1868 )

### Who is this change for?

☄️ 💻 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.

### DIFF
--- a/data_api/elasticdump/run_elasticdump.py
+++ b/data_api/elasticdump/run_elasticdump.py
@@ -14,7 +14,7 @@ import boto3
 from service_utils import service
 
 
-# This class is duplicated in the elasticdump app
+# This class is duplicated in the snapshot_scheduler app
 # Changes here will need to be reflected there.
 @attr.s
 class SnapshotRequest(object):
@@ -40,7 +40,7 @@ def get_message(sqs_client, sqs_queue_url):
 
 
 def parse_message(message):
-    raw_messsage_dict = json.loads(message)
+    raw_messsage_dict = json.loads(message['Body'])
 
     return SnapshotRequest(**raw_messsage_dict)
 
@@ -79,11 +79,6 @@ def aws_client(service_name):
 def run():
     print(os.environ)
     sqs_queue_url = os.environ['sqs_queue_url']
-
-    # TODO: Currently we read the target bucket from the environment config.
-    # It would be nice to support overriding the bucket name with config from
-    # the SQS message.
-    target_bucket = os.environ['upload_bucket']
 
     s3_client = aws_client('s3')
     sqs_client = aws_client('sqs')
--- a/data_api/elasticdump/run_elasticdump.py
+++ b/data_api/elasticdump/run_elasticdump.py
@@ -87,8 +87,9 @@ def run():
     message = get_message(sqs_client=sqs_client, sqs_queue_url=sqs_queue_url)
     snapshot_request = parse_message(message)
 
-    print('*** Constructing Elasticsearch URL')
+    print(f'*** Parsed snapshot request {snapshot_request!r} from message')
 
+    print('*** Constructing Elasticsearch URL')
     es_index = snapshot_request.es_index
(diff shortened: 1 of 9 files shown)


data_api/elasticdump/run_elasticdump.py
```diff
@@ -73,14 +90,11 @@ def run():
 
     print('*** Reading messages from SQS')
     message = get_message(sqs_client=sqs_client, sqs_queue_url=sqs_queue_url)
+    snapshot_request = parse_message(message)
```
alexwlchan: We should log the request and/or the message we received.

data_api/snapshot_scheduler/src/test_snapshot_scheduler.py
```diff
@@ -15,10 +15,20 @@ def utcnow(cls):
 
 @mock.patch('datetime.datetime', patched_datetime)
 def test_writes_message_to_sqs(sns_client, topic_arn):
-    snapshot_scheduler.main(sns_client=sns_client)
+    target_bucket_name="target_bucket_name"
+    es_index="es_index"
```
alexwlchan: nit: Python convention is to have spaces around assignment operators.

data_api/snapshot_scheduler/src/test_snapshot_scheduler.py
```diff
@@ -15,10 +15,20 @@ def utcnow(cls):
...
+
+    snapshot_scheduler._run(
```
alexwlchan: Ick. Underscore methods are generally treated as “private” – you can access them, but they’re treated as internal and you shouldn’t rely on them. Better would be to set up the environment variables you expect, and call the `main` function.

kenoir:
DO NOT MERG! FLAKY TESTS DETECTED!

data_api/elasticdump/run_elasticdump.py
```diff
@@ -28,6 +39,12 @@ def get_message(sqs_client, sqs_queue_url):
         return message
 
 
+def parse_message(message):
+    raw_messsage_dict = json.loads(message['Body'])
```
alexwlchan: messsage?

data_api/elasticdump/test_run_elasticdump.py
```diff
@@ -12,25 +13,25 @@
...
-        'index_with_no_scheme',
-        'https://elastic:changeme@example.org:9200/index_with_no_scheme'
+            {
+                'es_username': 'elastic',
+                'es_password': 'changeme',
+                'es_hostname': 'example.org',
+                'es_port': 9200,
+            },
+            'index_with_no_scheme',
+            'https://elastic:changeme@example.org:9200/index_with_no_scheme'
```
alexwlchan: Does this extra indent add anything except cluttering the diff?

data_api/snapshot_scheduler/src/test_snapshot_scheduler.py
```diff
@@ -15,10 +15,20 @@ def utcnow(cls):
 
 @mock.patch('datetime.datetime', patched_datetime)
 def test_writes_message_to_sqs(sns_client, topic_arn):
-    snapshot_scheduler.main(sns_client=sns_client)
+    target_bucket_name = "target_bucket_name"
+    es_index = "es_index"
+
+    snapshot_scheduler._run(
```
alexwlchan: See previous comment about testing with `_run`, come chat to me if you want to go over this.

data_api/elasticdump/test_run_elasticdump.py
kenoir: This is how intellij decided to format the file, will fix.

PR #1871, opened 2018-04-11
Updated rules for subject

silveroliver:
Updated diagrams and examples for subjects. Still need to expand it
beyound 650.


ontologies/WIP/sierratransformable
```diff
@@ -277,27 +275,27 @@
    *{
    *  "subjects": [
    *    {
-   *      "label": "Tuberculosis - prevention & control.",
+   *      "label": "Tuberculosis -- prevention & control -- Paris -- 1850 -- bibliography.",
```
kenoir: I do wonder if we want to agree a scheme for adding a label that makes is fairly display logic agnostic (i.e. make this data a bit more obviously a delimiter separated string).

silveroliver: Yes this is something I meant to bring up. Though the librarians have suggested double hyphens it does look odd to me.

Perhaps one we need to speak to Experience team about?

kenoir: Paging @jamesgorrie!

jamesgorrie: @kenoir @silveroliver You might need to run me through what this text is trying to say?
As it stands I can't really infer anything.

jennpb: Down with double hyphens! I will be very surprised if we actually need to use them in the UI; we're more likely to normalise to single hyphens if needed.

kenoir: So the question is ... is the label at all useful to anyone? Is there a way to make it more useful to consumers?

jennpb: We don't know that yet because we haven't done any subject-string testing.

kenoir: So maybe we just leave it as it is and reindex later based on testing?

PR #1875, opened 2018-04-12
Start using travistooling for our CI scripts

alexwlchan:
This pulls migrates the `run_travis_autoformat.py` script to use the new tooling library, with some extra tests to boot.


alexwlchan:
And yet, ironically, it yaps because the formatting code has a formatting error. Oops. 🙈

alexwlchan:
The last commit is a single whitespace change, so assuming it finally passes CI, I'll merge without waiting for another review.

PR #1876, opened 2018-04-12
loris: put the expiry policy on the correct bucket

alexwlchan:


alexwlchan:
We should really move this entirely out of the Loris stack.

alexwlchan:
Introduced in https://github.com/wellcometrust/platform/pull/1853.

PR #1879, opened 2018-04-12
Make v1 a parameter in the route declaration

alicefuzier:
get to a point where we can declare another route for v2
### Who is this change for?
🍹 🐝

### DIFF
--- a/catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/finatra/exceptions/ElasticsearchResponseExceptionMapper.scala
+++ b/catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/finatra/exceptions/ElasticsearchResponseExceptionMapper.scala
@@ -33,14 +33,21 @@ class ElasticsearchResponseExceptionMapper @Inject()(
   // When returning a 400 to the user, we wrap this error to avoid talking
   // about internal Elasticsearch concepts.
   private val resultSizePattern =
-  """Result window is too large, from \+ size must be less than or equal to: \[([0-9]+)\]""".r.unanchored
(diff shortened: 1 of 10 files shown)


catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/finatra/exceptions/ElasticsearchResponseExceptionMapper.scala
```diff
@@ -56,9 +32,57 @@ class ElasticsearchResponseExceptionMapper @Inject()(
...
+      // fails: "NoSuchElementException: next on empty iterator".  I think
+      // this happens when the cluster doesn't reply, but I'm not sure.
+      val jsonDocument = exception.getMessage
+        .split("\n")
+        .tail
+        .head
+      jsonToError(jsonDocument = jsonDocument, exception = exception)
+    } catch {
+      case e: Exception =>
+        serverError(s"Error ($e) parsing Elasticsearch response", exception)
```
kenoir: I don't think these errors should be user facing. We should log the ES error but not pass anything on to the user other than a 500 (we might want to create a request id, so we an match that to the actual error in logs if need be).

alicefuzier: I agree but this should probably be done in a different PR. This PR should not cause any change in the API behaviour

kenoir: 👍

PR #1881, opened 2018-04-12
Use v1 display models

alicefuzier:
Use the new DisplayWorkv1 case class and decommission the old DisplayWork. It's based on #1879 
### Who is this change for?
⚙️ 🍶


catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/finatra/exceptions/ElasticsearchResponseExceptionMapper.scala
```diff
@@ -56,9 +32,57 @@ class ElasticsearchResponseExceptionMapper @Inject()(
...
+      // fails: "NoSuchElementException: next on empty iterator".  I think
+      // this happens when the cluster doesn't reply, but I'm not sure.
+      val jsonDocument = exception.getMessage
+        .split("\n")
+        .tail
+        .head
+      jsonToError(jsonDocument = jsonDocument, exception = exception)
+    } catch {
+      case e: Exception =>
+        serverError(s"Error ($e) parsing Elasticsearch response", exception)
```
kenoir: Did I comment elsewhere about not exposing ES errors to the user?

alicefuzier: Yes, this branch was created of the one that was merged in #1879 , so when that got merged it caused conflicts and the diff was al over the place. It should be ok now

PR #1891, opened 2018-04-13
[wip] Fix SNS max message size

### LINKED ISSUE #1776: We're exceeding SNS maximum message size

- [x] Shared model
- [x] Pointer writer
- [x] Pointer reader

rgladwell:

- [ ] Deployed new versions
- [ ] Run `terraform apply`.

### DIFF
--- a/catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/receive/SQSMessageReceiverTest.scala
+++ b/catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/receive/SQSMessageReceiverTest.scala
@@ -20,7 +20,7 @@ import uk.ac.wellcome.models.{
   SourceIdentifier,
   UnidentifiedWork
 }
-import uk.ac.wellcome.sns.{PublishAttempt, SNSWriter}
+import uk.ac.wellcome.sns.SNSWriter
 import uk.ac.wellcome.test.fixtures.{S3, SNS, SQS, TestWith}
 import uk.ac.wellcome.test.fixtures.S3.Bucket
 import uk.ac.wellcome.test.fixtures.SNS.Topic
@@ -299,8 +299,8 @@ class SQSMessageReceiverTest
 
   private def mockSNSWriter = {
     val mockSNS = mock[SNSWriter]
-    when(mockSNS.writeMessage(anyString(), any[String]))
-      .thenReturn(Future.successful(Right(PublishAttempt("1234", "5678"))))
+    val result: Future[Unit] = Future.successful(Unit)
+    when(mockSNS.writeMessage(anyString(), anyString())).thenReturn(result)
     mockSNS
   }
 
@@ -308,8 +308,7 @@ class SQSMessageReceiverTest
     val mockSNS = mock[SNSWriter]
     when(mockSNS.writeMessage(anyString(), any[String]))
       .thenReturn(
-        Future.successful(
-          Left(new RuntimeException("Failed publishing message"))))
+        Future.failed(new RuntimeException("Failed publishing message")))
     mockSNS
   }
 }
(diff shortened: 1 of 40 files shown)


common/src/main/scala/uk/ac/wellcome/sns/SNSWriter.scala
```diff
@@ -6,32 +6,55 @@ import com.google.inject.Inject
...
-
-  private def toPublishRequest(message: String, subject: String) =
-    new PublishRequest(snsConfig.topicArn, message, subject)
+case class PublishAttempt(id: String, key: String)
+
+class SNSWriter @Inject()(
+  client: AmazonSNS,
+  config: SNSConfig,
+  s3Client: AmazonS3,
+  s3Config: S3Config
```
alicefuzier: I think the design that we came up with is not really to modify `SNSWriter` to write to s3 but rather to write a wrapper object (we haven't really decided on a name yet, something like `SNSPointerWriter`) that uses `SNSWriter` to send to a SNS message and `S3ObjectStore` to write to s3. So `SNSWriter` probably shouldn't change much

alicefuzier: Paging @alexwlchan and @kenoir just to make sure we're all on the same page

alexwlchan: Yeah, this looks wrong to me. I was thinking the existing SNSWriter would stay the same, and we'd add an entirely new class.

common/src/test/scala/uk/ac/wellcome/sns/SNSWriterTest.scala
```diff
@@ -3,49 +3,88 @@ package uk.ac.wellcome.sns
...
+
+          getContentFromS3(bucket, publishAttempt.right.value.key) should be(
+            message)
+        }
+      }
+    }
+  }
+
+  it(
+    "should return a failed future if it fails to publish the message pointer") {
```
kenoir: We try and avoid `should` in test descriptions as it's redundant. The test asserts something _will_ happen and fails if it doesn't. Apologies for the pedantry.

kenoir: This set of tests looks good though.

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/receive/SQSMessageReceiverTest.scala
```diff
@@ -72,7 +72,12 @@ class SQSMessageReceiverTest
   )(testWith: TestWith[SQSMessageReceiver, R]) = {
 
     val snsWriter =
-      maybeSnsWriter.getOrElse(new SNSWriter(snsClient, SNSConfig(topic.arn)))
+      maybeSnsWriter.getOrElse(
+        new SNSWriter(
+          snsClient,
+          SNSConfig(topic.arn),
+          s3Client,
```
kenoir: You can probably just inject: https://github.com/wellcometrust/platform/blob/master/common/src/main/scala/uk/ac/wellcome/s3/S3ObjectStore.scala

kenoir: 👍 Presumed this was implicit from previous comments.

common/src/main/scala/uk/ac/wellcome/s3/S3Uri.scala
```diff
@@ -0,0 +1,14 @@
+package uk.ac.wellcome.s3
+
+import java.net.URI
+
+object S3Uri {
```
kenoir: Why use a URI over having a mechanism to serialise/deserialise: https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/S3Object.html?

kenoir: There's also https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/AmazonS3URI.html that more conveniently allows us to extract bucket/key from a URI.

kenoir: Perhaps actually using a plain URI here is ok - as we can use `AmazonS3URI` the other side to extract the bucket and key, making the sending side know a bit less about the storage medium.

PR #1892, opened 2018-04-14
Panic! in the Build System (Use new library for Travis runner)

alexwlchan:
This PR migrates our Travis test runner to use the new `travistooling` library. Bonus is more testing, more reliability, and we can start acting on the Travis-ECR notifications to improve build times by patching the new library.

This is largely driven by thinking around https://github.com/HypothesisWorks/hypothesis/issues/1224 – I’d really like to have us using travistooling and getting it working properly _before_ I try to use it in another project!


alexwlchan:
Oh, and I ditched the `.travis` directory, because dotfiles being hidden on macOS was just making that mildly tedious.

alexwlchan:
Niiiiiice.

screen shot 2018-04-14 at 17 59 28

alexwlchan:
In general, changes to the Travis build scripts (travistooling, .travis, .travis.yml) probably shouldn’t trigger a full rebuild – but I’m going to save that for a separate patch, as I want to check we can still do a full build with this setup.

run_travis_task.py
```diff
@@ -0,0 +1,118 @@
...
+                  | Relevant changes | No relevant changes
+   ---------------+------------------+---------------------
+    pull request  | Run tests        | Don't run tests
+                  | Don't deploy     | Don't deploy
+   ---------------+------------------+---------------------
+    master        | Run tests        | Run tests
+                  | Deploy to AWS    | Don't deploy
+
+We always run tests on master so we get consistent build results, that's
+less important on master where results are transient.
```
alicefuzier: I think you meant less important on pull requests?

run_travis_task.py
```diff
@@ -0,0 +1,118 @@
...
+
+    publish_task = task.replace('-build', '-publish')
+    publish_task = task.replace('-test', '-publish')
+
+    if _should_publish(task=task, travis_event_type=travis_event_type):
+        print("*** We're going to run the publish task")
+        make(publish_task)
+    else:
+        print("*** We don't need to run the publish task")
+        make(publish_task, '--dry-run')
```
alicefuzier: Why do we do a dry run here rather than just return 0?

run_travis_task.py
alexwlchan: Oops. Yes, will fix.

run_travis_task.py
alexwlchan: Doing dry-run ensures the task actually exists, so we don’t get errors on master if we’ve defined a test task but no publish task. Will add a comment.

Issue #1878, opened 2018-04-12
After-action review items after accidentally deleting most of the Loris source images


alexwlchan:
- [x] Move the CloudFront logs bucket into the shared_infra stack (#1886)
- [x] Put all data stores/assets/buckets in their own file (#1880)
- [x] Restrict delete/modify S3/DynamoDB/RDS permissions for devs (https://github.com/wellcometrust/platform-private/pull/24)
- [x] Get a report of additional 404s during the outage
- [x] Add alarms on high CloudFront error rates (#1893)

PR #1893, opened 2018-04-16
[WIP] Get notifications of CloudFront 4xx errors in Slack

alexwlchan:
Required for #1878. Give us more visibility into when we've accidentally deleted a bunch of objects from the Loris S3 bucket. 🙈 

It's a bit of a faff, because CloudFront (and thus CloudFront metrics) all live in us-east-1, whereas the rest of our infra lives in eu-west-1. The monitoring stack isn't planning/applying cleanly, because Terraform can't seem to import resources that live in a different region.

Possibly we should just count ALB metrics instead?


alexwlchan:
Okay, with the multiple providers stuff, it’s applying cleanly now.

Issue #1842, opened 2018-04-06
Provide snapshots for both V1 and V2 Catalogue API


jtweed:

jtweed:
We need to remember to add the V2 snapshots to the docs.

Issue #1840, opened 2018-04-06
V1 Catalogue API with updatable Miro data


jtweed:
So the general gist is that we’re going to have two indexes, and the API will decide which index to query on a per-request basis.

API:

- [x] Modify the API prefix to be name-only, and split traffic on version
- [x] Fork the WorksController and underlying models for v1 and v2 (#1887)
- [x] Fork the SwaggerController for v1 and v2 (#1887)
- [x] Make the API read two different indexes based on the version in the path (#1899)

Ingestor:

- [x] Add logic to the ingestor to decide if a work goes to v1 or v1+v2 (#1907)
- [x] ~Fork the Elasticsearch mapping,~ add logic to the ingestor to create both mappings on startup (#1907)
- [x] Modify the ingestor to write to as many indexes as necessary (#1907)

Snapshot pipeline:

- [x] Configure the snapshot scheduler to schedule snapshots of both indexes (#1895)
- [x] Modify the elasticdump container to send an SNS message upon completion (#1895)
- [x] Delete the S3 bucket notifications and the snapshot_convertor_job_generator Lambda (#1895)
- [x] Modify the snapshot_convertor to convert something into a V1 or V2 snapshot (#1906)

alicefuzier:
I don't think we need to fork the elasticsearch mapping, because we are not versioning the internal model, only the DisplayModels

alexwlchan:
Oops. 🙈

PR #1895, opened 2018-04-16
Modify the elasticdump task to send a job to SNS directly, not via S3

alexwlchan:
Part of #1840 / #1842. Rather than going

```
[ elasticdump ] ~> [ S3 ] ~> [ event stream ] ~> [ job generator ] ~> [ snapshot convertor ]
```

now we go

```
[ elasticdump ] ~> [ S3 ]
                ~> [ SNS ] ~> [ snapshot convertor ]
```

This is laying the groundwork for us to take continually updated snapshots of both V1 and V2 indexes at the same time.


alexwlchan:
This is based on #1897.

alexwlchan:
This patch is just adding the ability for the elastic dump container to send the notification; I’ll remove the snapshot_job_generator in a separate patch.

PR #1897, opened 2018-04-16
data_api: Identify buckets/keys in the pipeline as "private/public", not "source/target"

alexwlchan:

Using the words "source"/"target" gets confusing in this stack, because those terms are relative to each application.

Identifying these buckets/keys as "private/public" is closer to the names we use for the associated S3 buckets, and are absolute terms, so it's easier to reason about what "public bucket" means in any given context.

### Who is this change for?

📛 💅 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


alexwlchan:
I haven’t run the Terraform change yet, but I’ve run a plan, so I know I’m not missing a variable somewhere.

PR #1898, opened 2018-04-16
Move the Loris cluster into a dedicated VPC

### LINKED ISSUE #1884: Give Loris its own VPC
Currently Loris is sharing a VPC with the API. Since we asked for an increase in VPCs recently, we should give Loris its own VPC and break this dependency between the two stacks.

alexwlchan:
Resolves #1884. This change has already been applied.


alexwlchan:
😭

PR #1887, opened 2018-04-13
V2 api

alicefuzier:
### What is this PR trying to achieve?
Have a v2 prefix in the api that serialises works using v2 display models. Depends on #1889 
Still TODO:
- ~Return correct error on wrong version( do we need to ?)~ For another PR
- [x] Add Swagger test to check that models returned are correct
- [x] Move ApiVersions object to a more suitable package
### Who is this change for?
🚨 💅

PR #1899, opened 2018-04-16
Use a different ES index for v1/v2 apis

kenoir:

For: https://github.com/wellcometrust/platform/issues/1840

Allows different versions of the API to use different ES indexes so they can serve different datasets I.e. (sierra inclusive or exclusive).

### Who is this change for?

🍎 🐍 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/V1WorksController.scala
```diff
@@ -0,0 +1,33 @@
...
+  @Flag("es.index.v1") indexName: String,
+  worksService: WorksService)
+    extends WorksController(
+      apiPrefix = apiPrefix,
+      apiContextSuffix = apiContextSuffix,
+      apiHost = apiHost,
+      apiScheme = apiScheme,
+      indexName = indexName,
+      defaultPageSize = defaultPageSize,
+      worksService = worksService
```
alexwlchan: Having all this duplication seems a bit icky. Could we use Guice injection to push some of these parameters down into WorksController, or does it not work that way?

alicefuzier: This is actually already in master as it was merged as part of #1887 (the diff is a bit screwed because this is branched of that one and I think it needs a rebase). ANyway, no Guice can't inject arguments into abstract classes as far as I know cause they can't be instantiatetd

alicefuzier:
We forgot to add a test to check that v1 api doesn't return items from the v2 index and viceversa

catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/works/v2/ApiV2WorksTest.scala
```diff
@@ -11,7 +11,7 @@ class ApiV2WorksTest extends ApiWorksTestBase {
 
   it("returns a list of works") {
     withV2Api {
-      case (apiPrefix, indexName, itemType, server: EmbeddedHttpServer) =>
+      case (apiPrefix, _, indexName, itemType, server: EmbeddedHttpServer) =>
```
alexwlchan: Let’s fix it in a separate patch, but this seems like quite a fragile way to define the fixtures. Would be easy for somebody to get the wrong index by accident.

kenoir: I agree this is fragile. If i'm going to add another test case I might as well rename these.

alicefuzier:
Oh I've approved, but since we changed the es flag, we need to change the template config file and the terraform config as well. Definitely the change for the template config file should go in this PR and possibly the terraform changes as well

PR #1901, opened 2018-04-17
Increase the number of retries in the Sierra bibs pipeline

alexwlchan:
Both the bib windows and bib merger queues are getting DLQ entries fairly consistently. Redriving usually fixes the problem – it doesn’t seem to be a systemic error with those messages – so increase the number of times a message is automatically redriven.


alexwlchan:
This has passed the format task, so I’m calling it done and merging immediately.

Issue #1900, opened 2018-04-16
Error in “Post to Slack” caused by new alarms


alexwlchan:
```
event =
{'Records': [{'EventSource': 'aws:sns',
              'EventSubscriptionArn': 'arn:aws:sns:eu-west-1:760097843905:shared_alb_server_error_alarm:7701d498-64a4-425e-b966-f9069ecf91a2',
              'EventVersion': '1.0',
              'Sns': {'Message': '{"AlarmName":"loris-alb-not-enough-healthy-hosts","AlarmDescription":"This '
                                 'metric monitors '
                                 'loris-alb-not-enough-healthy-hosts","AWSAccountId":"760097843905","NewStateValue":"ALARM","NewStateReason":"Threshold '
                                 'Crossed: no datapoints were received for 1 '
                                 'period and 1 missing datapoint was treated '
                                 'as '
                                 '[Breaching].","StateChangeTime":"2018-04-16T15:34:56.438+0000","Region":"EU '
                                 '(Ireland)","OldStateValue":"INSUFFICIENT_DATA","Trigger":{"MetricName":"HealthyHostCount","Namespace":"AWS/ApplicationELB","StatisticType":"Statistic","Statistic":"SUM","Unit":null,"Dimensions":[{"name":"TargetGroup","value":"targetgroup/loris/087013874ebbe481"},{"name":"LoadBalancer","value":"app/loris/c0af06f31b54a8c9"}],"Period":60,"EvaluationPeriods":1,"ComparisonOperator":"LessThanThreshold","Threshold":2.0,"TreatMissingData":"- '
                                 'TreatMissingData: '
                                 'Breaching","EvaluateLowSampleCountPercentile":""}}',
                      'MessageAttributes': {},
                      'MessageId': '6bd3a539-6665-5449-9496-f677a285954e',
[...]

PR #1902, opened 2018-04-17
Fix post to slack bug when no timestamp

alexwlchan:
Quick bugfix in #1900. When there isn’t a timestamp in the message, we were trying to find relevant CloudWatch logs and throwing a TypeError. This is a quick patch which means it won’t be thrown again.

In practice this only occurs just after we’ve created new alarms, so it’s quite rare – otherwise I’d write a more sophisticated test.

### DIFF
--- a/monitoring/post_to_slack/src/post_to_slack.py
+++ b/monitoring/post_to_slack/src/post_to_slack.py
@@ -184,7 +184,7 @@ def prepare_slack_payload(alarm, bitly_access_token, sess=None):
     return slack_data
 
 
-def main(event, context):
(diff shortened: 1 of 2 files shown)


alexwlchan:
When an alarm is created with “missing data = breaching”, it immediately fires off an alarm in CloudWatch and triggers our Lambda.

That message doesn’t have a timestamp, but the Lambda tries to work out when it should be getting CloudWatch Logs for… and then it throws a TypeError because `None` isn't a valid timestamp.

Subsequent alarm notifications usually include values, so this is much less likely to recur after the initial creation.

PR #1905, opened 2018-04-17
Explicitly state the string representation of values in ApiVersionsEnum

alicefuzier:
Not run into a scala bug (probably) which makes enum serialisation unreliable by explicitly forcing the string representations of each value
### Who is this change for?
📒 🎁

### DIFF
--- a/common/src/main/scala/uk/ac/wellcome/versions/ApiVersions.scala
+++ b/common/src/main/scala/uk/ac/wellcome/versions/ApiVersions.scala
@@ -1,6 +1,7 @@
 package uk.ac.wellcome.versions
 
 object ApiVersions extends Enumeration {
-  val v1, v2 = Value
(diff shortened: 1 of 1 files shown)


common/src/main/scala/uk/ac/wellcome/versions/ApiVersions.scala
```diff
@@ -1,6 +1,10 @@
 package uk.ac.wellcome.versions
 
 object ApiVersions extends Enumeration {
-  val v1, v2 = Value
+  // We need to explicitly state the string representation of
+  // each value as we've seen nconsistent behaviour where v2 would be
```
alexwlchan: sp: "inconsistent"

alicefuzier: oops, thanks

PR #1906, opened 2018-04-17
Modify the snapshot convertor to write either V1 or V2 snapshots

alexwlchan:

Messages to the snapshot convertor include a new field "apiVersion", which is either V1 or V2. This is used to decide which of the DisplayWork models should be used to serialise the snapshot. Part of #1840, #1842.

### Who is this change for?

🐮 🍮 

- [ ] Deployed new versions

### DIFF
--- a/data_api/snapshot_convertor/src/test/scala/uk/ac/wellcome/platform/snapshot_convertor/services/ConvertorServiceTest.scala
+++ b/data_api/snapshot_convertor/src/test/scala/uk/ac/wellcome/platform/snapshot_convertor/services/ConvertorServiceTest.scala
@@ -93,7 +93,7 @@ class ConvertorServiceTest
             privateObjectKey = objectKey,
             publicBucketName = publicBucket.name,
             publicObjectKey = publicObjectKey,
-            modelVersion = ApiVersions.v1
+            apiVersion = ApiVersions.v1
           )
 
           val future = convertorService.runConversion(conversionJob)
@@ -151,7 +151,7 @@ class ConvertorServiceTest
             privateObjectKey = objectKey,
             publicBucketName = publicBucket.name,
             publicObjectKey = publicObjectKey,
-            modelVersion = ApiVersions.v2
+            apiVersion = ApiVersions.v2
           )
 
           val future = convertorService.runConversion(conversionJob)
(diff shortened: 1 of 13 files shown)


data_api/snapshot_convertor/src/test/scala/uk/ac/wellcome/platform/snapshot_convertor/services/ConvertorServiceTest.scala
```diff
@@ -243,7 +282,8 @@ class ConvertorServiceTest
           privateBucketName = privateBucket.name,
           privateObjectKey = "doesnotexist.txt.gz",
           publicBucketName = publicBucket.name,
-          publicObjectKey = "target.txt.gz"
+          publicObjectKey = "target.txt.gz",
```
kenoir: Should v1 and v2 not have different output names?

alexwlchan: Yeah, will fix – although I think the fixture creates a different bucket for each test, so we _should_ be protected.

alicefuzier: Are you concerned that they will blat each other work in production? The snapshot converter reads the bucket and the key to write to from the SQS message sent by elasticdump, which in turn reads it from the SQS message generated by the snapshot_scheduler lambda. The terraform config for that lambda is [here](https://github.com/wellcometrust/platform/blob/master/data_api/terraform/service_snapshot_scheduler.tf) and is already ensuring that the snapshot converter ends up generating two different files.

alexwlchan:
I’ve just rebased against master to get Travis working. As @alicefuzier said, I’m pretty sure we’re already protected against prod snapshots overwriting each other, and test fixtures protect us there.

Does that make sense @kenoir?

data_api/snapshot_convertor/src/test/scala/uk/ac/wellcome/platform/snapshot_convertor/services/ConvertorServiceTest.scala
kenoir: Yup. Thanks for clarifying.

PR #1909, opened 2018-04-18
Get rid of es_index as both apis now have two config flags for indexes

alicefuzier:
### Who is this change for?
👖 🎻


data_api/terraform/service_elasticdump.tf
```diff
@@ -19,10 +19,9 @@ module "elasticdump" {
     es_password   = "${local.es_password}"
     es_hostname   = "${local.es_name}.${local.es_region}.aws.found.io"
     es_port       = "${local.es_port}"
-    es_index      = "${local.es_index}"
   }
 
-  env_vars_length = 9
```
alexwlchan: Oh yeah, we need to change the elasticdump container to take v1/v2 flags, right?

alicefuzier: I don't think so because it reads the index from the message? The snapshot scheduler lambda is the thing that reads those variables I think?

alexwlchan: Oh, good point! I stand corrected.

PR #1922, opened 2018-04-19
Allow passing messages larger than SNS size cap.

### LINKED ISSUE #1776: We're exceeding SNS maximum message size

kenoir:

Supersedes: https://github.com/wellcometrust/platform/pull/1891

This change stores references to objects in S3 and hands around message pointers to them rather than handling the entire object, bypassing the size limit.

The `MessageWorker` class introduced here allows direct passing of objects serialisable with circe, and is intended to supersede SQSWorker where appropriate.

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


common/src/main/scala/uk/ac/wellcome/message/MessageWorker.scala
```diff
@@ -0,0 +1,62 @@
...
+
+import scala.concurrent.duration._
+import com.amazonaws.services.s3.AmazonS3
+import io.circe.Decoder
+import uk.ac.wellcome.s3.S3Uri
+import uk.ac.wellcome.s3.S3ObjectStore
+import uk.ac.wellcome.sqs.{SQSReader, SQSWorker}
+import uk.ac.wellcome.utils.JsonUtil._
+
+abstract class MessageWorker[T](sqsReader: SQSReader,
```
kenoir: @alicefuzier would appreciate your input into how to remove the inheritance dependency here.

PR #1923, opened 2018-04-19
[WIP] Allow passing messages larger than SNS size cap

### LINKED ISSUE #1776: We're exceeding SNS maximum message size

kenoir:

Supersedes: https://github.com/wellcometrust/platform/pull/1891

This change stores references to objects in S3 and hands around message pointers to them rather than handling the entire object, bypassing the size limit.

The `MessageWorker` class introduced here allows direct passing of objects serialisable with circe, and is intended to supersede SQSWorker where appropriate.

**TODO:**
- [ ] Consider options for removing inheritance requirement from `MessageWorker`
- [ ] Test including Worker & Writer to cover send and receive path in single test

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


alicefuzier:
> The MessageWorker class introduced here allows direct passing of objects serialisable with circe, and is intended to supersede SQSWorker where appropriate.

I don't see that class?
I only see the URI changes in S3ObjectStore which make total sense and I think could be a standalone PR.

common/src/main/scala/uk/ac/wellcome/sns/NotificationMessage.scala
```diff
@@ -0,0 +1,13 @@
+package uk.ac.wellcome.sns
+
+case class NotificationMessage(
```
kenoir: https://github.com/guardian/sns-event-stream/blob/master/app/sns/SubscriptionConfirmationMessage.scala#L42

@robertberry 👏

common/src/test/scala/uk/ac/wellcome/test/fixtures/Messaging.scala
```diff
@@ -31,16 +32,26 @@ trait Messaging
   def withSubscription[R](queue: Queue, topic: Topic)(
     testWith: TestWith[SubscribeResult, R]): R = {
 
-    info(s"Attempting to subscribe ${endpoint(queue)} to ${topic.arn}")
+    info(s"Attempting to subscribe $queue to $topic")
 
-    val subRequest = new SubscribeRequest(topic.arn, "sqs", endpoint(queue))
+    // HACK https://github.com/yourkarma/fake_sns/issues/2
```
kenoir: Is this the implementation used in localstack?

kenoir:
🎉 🎉 🎉 🎉 🎉 🎉 🎉

common/src/main/scala/uk/ac/wellcome/s3/S3ObjectStore.scala
```diff
@@ -57,20 +59,35 @@ class S3ObjectStore[T] @Inject()(
...
+  def put(sourcedObject: T)(implicit encoder: Encoder[T]): Future[URI] = {
     val keyPrefix = keyPrefixGenerator.generate(sourcedObject)
+
     S3ObjectStore.put[T](s3Client, s3Config.bucketName)(keyPrefix)(
       sourcedObject)
   }
 
-  def get(key: String)(implicit decoder: Decoder[T]): Future[T] = {
-    S3ObjectStore.get[T](s3Client, s3Config.bucketName)(key)
+  def get(uri: URI)(implicit decoder: Decoder[T]): Future[T] = {
```
alicefuzier: There probably is an obvious answer but, if this only works with `S3Uri`, why don't we enforce it in the type signature? Like `def get(s3Uri: S3Uri) ...`

common/src/test/scala/uk/ac/wellcome/message/MessagePointerTest.scala
```diff
@@ -0,0 +1,20 @@
...
+  it("reads from json") {
+    val pointer = fromJson[MessagePointer]("""{"src":"http://example.org"}""")
+    pointer shouldBe Success(MessagePointer("http://example.org"))
+  }
+
+  it("writes to json") {
+    val pointer = MessagePointer(new URI("http://example.org"))
+    toJson(pointer) shouldBe Success("""{"src":"http://example.org"}""")
+  }
+
```
alicefuzier: Maybe there could be a test for the failure case? Ie trying to read as Ur something that is not a uri

common/src/main/scala/uk/ac/wellcome/sns/NotificationMessage.scala
alicefuzier: This is a better equivalent of [this](https://github.com/wellcometrust/platform/blob/master/common/src/main/scala/uk/ac/wellcome/models/aws/SQSMessage.scala). Can we add a TODO there to remove it?

common/src/main/scala/uk/ac/wellcome/message/MessageReader.scala
```diff
@@ -0,0 +1,32 @@
+package uk.ac.wellcome.message
+
+import com.amazonaws.services.sqs.model.{Message => SQSMessage}
```
alicefuzier: Why do we rename the import? It's a bit confusing cause we have another `SQSMessage` class

common/src/main/scala/uk/ac/wellcome/s3/S3ObjectStore.scala
```diff
@@ -57,20 +59,35 @@ class S3ObjectStore[T] @Inject()(
...
+    uri match {
+      case S3Uri(bucket, key) => {
```
alicefuzier: Maybe a comment that this calls the unapply method on `S3Uri`. It's a bit obscure otherwise

common/src/main/scala/uk/ac/wellcome/message/MessageReader.scala
```diff
@@ -0,0 +1,32 @@
...
+import com.google.inject.Inject
+import uk.ac.wellcome.s3.S3ObjectStore
+
+import uk.ac.wellcome.utils.GlobalExecutionContext.context
+
+import scala.concurrent.Future
+
+class MessageReader[T] @Inject()(s3ObjectStore: S3ObjectStore[T]) {
+
+  def process(message: SQSMessage)(
```
alicefuzier: why is this function called `process`? Could we call it `read`?

common/src/test/scala/uk/ac/wellcome/message/MessageReaderTest.scala
```diff
@@ -0,0 +1,57 @@
...
+        val exampleMessage = new Message()
+          .withBody(serialisedExampleNotification)
+
+        val actualObjectFuture = messageReader.process(exampleMessage)
+
+        whenReady(actualObjectFuture) { actualObject =>
+          expectedObject shouldBe actualObject
+        }
+    }
+  }
```
alicefuzier: This test looks all right, but it would be nice to have tests for non happy path cases, like what happens if the pointer is not in s3 and what happens if the thing in s3 can't be deserialised

common/src/test/scala/uk/ac/wellcome/message/MessageWorkerTest.scala
```diff
@@ -0,0 +1,154 @@
...
+
+        s3Client.putObject(bucket.name, key, exampleObjectJson)
+
+        sqsClient.sendMessage(
+          queue.url,
+          toJson(exampleNotification).get
+        )
+
+        eventually {
+          verify(
```
alicefuzier: Sending metrics s not the primary feature of `MessageWorker`. The main thing is that it calls the `processMessage` function. Can we make a different assertion that reflects that?

common/src/test/scala/uk/ac/wellcome/message/MessageWriterTest.scala
```diff
@@ -0,0 +1,150 @@
...
+
+        val snsConfig = SNSConfig(topic.arn)
+        val snsWriter = new SNSWriter(snsClient, snsConfig)
+
+        val s3ObjectStore = new S3ObjectStore[ExampleObject](
+          s3Client,
+          s3Config,
+          keyPrefixGenerator)
+
+        val messages =
```
alicefuzier: messages?

common/src/test/scala/uk/ac/wellcome/message/MessageWorkerTest.scala
```diff
@@ -0,0 +1,154 @@
...
+        eventually {
+          verify(metrics)
+            .incrementCount(
+              matches(".*_MessageProcessingFailure"),
+              anyDouble())
+        }
+    }
+  }
+
+  it("reports an error when handling unsupported scheme") {
```
alicefuzier: Aren't this and the test above basically testing the same thing?

common/src/test/scala/uk/ac/wellcome/message/MessageWorkerTest.scala
kenoir: See `SQSReaderTest` for an example of this.

common/src/test/scala/uk/ac/wellcome/message/MessageWorkerTest.scala
```diff
@@ -0,0 +1,154 @@
...
+        eventually {
+          verify(metrics)
+            .incrementCount(
+              matches(".*_MessageProcessingFailure"),
+              anyDouble())
+        }
+    }
+  }
+
+  it("reports an error when unable to parse a message") {
```
alicefuzier: And this one

common/src/test/scala/uk/ac/wellcome/message/MessageWriterTest.scala
```diff
@@ -0,0 +1,150 @@
...
+
+        whenReady(eventualAttempt) { pointer =>
+          val messages = listMessagesReceivedFromSNS(topic)
+          messages should have size (1)
+          messages.head.subject shouldBe subject
+
+          val pointer = fromJson[MessagePointer](messages.head.message)
+
+          inside(pointer) {
+            case Success(MessagePointer(S3Uri(bucketName, _))) =>
```
alicefuzier: Does it fail the test if the try is not a `Success`? Have you tried it?

common/src/test/scala/uk/ac/wellcome/message/MessagingIntegrationTest.scala
```diff
@@ -0,0 +1,71 @@
...
+            val snsConfig = SNSConfig(topic.arn)
+            val snsWriter = new SNSWriter(localStackSnsClient, snsConfig)
+
+            val s3ObjectStore = new S3ObjectStore[ExampleObject](
+              s3Client,
+              s3Config,
+              keyPrefixGenerator
+            )
+
+            val messages =
```
alicefuzier: messages?

common/src/test/scala/uk/ac/wellcome/message/MessagingIntegrationTest.scala
```diff
@@ -0,0 +1,71 @@
...
+                snsWriter,
+                s3Config,
+                s3ObjectStore)
+
+            val exampleObject = ExampleObject("some value")
+
+            messages.write(exampleObject, "subject")
+
+            eventually {
+              verify(
```
alicefuzier: To have a real integration test we should be asserting that what the message reader gets is the same thing that was sent by the message writer

common/src/test/scala/uk/ac/wellcome/test/fixtures/Messaging.scala
```diff
@@ -0,0 +1,118 @@
...
+    )
+
+  case class ExampleObject(name: String)
+
+  val keyPrefixGenerator: KeyPrefixGenerator[ExampleObject] =
+    new KeyPrefixGenerator[ExampleObject] {
+      override def generate(obj: ExampleObject): String = "/"
+    }
+
+  def withMessageReader[R](s3Client: AmazonS3)(bucket: Bucket)(
```
alicefuzier: Why two different argument lists?

common/src/test/scala/uk/ac/wellcome/test/fixtures/Messaging.scala
```diff
@@ -0,0 +1,118 @@
...
+    testWith(testReader)
+  }
+
+  def withMessageWorker[R](
+    sqsClient: AmazonSQS,
+    s3Client: AmazonS3
+  )(actors: ActorSystem,
+    queue: Queue,
+    metrics: MetricsSender,
+    bucket: S3.Bucket)(testWith: TestWith[MessageWorker[ExampleObject], R]) = {
```
alicefuzier: And here?

common/src/test/scala/uk/ac/wellcome/message/MessageWriterTest.scala
```diff
@@ -0,0 +1,150 @@
...
+              bucketName shouldBe bucket.name
+          }
+
+          getContentFromS3(bucket)
```
alicefuzier: Why don't we use the key to get the object instead of getting all of the objects in the bucket?

common/src/main/scala/uk/ac/wellcome/s3/S3ObjectStore.scala
```diff
@@ -57,20 +59,36 @@ class S3ObjectStore[T] @Inject()(
...
+      case S3ObjectLocation(bucket, key) => {
+
+        if (bucket != s3Config.bucketName) {
+          debug(
+            s"Bucket name in URI ($bucket) does not match configured bucket (${s3Config.bucketName})")
+        }
+
+        S3ObjectStore.get[T](s3Client, bucket)(key)
+      }
+      case _ =>
```
alicefuzier: This case is impossible now, isn't it?

common/src/main/scala/uk/ac/wellcome/models/aws/SQSMessage.scala
```diff
@@ -2,6 +2,7 @@ package uk.ac.wellcome.models.aws
 
 import io.circe.generic.extras.JsonKey
 
+// Deprecated in favour of uk.ac.wellcome.sns.NotificationMessage
```
alicefuzier: Better yet here is a (scala deprecated annotation)[https://www.scala-lang.org/api/2.11.8/index.html#scala.deprecated]. This will cause IntelliJ to winge every time we use this class

common/src/main/scala/uk/ac/wellcome/s3/S3ObjectLocation.scala
```diff
@@ -0,0 +1,20 @@
+package uk.ac.wellcome.s3
+
+import java.net.URI
+
+import scala.util.Try
+
+case class S3ObjectLocation(bucket: String, key: String) {
+  def toURI: URI = {
```
alicefuzier: I don't see this ever being used..

common/src/main/scala/uk/ac/wellcome/message/MessagePointer.scala
```diff
@@ -0,0 +1,17 @@
...
+import java.net.URI
+
+import uk.ac.wellcome.s3.S3ObjectLocation
+
+import scala.util.Try
+
+case class MessagePointer(src: S3ObjectLocation)
+
+object MessagePointer {
+  def create(uri: URI): Try[MessagePointer] =
```
alicefuzier: This is never used in production code or am I missing something?

common/src/main/scala/uk/ac/wellcome/utils/JsonUtil.scala
```diff
@@ -1,17 +1,43 @@
...
 import io.circe.parser.decode
 import io.circe.syntax._
-import io.circe.{Decoder, Encoder}
+import io.circe.{Decoder, DecodingFailure, Encoder, Json}
 import uk.ac.wellcome.exceptions.GracefulFailureException
 
 import scala.util.Try
 
-object JsonUtil extends AutoDerivation with TimeInstances with Logging {
+trait URIInstances {
```
alicefuzier: If URI is being replaced for `S3ObjectLocation` we can also get rid of this, I think

common/src/main/scala/uk/ac/wellcome/s3/S3ObjectLocation.scala
```diff
@@ -0,0 +1,20 @@
...
+    new URI(s"s3://$bucket/$key")
+  }
+}
+
+object S3ObjectLocation {
+  def create(uri: URI): Try[S3ObjectLocation] =
```
alicefuzier: This is also never used except for the `create` method in `MessagePointer` which is not used

common/src/test/scala/uk/ac/wellcome/message/MessagePointerTest.scala
```diff
@@ -0,0 +1,37 @@
...
+import org.scalatest._
+import uk.ac.wellcome.utils.JsonUtil._
+import java.net.URI
+
+import uk.ac.wellcome.exceptions.GracefulFailureException
+import uk.ac.wellcome.s3.S3ObjectLocation
+
+import scala.util.{Failure, Success}
+
+class MessagePointerTest extends FunSpec with Matchers {
```
alicefuzier: These are all testing creating a `MessagePointer` from a `URI` which I think is not necessary anymore

common/src/test/scala/uk/ac/wellcome/message/MessageReaderTest.scala
```diff
@@ -0,0 +1,123 @@
...
+
+          val serialisedExampleNotification = toJson(exampleNotification).get
+
+          val exampleMessage = new Message()
+            .withBody(serialisedExampleNotification)
+
+          val actualObjectFuture = messageReader.read(exampleMessage)
+
+          whenReady(actualObjectFuture) { actualObject =>
+            expectedObject shouldBe actualObject
```
alicefuzier: minor  but if you want a accurate error message from scala test in case of failure, they should be the other way round

common/src/test/scala/uk/ac/wellcome/message/MessageReaderTest.scala
```diff
@@ -0,0 +1,123 @@
...
+    it("fail gracefully when NotificationMessage cannot be deserialised") {
+      withMessageReaderFixtures {
+        case (bucket, messageReader) =>
+          val key = "key.json"
+          val expectedObject = ExampleObject("some value")
+          val serialisedExampleObject = toJson(expectedObject).get
+
+          s3Client.putObject(bucket.name, key, serialisedExampleObject)
+
+          val examplePointer =
```
alicefuzier: I don't think this is used?

common/src/test/scala/uk/ac/wellcome/message/MessageWriterTest.scala
```diff
@@ -0,0 +1,150 @@
...
+          messages.head.subject shouldBe subject
+
+          val pointer = fromJson[MessagePointer](messages.head.message)
+
+          inside(pointer) {
+            case Success(MessagePointer(S3ObjectLocation(bucketName, _))) =>
+              bucketName shouldBe bucket.name
+          }
+
+          getContentFromS3(bucket)
```
alicefuzier: This isn't correct. It's reading every object in the bucket, mapping them into a list of `ExampleObject` and asserting that one of them is `message`. We should retrieve the object we are interested in using bucketName and key

common/src/test/scala/uk/ac/wellcome/message/MessageWriterTest.scala
```diff
@@ -0,0 +1,150 @@
...
+        val eventualAttempt = messageWriter.write(message, subject)
+
+        whenReady(eventualAttempt) { pointer =>
+          val messages = listMessagesReceivedFromSNS(topic)
+          messages should have size (1)
+          messages.head.subject shouldBe subject
+
+          val pointer = fromJson[MessagePointer](messages.head.message)
+
+          inside(pointer) {
```
alicefuzier: I think it would be nicer to deconstruct without using inside. We need to access the bucketName and the key anyway to be able to retrieve the object and assert on it

common/src/test/scala/uk/ac/wellcome/message/MessageReaderTest.scala
kenoir: \/ `val serialisedExamplePointer = toJson(examplePointer).get`

common/src/test/scala/uk/ac/wellcome/message/MessagingIntegrationTest.scala
```diff
@@ -0,0 +1,55 @@
...
+                snsWriter,
+                s3Config,
+                s3ObjectStore)
+
+            val exampleObject = ExampleObject("some value")
+
+            messageWriter.write(exampleObject, "subject")
+
+            eventually {
+              worker.calledWith shouldBe Some(exampleObject)
```
alicefuzier: 🎉 🎉

common/src/test/scala/uk/ac/wellcome/test/fixtures/S3.scala
```diff
@@ -79,6 +79,12 @@ trait S3 extends Logging with Eventually with ImplicitLogging {
     IOUtils.toString(s3Client.getObject(bucket.name, key).getObjectContent)
   }
 
+  def getContentFromS3(bucket: Bucket): Seq[String] = {
```
alicefuzier: I think this can go away

PR #1915, opened 2018-04-18
Make snapshot-convertor become snapshot-generator

alicefuzier:
### What is this PR trying to achieve?
Get rid of elasticdump and let the snapshot convertor query the index directly.
This lets us:
* avoid using the filesystem to store temporary files as elasticdump currently does
* simplify our architecture
* get rid of a whole bunch of code
* get rid of intermediate dumps in the private bucket
### Who is this change for?
⛺️ 🍝

PR #1924, opened 2018-04-20
Check the cache directories in our .travis.yml file in CI

alexwlchan:
Prompted by [my comment on #1915](https://github.com/wellcometrust/platform/pull/1915#discussion_r182960124).

We cache all our Scala `target` directories in Travis – this adds a test that we're caching all the directories we want, and just as importantly, not caching anything old and stale!


alexwlchan:
Closing for now, as I don’t want to spend more time working on this.

PR #1917, opened 2018-04-19
Model changes for contributors instead of creators

alexwlchan:
This changes our internal model to use "contributors", and modifies the DisplayWork models to convert this into creators or contributors as appropriate.

This patch is _just_ the model changes. They’re substantial enough to warrant their own patch – the actual transformation will get a separate test.

Required for #1682.

PR #1925, opened 2018-04-20
Implement the new Sierra contributors transformation

alexwlchan:
Builds on #1917, required for #1682.

I think this is okay, but not double-checked yet; PR mostly to get CI started while I walk from the station to the office!


catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraContributorsTest.scala
```diff
@@ -0,0 +1,662 @@
...
+            MarcSubfield(tag = "0", content = "lcsh3384149")
+          )
+        )
+      )
+
+      assertTransformFails(varFields = varFields)
+    }
+  }
+
+  describe("Organisation") {
```
alicefuzier: This test class is a bit huge. Could we move the person tests into a `SierraContributorsPersonsTest` (or something like that) and do the same for organisations?

alexwlchan: For the sake of getting this merged, I’ll leave it for now – we can break it up if/when we’re next in this part of the codebase.

PR #1935, opened 2018-04-26
Migrate work related models into a separate package

alicefuzier:
Continue breaking up common an reduce the number of modules that depend on it
### Who is this change for?
🐝 🚿


build.sbt
```diff
@@ -2,62 +2,78 @@ import sbt.Keys._
...
 import java.io.File
 
-def doSharedSetup(project: Project, folder: String) =
+def doServiceSetup(project: Project, folder: String) =
   project
     .in(new File(folder))
     .dependsOn(common % "compile->compile;test->test")
     .settings(Common.settings: _*)
     .settings(Finatra.settings: _*)
+    //I don't hink we use this anymore?
```
kenoir: Shall we remove it in a future PR?

build.sbt
```diff
@@ -2,62 +2,78 @@ import sbt.Keys._
...
   .settings(Common.settings: _*)
   .settings(DockerCompose.settings: _*)
   .enablePlugins(DockerComposePlugin)
   .settings(libraryDependencies ++= Dependencies.commonDependencies)
 
-// We still inherit from the main common because we need the models it
-// contains (e.g. IdentifiedWork) to construct display models.
-lazy val common_display = doSharedSetup(project, "sbt_common/display")
+// It depends on common because it uses JsonUtil
+lazy val work_model = doSharedLibrarySetup(project, "sbt_common/work_model")
```
kenoir: I don't really like the naming scheme here, but it's not blocking to merge this PR.

Maybe something like:
- `uk.ac.wellcome.models.work.internal._`
- `uk.ac.wellcome.models.work.display._`

build.sbt
alicefuzier: 👍

PR #1939, opened 2018-04-26
Adds data science stack

kenoir:

Allow Data Science infrastructure to be cheaply and easily deployed and used.

This uses an instance of the [new asg module](https://github.com/wellcometrust/terraform-modules/tree/master/ec2/asg). 

In it's current state this will deploy an ASG with 0 p2.xlarge by default  running the default Amazon Deep Learning AMI on Ubuntu - when scaled up manually it will request spot instances max priced at 0.35$ and hour. The ASG will be automatically scaled down to 0 at 8pm UTC every evening.

### Who is this change for?

@harrisonpim 

- [ ] Run `terraform apply`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ include assets/Makefile
 include loris/Makefile
 include shared_infra/Makefile
 include data_api/Makefile
+include data_science/Makefile
 include catalogue_api/Makefile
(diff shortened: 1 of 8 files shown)


kenoir:
Relates to: https://github.com/wellcometrust/data-labs-infra/issues/2

PR #1944, opened 2018-04-27
Add new messaging libs between transformer and id_minter

kenoir:
### What is this PR trying to achieve?

Move to the new messaging libs that allow for larger messages.

### Who is this change for?

✉️ 🐛 

### Have the following been considered/are they needed?

- [ ] Deployed new versions
- [ ] Run `terraform apply`

PR #1945, opened 2018-04-30
Migrate id_minter - ingestor to communicate with MessagePointers

alicefuzier:
Part of #1776 . Fix id_minter ingestor boundary
### Who is this change for?
🍈 🍁


alicefuzier:
Based on #1944. Will need to merge that before this one

PR #1946, opened 2018-04-30
Add a script for toggling the data science ASG, and connecting to instances

alexwlchan:

A standalone piece related to #1639.

This adds a script which either “starts” or “stops” an autoscaling group. The ASG name is “discovered” by use of tags – currently it looks for a test group with `{'data_science': 'true'}`, but we can define other tags in Terraform and have this script search for them.

I’d like review from:

* @wellcometrust/platform-devs for code quality
* @harrisonpim for usability – is this what you want?

I’ve set up a test ASG `alex-test` for testing

### Who is this change for?

@harrisonpim.

### DIFF
--- a/data_science/scripts/asg_utils.py
+++ b/data_science/scripts/asg_utils.py
@@ -1,14 +1,16 @@
 # -*- encoding: utf-8
 
+import sys
+
 
 DATA_SCIENCE_ASG_TAGS = {
     'data_science': 'true',
 }
 
 
-def discover_asg_name(asg_client):
+def discover_data_science_asg(asg_client):
     """
-    Returns the name of the first autoscaling group whose tags exactly match
+    Return data about the first autoscaling group whose tags exactly match
     the supplied input.
     """
     # This API is paginated, but for now we assume we have less than 100 ASGs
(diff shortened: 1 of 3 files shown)


kenoir:
A couple of questions:
- Is it useful to containerise these operations and farm them out to make tasks?
- Could we bundle these two operations into a single one, at least for the connection step? Something like:
  - Check if the asg desired is already 1
  - Poll until an instance is connectable
  - Initiate the tunnel

alexwlchan:
We could turn them into Make tasks, but I’m assuming @harrisonpim has a local Python installation, and passing an SSH key into the Docker container is moderately tedious.

Polling until it’s up might be nice too, but I’ll leave it until e need it. Simplicity FTW!

alexwlchan:
Waiting for #1951 before merging.

data_science/scripts/asg_utils.py
```diff
@@ -0,0 +1,62 @@
+# -*- encoding: utf-8
+
+import sys
+
+
+DATA_SCIENCE_ASG_TAGS = {
+    'data_science': 'true',
```
kenoir: For our purposes this is `'name': 'jupyter'`

alexwlchan: Fixed.

data_science/scripts/asg_utils.py
```diff
@@ -0,0 +1,62 @@
+# -*- encoding: utf-8
+
+import sys
+
+
+DATA_SCIENCE_ASG_TAGS = {
+    'name': 'jupyter',
```
kenoir: 👍

PR #1947, opened 2018-04-30
Added Genres type and transformations from Miro and Sierra.

mklander:
Part of #1682.
Adding Genres type and updated transformations from Sierra and Miro

### DIFF
--- a/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraGenres.scala
+++ b/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraGenres.scala
@@ -1,0 +1,27 @@
+package uk.ac.wellcome.transformer.transformers.sierra
+
+import uk.ac.wellcome.models.work.internal.{Genre, Concept, Period, Place}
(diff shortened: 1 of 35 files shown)


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraGenres.scala
```diff
@@ -0,0 +1,28 @@
...
+import uk.ac.wellcome.models.work.internal.{Concept, Genre, Period, Place}
+import uk.ac.wellcome.transformer.source.SierraBibData
+
+trait SierraGenres extends MarcUtils {
+
+  def getGenres(bibData: SierraBibData): List[Genre] = {
+    getGenresForMarcTag(bibData, "655")
+  }
+
+  private def getGenresForMarcTag(bibData: SierraBibData, marcTag: String) = {
```
kenoir: We routinely add the described transformation as a comment here. See: https://github.com/wellcometrust/platform/blob/787ac7d009a7e611feb69288716bbc157dfcfd42/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/sierra/SierraDescription.scala#L25

PR #1948, opened 2018-04-30
Add docker compose tasks for scala libraries, not just services

alicefuzier:
Same as #1942.  I don't like having to navigate to the root of each project to start docker-compose
### Who is this change for?
🍼 🇹🇱


alexwlchan:
How do we feel about an `__sbt_lib_target_template` in functions.Makefile, that provides these for us?

PR #1952, opened 2018-04-30
Separate messaging from the rest of sbt-common

alexwlchan:
This PR moves everything related to inter-app messaging (SQS and SNS) into a separate common lib.

I rearranged the structure while I was here. Previously, this code was split across:

```
common/
    finatra/controllers/
    message/
    models/aws
    sqs/
    sns/
    test/fixtures
```

and now it’s slightly simpler:

```
messaging/
    message/
    sqs/
    sns/
    test/fixtures
```

with the models/controllers moving into `sns` or `sqs` as appropriate.

Most of our applications take a new dependency, but the API now contains slightly less code. \o/

I’d be astounded if all the tests pass on the first try, so WIP until I have something worth reviewing.


alexwlchan:
Huzzah! We’ve also been able to move `com.fasterxml.jackson.dataformat` out of the common dependencies, so the common classpath shrinks.

build.sbt
```diff
@@ -43,6 +43,10 @@ lazy val common_elasticsearch = doSharedLibrarySetup(project, "sbt_common/elasti
   .dependsOn(internal_model % "compile->compile;test->test")
   .settings(libraryDependencies ++= Dependencies.commonElasticsearchDependencies)
 
+lazy val common_messaging = doSharedLibrarySetup(project, "sbt_common/messaging")
+  .dependsOn(common % "compile->compile;test->test")
```
alicefuzier: Can we add a comment here to explain that this is because of the VHS dependency and to remove this once we take the VHS out?

project/Dependencies.scala
```diff
@@ -94,14 +95,16 @@ object Dependencies {
     "org.scalacheck" %% "scalacheck" % versions.scalaCheckVersion % "test",
     "javax.xml.bind" % "jaxb-api" % versions.jaxbVersion,
     "com.github.alexarchambault" %% "scalacheck-shapeless_1.13" % versions.scalaCheckShapelessVersion % "test"
-  ) ++ awsDependencies ++ akkaDependencies ++ dynamoDependencies ++ jacksonDependencies ++ circeDependencies
+  ) ++ awsDependencies ++ akkaDependencies ++ dynamoDependencies ++ circeDependencies
```
alicefuzier: 🎉

project/Dependencies.scala
```diff
@@ -94,14 +95,16 @@ object Dependencies {
...
 
   val pipelineModelDependencies = circeDependencies
 
   val commonDisplayDependencies: Seq[ModuleID] = swaggerDependencies
 
   val commonElasticsearchDependencies = commonDependencies ++ esDependencies
 
+  val commonMessagingDependencies = commonDependencies ++ jacksonDependencies
```
alicefuzier: Actually, why does messaging depend on jackson? I thought we always use circe for messaging..

alexwlchan: I’ll add a comment – we’re using jackson-databind to parse YAML in the SNS fixture.

alicefuzier: pff, I see

alicefuzier: Maybe we should be using [this](https://github.com/circe/circe-yaml) (Not in this PR, but link it in the comment?)

alexwlchan: Yep, standardising on Circe would be good. In the comment.

alexwlchan:
Last commit is just two comments, so merging without waiting for Travis again.

PR #1953, opened 2018-04-30
How hard can the recorder be?

alexwlchan:
Okay, so it turns out “a bit”.

This isn't passing compiling yet, but it illustrates the general idea. It's unable to find the DynamoFormat or a few other implicits (why?), but hopefully that's fixable. Add another test, and it's done. Easy, right? 😜


catalogue_pipeline/recorder/src/main/scala/uk/ac/wellcome/platform/recorder/models/RecorderWorkEntry.scala
```diff
@@ -0,0 +1,18 @@
+package uk.ac.wellcome.recorder.models
+
+import uk.ac.wellcome.models.Sourced
+import uk.ac.wellcome.models.work.internal.UnidentifiedWork
+
+case class RecorderWorkEntry(
```
alicefuzier: Why do we wrap the `UnidentifiedWork`? `UnidentifiedWork` already contains a `sourceIdentifier` field with the same information that this wrapping class is adding.

alexwlchan: VHS can only store classes that have the `Id` trait, and putting that on UnidentifiedWork feels a bit weird.

catalogue_pipeline/recorder/src/main/scala/uk/ac/wellcome/platform/recorder/services/RecorderWorkerService.scala
```diff
@@ -2,23 +2,39 @@ package uk.ac.wellcome.platform.recorder.services
...
+  versionedHybridStore: VersionedHybridStore[RecorderWorkEntry],
   reader: SQSReader,
   system: ActorSystem,
   metrics: MetricsSender
 ) extends SQSWorkerToDynamo[UnidentifiedWork](reader, system, metrics) {
 
-  override implicit val decoder = implicitly[Decoder[UnidentifiedWork]]
+  implicit val decoder = Decoder[RecorderWorkEntry]
+  implicit val encoder = Encoder[RecorderWorkEntry]
+  implicit val dynamoFormat = DynamoFormat[RecorderWorkEntry]
```
alicefuzier: If this is the line that it's whingeing about, you might need to do `import uk.ac.wellcome.dynamo._`

catalogue_pipeline/recorder/src/main/scala/uk/ac/wellcome/platform/recorder/models/RecorderWorkEntry.scala
alicefuzier: Ah right. Pff. I think we could loosen up that constraint because it's already using an implicit `IdGetter` almost everywhere to retrieve the id (Actually we would have to do that when we take the VHS out). So we could make the VHS not restrict to classes that extend `Id` and create a `IdGetter` for `UnidentifiedWork` but it would probably have a very similar logic to this so probably we wouldn't gain that much (other than loosing a wrapping case class)

alexwlchan: Okay, so maybe drop this for now, and instead I’ll look at making VHS not require Id?

alexwlchan: I'll have a poke at this a bit longer, because it would be nice to get the recorder done rather than do more refactoring.

I think not having the wrapper would be nice, but it's not a massive issue.

I will add a comment explaining why we have the wrapper.

catalogue_pipeline/recorder/src/main/scala/uk/ac/wellcome/platform/recorder/services/RecorderWorkerService.scala
```diff
@@ -2,23 +2,39 @@ package uk.ac.wellcome.platform.recorder.services
...
 import uk.ac.wellcome.metrics.MetricsSender
+import uk.ac.wellcome.models.SourceMetadata
 import uk.ac.wellcome.models.work.internal.UnidentifiedWork
+import uk.ac.wellcome.recorder.models.RecorderWorkEntry
+import uk.ac.wellcome.storage.VersionedHybridStore
 import uk.ac.wellcome.utils.JsonUtil._
 
 import scala.concurrent.Future
 
+case class EmptyMetadata()
```
kenoir: I don't think this is used?

alexwlchan:
🎉

PR #1954, opened 2018-04-30
Start Jupyter on startup

kenoir:

Start jupyter server on instance start.

This should enable an SSH tunnel to be opened by running:

```sh
ssh -i ~/.ssh/wellcomedigitalplatform -N -f -L localhost:8888:localhost:8888 ubuntu@host
```

### Who is this change for?

@harrisonpim 

- [X] Run `terraform apply`.

### DIFF
--- a/data_science/terraform/userdata.sh.tpl
+++ b/data_science/terraform/userdata.sh.tpl
@@ -9,7 +9,7 @@ cat << EOF > /home/${notebook_user}/.jupyter/jupyter_notebook_config.py
 
 c.NotebookApp.notebook_dir = u"/home/${notebook_user}/"
 c.NotebookApp.ip = "*"
-c.NotebookApp.port = "${notebook_port}"
+c.NotebookApp.port = ${notebook_port}
 c.NotebookApp.open_browser = False
 c.NotebookApp.password = u'${hashed_password}'
 
--- a/data_science/terraform/userdata.sh.tpl
+++ b/data_science/terraform/userdata.sh.tpl
@@ -1,4 +1,7 @@
-#!/bin/bash
+#!/usr/bin/env bash
+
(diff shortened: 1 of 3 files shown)


kenoir:
I've verified this works now (as does the scale down instruction). The scripts in https://github.com/wellcometrust/platform/pull/1946 will need to be modified slightly to work with the new tag scheme.

kenoir:
Pinging @alexwlchan ^

data_science/terraform/userdata.sh.tpl
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
alexwlchan: Let's use our standard bash boilerplate here

```bash
#!/usr/bin/env bash

set -o errexit
set -o nounset
```

data_science/terraform/userdata.sh.tpl
```diff
@@ -0,0 +1,19 @@
...
+c.NotebookApp.notebook_dir = u"/home/${notebook_user}/"
+c.NotebookApp.ip = "*"
+c.NotebookApp.port = ${notebook_port}
+c.NotebookApp.open_browser = False
+c.NotebookApp.password = u'${hashed_password}'
+
+EOF
+
+# Start notebook server
+runuser -l ${notebook_user} -c '/home/ubuntu/anaconda3/bin/jupyter notebook'
```
alexwlchan: Let’s use the long form of `-l` and `-c` for readability. i.e.,

```bash
runner --login ${notebook_user} --command '/home/ubuntu/anaconda3/bin/jupyter notebook'
```

alexwlchan:
It passed format checks, so merging to save us any more Travis build time.

PR #1955, opened 2018-05-01
Ignore changes to the data_science scripts directory in Travis

### LINKED ISSUE #1951: Teach Travis about the data_science stack
And why it should probably ignore changes.

alexwlchan:
Resolves #1951.

This should improve our Travis times as we continue to add stuff here – the tests for #1946 took nearly two hours of vCPU time, even though they didn’t affect any of the Scala apps.

Simply ignoring the entire directory is a cheap heuristic, but it’ll do for now. If/when we add proper tests in this directory, we can revisit these rules.

### DIFF
--- a/travistooling/decisionmaker.py
+++ b/travistooling/decisionmaker.py
@@ -51,7 +51,7 @@ def does_file_affect_build_task(path, task):
         'LICENSE',
         '.travis.yml',
         'run_travis_task.py',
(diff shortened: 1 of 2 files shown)


alexwlchan:
Merging by fiat as this is fairly simple, and I’d like to speed up Travis *now*.

PR #1960, opened 2018-05-01
Data Science Infra v2

kenoir:

Added preinstalled modules:

```
pillow
seaborn
scikit-learn
tqdm
```

Also using and configuring: https://github.com/danielfrg/s3contents

Access to read all S3 buckets including write access to a data science specific bucket used by `s3contents`

Multiple instance types can be specified when starting instances.

### Who is this change for?

@harrisonpim 

- [ ] Run `terraform apply`.


data_science/scripts/asg_utils.py
```diff
@@ -17,34 +12,34 @@ def discover_data_science_asg(asg_client):
...
         actual_tags = {t['Key']: t['Value'] for t in asg_data['Tags']}
 
-        if actual_tags == DATA_SCIENCE_ASG_TAGS:
+        if ('name' in actual_tags) and (actual_tags['name'] == tag_name):
             return asg_data
 
-    else:
-        sys.exit("Can't find an ASG with tags %r!" % DATA_SCIENCE_ASG_TAGS)
+
+    sys.exit("Can't find an ASG with tag: %r!" % tag_name)
```
alexwlchan: If we're checking for name, we should change this error message:

```python
sys.exit("Can't find an ASG with name %r!" % tag_name)
```

data_science/scripts/asg_utils.py
```diff
@@ -17,34 +12,34 @@ def discover_data_science_asg(asg_client):
...
+    #    'ResourceType': 'auto-scaling-group',
+    #    'Value': '<VALUE>'}]
+    #
+    # We only care about the tag values, so we extract them into a
+    # Python dict.
+    for asg_data in resp['AutoScalingGroups']:
         actual_tags = {t['Key']: t['Value'] for t in asg_data['Tags']}
 
-        if actual_tags == DATA_SCIENCE_ASG_TAGS:
+        if ('name' in actual_tags) and (actual_tags['name'] == tag_name):
```
alexwlchan: Neater: assuming `name` will be not-None (add an assert for this if you want), you can do:

```python
if actual_tags.get(name) == tag_name:
```

which gets the value from the dict if present, or None if not.

data_science/scripts/asg_utils.py
```diff
@@ -3,12 +3,7 @@
 import sys
 
 
-DATA_SCIENCE_ASG_TAGS = {
-    'name': 'jupyter',
-}
-
-
-def discover_data_science_asg(asg_client):
+def discover_asg(asg_client, tag_name):
```
alexwlchan: If you're changing the method signature, you should change the docstring to match.

data_science/scripts/create_tunnel_to_asg.py
```diff
@@ -3,35 +3,48 @@
...
+Usage: create_tunnel_to_asg.py [--key=<KEY>] [--port=<PORT>] [--type=<INSTANCE_TYPE>]
 
 Actions:
-  --key=<KEY>     Path to an SSH key with access to the instances in the ASG.
-  --port=<PORT>   Local port to use for the remote Jupyter notebook
-                  (default: 8888).
+  --key=<KEY>             Path to an SSH key with access to the instances in the ASG.
+  --port=<PORT>           Local port to use for the remote Jupyter notebook
+                          (default: 8888).
+  --type=<INSTANCE_TYPE>  AWS Instance type (valid values: p2,t2) defaults to t2
```
alexwlchan: If you modify the help string to be:

```
--type=(p2 | t2)
```

I think docopt will check you only pass one of these values.

data_science/scripts/create_tunnel_to_asg.py
```diff
@@ -3,35 +3,48 @@
...
+from os.path import expanduser
+
 import boto3
 import docopt
 
 from asg_utils import discover_data_science_asg
 
 
+def _default_ssh_key_path():
+    return '%s/.ssh/wellcomedigitalplatform' % expanduser("~")
```
alexwlchan: FWIW, I'd normally use `os.environ['HOME']`, but I'm not sure if there's a difference.

data_science/scripts/create_tunnel_to_asg.py
```diff
@@ -3,35 +3,48 @@
...
+
+
 if __name__ == '__main__':
     args = docopt.docopt(__doc__)
 
-    key_path = os.path.abspath(args['--key'])
+    key_path = args['--key'] or _default_ssh_key_path()
+
+    print(key_path)
```
alexwlchan: Better: print a complete sentence that explains the significance of this path! e.g.

```python
print('Using SSH key at path %r' % path)
```

data_science/terraform/dlami_asg/userdata.sh.tpl
```diff
@@ -0,0 +1,44 @@
...
+            seaborn \
+            scikit-learn \
+            tqdm
+
+pip install jupyter \
+            boto3 \
+            requests \
+            s3fs==0.1.2 \
+            mock \
+            nose
```
alexwlchan: If we just install the latest version of every package, at some point one of these packages will break and we'll have a sad Harrison. It would be better to install a pinned set of packages that changes _only at our convenience_.

This is why we use pip-tools everywhere else: https://github.com/jazzband/pip-tools

It would be nice to do the same here (although how we get a requirements.txt into this script I'm not sure).

data_science/terraform/dlami_asg/variables.tf
```diff
@@ -0,0 +1,27 @@
+variable "key_name" {}
+variable "vpc_id" {}
+
+variable "hashed_password" {
+  default = "sha1:5310f21e370d:a4d66e725c179218638c21c03d83933aa066db2d"
```
alexwlchan: It would be useful to have a comment explaining how to generate this variable, _or_ declare the plaintext password as a Terraform variable, and do the SHA1 hashing at plan-time.

data_science/terraform/userdata.sh.tpl
```diff
@@ -18,5 +26,19 @@ c.NotebookApp.password = u'${hashed_password}'
...
+            tqdm
+
+pip install jupyter \
+            boto3 \
+            requests \
+            s3fs==0.1.2 \
+            mock \
+            nose
+
+pip install s3contents
```
alexwlchan: Why do we have two copies of our userdata? 😱

data_science/scripts/asg_utils.py
kenoir: I did wonder if there were a safe way to do this 👍

data_science/scripts/create_tunnel_to_asg.py
kenoir: I think the difference might be `expanduser` will work on windows?

data_science/terraform/dlami_asg/userdata.sh.tpl
kenoir: Yup, this approach using userdata to run a bootstrap script is tricky as soon as you want to do anything complex, so we should probably just pin them here. I want to avoid building our own AMI as long as I can because the dev loop on that is tedious (as you have to wait for the ami to build in AWS for every change).

data_science/terraform/dlami_asg/userdata.sh.tpl
```diff
@@ -18,5 +26,16 @@ c.NotebookApp.password = u'${hashed_password}'
...
+# Upgrade and install latest versions of modules
+/home/ubuntu/anaconda3/bin/pip install --upgrade pip
+/home/ubuntu/anaconda3/bin/pip install \
+            pillow==5.1.0 \
+            seaborn==0.8.1 \
+            scikit-learn==0.19.1 \
+            tqdm==4.19.7
+
+# Install s3contents
+/home/ubuntu/anaconda3/bin/pip install s3contents==0.2.2
```
alexwlchan: Better, but we’re only pinning the top-level packages.

e.g. pillow depends on olefile, but we haven't pinned the version of the latter

Yes, packaging sucks. Sorry. 😕

PR #1961, opened 2018-05-01
[WIP] Implement new config models for messaging

alexwlchan:
With the new messaging stack, some of our applications need to read from multiple buckets.

This PR adds a new config class – `MessageConfig` – which is to be used in place of SNSConfig and S3Config for the messaging stack.

A separate PR will add VHSConfig.

### DIFF
--- a/sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/message/MessageConfig.scala
+++ b/sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/message/MessageConfig.scala
@@ -1,0 +1,9 @@
+package uk.ac.wellcome.messaging.message
+
+import uk.ac.wellcome.messaging.sns.SNSConfig
+import uk.ac.wellcome.models.aws.S3Config
+
+case class MessageConfig(
+  snsConfig: SNSConfig,
+  s3Config: S3Config
+)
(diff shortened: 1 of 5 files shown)


alexwlchan:
Before we merge this, we need to think carefully about what the new injectors look like, because I’m not convinced it’s as simple as I first thought.

sbt_common/messaging/src/test/scala/uk/ac/wellcome/messaging/test/fixtures/Messaging.scala
```diff
@@ -115,22 +124,83 @@ trait Messaging
...
+                metricsSender,
+                queue,
+                messageReader) { worker =>
+                testWith((metricsSender, queue, bucket, worker))
+              }
+          }
+        }
+      }
+    }
+  }
```
alexwlchan: This takes up more lines that it used to, but I broke the old code and I couldn’t work out how to fix it without just guessing.

This version is more verbose, but substantially easier to read (for me, at least!).

sbt_common/messaging/src/test/scala/uk/ac/wellcome/messaging/test/fixtures/Messaging.scala
```diff
@@ -115,22 +124,83 @@ trait Messaging
...
+    withLocalS3Bucket { bucket =>
+      withLocalSnsTopic { topic =>
+        withMessageReader(bucket = bucket, topic = topic) { reader =>
+          testWith((bucket, reader))
+        }
+      }
+    }
+  }
 
-  def withMessageWorkerFixturesAndMockedMetrics[R] =
```
kenoir: 👍 ✂️

sbt_common/messaging/src/test/scala/uk/ac/wellcome/messaging/test/fixtures/Messaging.scala
kenoir: And _understand_ for all of us I think.

PR #1963, opened 2018-05-02
Separate our storage components into a separate library

alexwlchan:
A la #1952. I bet tests will fail and I haven't gone through the dependency list yet, but this should let us shrink the classpath of sbt-common.

It also gets most of the VHS pieces out of the common lib, which should make them easier to extract later if we want.


alexwlchan:
Okay, now I’ve chased all the compiler errors, this is what the new structure looks like:

```
storage/
    dynamo/
    s3/
    test/fixtures/
    type_classes/
    vhs/
````

PR #1964, opened 2018-05-02
Try to fix the flaky Elasticsearch test

### LINKED ISSUE #1962: elasticsearch test in CI is flaky
Log attached from a recent flaky run.

[log.txt](https://github.com/wellcometrust/platform/files/1966118/log.txt)

alexwlchan:
The issue was in the test that "updates an already existing index". We get the occasional 404 "index doesn't exist" error.

The index it uses has the same name as the other tests.  I think we were hitting a race condition where the index was being created in this test, then deleted by *another* test before we could use it a second time.

An attempt to resolve #1962.


kenoir:
I think there are some missing references to the old `withLocalElasticsearchIndex` that are causing this to fail.

PR #1967, opened 2018-05-02
Add new modules for VHS and VHS configuration

alexwlchan:
The spiritual successor to #1961 and #1963. Now you inject an instance of the VHS directly, rather than the individual pieces. I had to apply it to the sierra bib/item mergers, which are the only applications currently using VHS.

~This patch got more involved than I was intending, and I might break it apart further for easier review.~ (I did.)

### DIFF
--- a/sierra_adapter/sierra_bib_merger/src/test/scala/uk/ac/wellcome/platform/sierra_bib_merger/SierraBibMergerFeatureTest.scala
+++ b/sierra_adapter/sierra_bib_merger/src/test/scala/uk/ac/wellcome/platform/sierra_bib_merger/SierraBibMergerFeatureTest.scala
@@ -68,8 +68,8 @@ class SierraBibMergerFeatureTest
     withLocalSqsQueue { queue =>
       withLocalS3Bucket { bucket =>
         withLocalDynamoDbTable { table =>
-          val flags = sqsLocalFlags(queue) ++ s3LocalFlags(bucket) ++ dynamoDbLocalEndpointFlags(
-            table)
+          val flags = sqsLocalFlags(queue) ++ s3LocalFlags(bucket, namespace = "vhs") ++ dynamoDbLocalEndpointFlags(
+            table, namespace = "vhs")
           withServer(flags) { _ =>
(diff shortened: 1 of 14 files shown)


sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/vhs/VersionedHybridStoreModule.scala
```diff
@@ -0,0 +1,37 @@
...
+  def providesVHS[T <: Id](
+    awsConfig: AWSConfig,
+    vhsConfig: VHSConfig,
+    keyPrefixGenerator: KeyPrefixGenerator[T]): VersionedHybridStore[T] = {
+    val s3Client = S3ClientModule.providesS3Client(
+      awsConfig = awsConfig,
+      s3Config = vhsConfig.s3Config
+    )
+
+    val dynamoDbClient = DynamoClientModule.providesDynamoClient(
```
alexwlchan: Is this pattern sensible, or is this a bit left field?

sierra_adapter/sierra_bib_merger/src/test/scala/uk/ac/wellcome/platform/sierra_bib_merger/SierraBibMergerFeatureTest.scala
```diff
@@ -279,8 +291,11 @@ class SierraBibMergerFeatureTest
     withLocalSqsQueue { queue =>
       withLocalS3Bucket { bucket =>
         withLocalDynamoDbTable { table =>
-          val flags = sqsLocalFlags(queue) ++ s3LocalFlags(bucket) ++ dynamoDbLocalEndpointFlags(
-            table)
+          val flags = sqsLocalFlags(queue) ++ s3LocalFlags(
+            bucket,
+            namespace = "vhs") ++ dynamoDbLocalEndpointFlags(
+            table,
+            namespace = "vhs")
```
alexwlchan: Better than namespacing would be to take a leaf out of #1944, and have a vhsLocalFlags instead.

PR #1972, opened 2018-05-03
Simplify ImplicitLogging

kenoir:

The `ImplicitLogging` trait appeared in a few places without it being clear what it was for. This change moves all references to the fixtures package object where it is used in the first case.

### Who is this change for?

🎲 💯

### DIFF
--- a/common/src/test/scala/uk/ac/wellcome/test/fixtures/package.scala
+++ b/common/src/test/scala/uk/ac/wellcome/test/fixtures/package.scala
@@ -5,7 +5,6 @@ import grizzled.slf4j.Logger
 
 import scala.util.Try
 
-
 package object fixtures extends Logging {
 
   type TestWith[T, R] = T => R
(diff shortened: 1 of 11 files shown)


common/src/test/scala/uk/ac/wellcome/test/fixtures/package.scala
```diff
@@ -4,26 +4,31 @@ import grizzled.slf4j.Logger
...
+  extends FixtureComposers
+    with ImplicitLogging {
 
   type TestWith[T, R] = T => R
 
   type Fixture[L, R] = TestWith[L, R] => R
 
-  def safeCleanup[L](resource: L)(f: L => Unit)(
-    implicit logger: Logger): Unit = {
+  def safeCleanup[L](resource: L)(f: L => Unit): Unit = {
```
alicefuzier: If you're not passing logger implicitly here (which I think it' nice to not do) you can just extend Logging directly instead of extending ImplicitLogging

alicefuzier: It means you can probably delete ImplicitLogging cause this was probably the reason for adding it in . the first place

PR #1974, opened 2018-05-03
Remove the FixtureComposer/s

kenoir:

Simplify the tests by removing the FixtureComposer/s.

### Who is this change for?

🐐 🚪


common/src/test/scala/uk/ac/wellcome/test/fixtures/package.scala
```diff
@@ -1,17 +1,18 @@
...
+import com.twitter.inject.Logging
 import grizzled.slf4j.Logger
 
 import scala.util.Try
 
-package object fixtures extends FixtureComposers {
+package object fixtures extends Logging {
 
-  type TestWith[T, R] = T => R
+  implicit val implicitLogger = logger
```
alicefuzier: I don't think you need this?

PR #1977, opened 2018-05-03
Only run travis tests for library changes if an app depends on that library

### LINKED ISSUE #1976: Travis should skip testing an SBT app if the only changes are to are a library it doesn't use
e.g. changes so `common/elasticsearch` shouldn't trigger an ID minter build

alexwlchan:

Resolves #1976.

If a build contains changes to library X, then a Scala app will be tested _if and only if_ it has a _direct_ dependency on X. The aim is to speed up builds that edit library code by only re-testing affected applications.

This is an incredibly crude heuristic – grepping for import statements in the application code. I looked at getting a dependency tree from sbt, but doing so is (1) slow and (2) I couldn't find a way to get the dependency-tree in a machine-readable format.

We still have the overnight cron jobs to catch any mistakes that slip through.

### Who is this change for?

🎩


kenoir:
Closing for now, but not deleting.

PR #1978, opened 2018-05-03
Remove SQSReader from worker

kenoir:

Simplify the `MessageWorker` by pushing the `SQSReader` inside the `MessageReader` class.

### Who is this change for?

🍔 

- [ ] Deployed new versions

### DIFF
--- a/catalogue_pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/steps/IdEmbedderTests.scala
+++ b/catalogue_pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/steps/IdEmbedderTests.scala
@@ -24,10 +24,6 @@ class IdEmbedderTests
     with MockitoSugar
     with JsonTestUtil
     with ExtendedPatience {
-  override implicit val patienceConfig: PatienceConfig = PatienceConfig(
-    timeout = scaled(Span(3, Seconds)),
-    interval = scaled(Span(50, Millis))
-  )
 
   private val metricsSender =
     new MetricsSender(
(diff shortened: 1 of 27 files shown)


catalogue_pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/steps/IdEmbedderTests.scala
```diff
@@ -23,7 +23,7 @@ class IdEmbedderTests
     with BeforeAndAfterEach
     with MockitoSugar
     with JsonTestUtil
-    with PatienceConfiguration {
+    with ExtendedPatience {
```
alexwlchan: If you're replacing PatienceConfiguration with ExtendedPatience, are you going to remove the patienceConfig below?

sbt_common/messaging/src/test/scala/uk/ac/wellcome/messaging/test/fixtures/Messaging.scala
```diff
@@ -206,22 +218,52 @@ trait Messaging
...
-              }
-          }
+            }
         }
+
       }
     }
   }
 
+  def assertQueueEmpty(queue: Queue) = {
```
alexwlchan: Would these be more usefully defined on the SQS fixture, so they're available to other tests?

alexwlchan:
This PR also seems to be splitting MessageConfig into MessageReaderConfig and MessageWriterConfig. Could we get those changes as a separate PR?

catalogue_pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/steps/IdEmbedderTests.scala
alicefuzier: oh thanks

sbt_common/messaging/src/test/scala/uk/ac/wellcome/messaging/test/fixtures/Messaging.scala
alicefuzier: yep, will move

alicefuzier:
Addressed review comments. As for splitting the PR, the commit history is in a state where it would be a bit of a hassle to split. Also, the reason for splitting the config class came from the changes to `MessageReader`, so I don't think they are totally independent changes

PR #1980, opened 2018-05-04
Use variable spot price module

kenoir:

Use the variable spot price version of the terraform module.

Depends: https://github.com/wellcometrust/terraform-modules/pull/68

**NOTE:** This will need to be updated with the correct ref after the module is released.

### Who is this change for?

🔬


kenoir:
@alexwlchan should be updated now.

PR #1985, opened 2018-05-06
[RFC] Remove shared-infra tests from Travis

alexwlchan:
The last non-Terraform change to this stack was in February:

```console
$ git log -1 (find shared_infra -type f | grep -v '\.tf$')
commit f3c1644f90c2cbc009ddac4787f72d452960716e
Author: Alex Chan <a.chan@wellcome.ac.uk>
Date:   Mon Feb 19 14:09:18 2018 +0000

    Makefile: pass the entire repo into the Terraform containers

    This means that Git info -- in particular `.git` -- is accessible when
    running inside the containers.
```

So as a short-term fix for our build problem, I think we can drop this for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -109,10 +109,10 @@ jobs:
     - env: TASK=sierra_item_merger-test
     - env: TASK=s3_demultiplexer-test
 
-    # Shared infra stack
-    - env: TASK=shared_infra-test
-
     # (not under active development)
+
+    # Shared infra stack
+    # - env: TASK=shared_infra-test
 
     # nginx stack
     # - env: TASK=nginx-build


kenoir:
👍 Also I like this as an RFC?

alicefuzier:
As a general rule, I think that if it has code that runs in production, it should be actively built. We will never remember to turn tests back on if we make changes. Also, some of the lambdas in here deal with pretty delicate bits so I think for the sake of saving 30 seconds, we should leave this in.

alexwlchan:
> if it has code that runs in production, it should be actively built

The nginx, loris, and monitoring stacks are already disabled in Travis, for broadly the same reasons as I proposed this PR. Should we turn those back on?

alicefuzier:
Maybe. The loris and nginx ones don't seem to have tests at all, so it might make sense to keep those turned off and turn on the publishing bit only when we need it. As for the monitoring, yes, I would argue that it should be turned on, although those lambdas are a bit less critical than the ones in shared_infra. Anyway, we do rely on them and assume they work. That assumption should be based on tests running for every change without us having to remember to turn them on

kenoir:
Closing as "not doing this thing right now".

PR #1991, opened 2018-05-08
Add type parameters to Subject and Genre

alexwlchan:
A first step towards #1959.

We want to be able to put identifiers on Concept – but unlike Agent, instances of Concept aren’t used at the top level of a work; they’re wrapped by Genre and Subject. This patch adds a type parameter to both types, laying the groundwork to wrap them in the appropriate IdentityState / MaybeDisplayable / Displayable wrappers (**which will be a separate PR**).

cf. Contributor and Agent

### Who is this change for?

Me, as I add identifiers to Concept.

### DIFF
--- a/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/MiroTransformableTransformer.scala
+++ b/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/MiroTransformableTransformer.scala
@@ -1,15 +1,16 @@
 package uk.ac.wellcome.transformer.transformers
 import uk.ac.wellcome.models.transformable.MiroTransformable
-import uk.ac.wellcome.models.work.internal
+
 import uk.ac.wellcome.models.work.internal._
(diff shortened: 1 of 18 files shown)


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/MiroTransformableTransformer.scala
```diff
@@ -215,46 +217,6 @@ class MiroTransformableTransformer
...
-  }
-
-  private def getGenres(miroData: MiroTransformableData): List[Genre] = {
-    // Populate the subjects field.  This is based on two fields in the XML,
-    // <image_phys_format> and <image_lc_genre>.
-    (miroData.physFormat.toList ++ miroData.lcGenre.toList).map { g =>
-      Genre(label = g, concepts = List(Concept(g)))
-    }.distinct
-  }
-
```
alexwlchan: Breaking up the MiroTransformableTransformer has been an open TODO for a while, so I snuck in a change to break out Genres and Subjects while I was here.

PR #1993, opened 2018-05-08
[RFC] Proposing an Archival Storage Service

kenoir:

An RFC for the proposed archival storage service handling large archival filesets.

### Who is this change for?

Archivists.


docs/rfcs/archival_storage_service.md
```diff
@@ -0,0 +1,59 @@
...
+![archival storage service - page 1](https://user-images.githubusercontent.com/953792/39753255-0987d2fe-52b6-11e8-97ee-633ac09e1a9e.png)
+
+### Process flow
+
+- Work is uploaded in gzipped BagIt format to S3 to an "Ingest Bucket"
+- The S3 event stream triggers an AWS Lambda, translating the event to an SNS notification
+- The "Archival Service" picks up work as an [SQS Autoscaling Service](https://github.com/wellcometrust/terraform-modules/tree/master/sqs_autoscaling_service) subscribed to the topic.
+- The Archival Service performs the following operations:
+  - Obtain a filestream of the uploaded compressed artefact from its ingest location
+  - Stream the decompressed artefact to a processing location
```
kenoir: I think practically this breaks down into a number of smaller services:
- Decompress
- Format check
- Move to long term storage

Though whether we choose to implement it that way or not is up for grabs.

kenoir:
It's likely that this solution will also need a UI to display failed accessions, probably fed from DLQs between services - possibly even a mechanism of retrying or extracting failed content.

kenoir:
In addition we'll need to consider integrations with other services such as:

- [archivematica](https://www.archivematica.org/en/)
- [goobi](https://www.intranda.com/en/digiverso/goobi/goobi-overview/)

Likely there will be a requirements for these services to provide accessions in the BagIt bag format, gzipped and uploaded to an S3 location.

docs/rfcs/archival_storage_service.md
```diff
@@ -0,0 +1,79 @@
...
+
+- Ensure safe long term storage assets of digitised assets
+- Follow industry practises around file integrity and audit trails
+- Provide a scalable mechinsm for indetifying and retrieving content
+
+## Suggested Solution
+
+We propose to build a storage service based on AWS S3 and DynamoDb. 
+
+![archival storage service - page 1](https://user-images.githubusercontent.com/953792/39753255-0987d2fe-52b6-11e8-97ee-633ac09e1a9e.png)
```
alexwlchan: minor: This PNG file should be checked in.

docs/rfcs/archival_storage_service.md
```diff
@@ -0,0 +1,79 @@
+# Archival Storage Service
+
+## Problem Statement
+
+In order to provide archivists with reliable long term digital storage of both digitised and "born-digital" assets we need a service that can: 
+
+- Ensure safe long term storage assets of digitised assets
+- Follow industry practises around file integrity and audit trails
+- Provide a scalable mechinsm for indetifying and retrieving content
```
alexwlchan: nit: mecinsm, indetifying

docs/rfcs/archival_storage_service.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Suggested Solution
+
+We propose to build a storage service based on AWS S3 and DynamoDb. 
+
+![archival storage service - page 1](https://user-images.githubusercontent.com/953792/39753255-0987d2fe-52b6-11e8-97ee-633ac09e1a9e.png)
+
+### Process flow
+
+- Work is uploaded in gzipped BagIt format to S3 to an "Ingest Bucket"
```
alexwlchan: What is BagIt format? A glossary might be helpful.

kenoir:
Resolves: https://github.com/wellcometrust/platform/issues/1818

docs/rfcs/archival_storage_service.md
kenoir: 👍 yup, occurred to me that would be preferable, i'm not sure how to reference checked in images from markdown.

docs/rfcs/archival_storage_service.md
kenoir: There is a `Terminology` section. I hacked together something after reading the RFC RFC: https://tools.ietf.org/html/rfc7990.

docs/rfcs/archival_storage_service.md
alexwlchan: If it’s in the same image as the Markdown file, the following syntax is usually sufficient:

```markdown
![alt text](name_of_file.png)
```

PR #1994, opened 2018-05-08
added properties to recorder init config

mklander:
Update VHS recorder configuration to support S3 message queues

### DIFF
--- a/catalogue_pipeline/recorder/src/universal/conf/application.ini.template
+++ b/catalogue_pipeline/recorder/src/universal/conf/application.ini.template
@@ -1,4 +1,5 @@
--aws.sqs.queue.url=${windows_queue_url}
+-aws.sqs.queue.url=${recorder_queue_id}
 -aws.metrics.namespace=${metrics_namespace}
--aws.s3.bucketName=${bucket_name}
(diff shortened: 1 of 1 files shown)


catalogue_pipeline/recorder/src/universal/conf/application.ini.template
```diff
@@ -1,4 +1,6 @@
--aws.sqs.queue.url=${windows_queue_url}
+-aws.sqs.queue.url=${recorder_queue_id}
 -aws.metrics.namespace=${metrics_namespace}
--aws.s3.bucketName=${bucket_name}
--aws.dynamo.tableName=${dynamo_table_name}
\ No newline at end of file
+-aws.message.sns.topic.arn=${sns_arn}
```
kenoir: This should be a queue url.

kenoir: `aws.message.sqs.queue.url`

kenoir: Strike that, no topic and make the reference to a queue as above.

PR #1996, opened 2018-05-08
Add identifier pieces to the Concept model

alexwlchan:

Now we a Concept model that takes a type parameter (#1991), and we've split off the V1 and V2 DisplayConcepts (#1995), we can modify the Concept model to support identifiers.

This patch only includes model changes – the logic to actually add identifiers will be another patch.

Another part of #1959.

### Who is this change for?

📴 🔢

### DIFF
--- a/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/miro/MiroGenres.scala
+++ b/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/miro/MiroGenres.scala
@@ -1,21 +1,15 @@
 package uk.ac.wellcome.transformer.transformers.miro
 
-import uk.ac.wellcome.models.work.internal.{
-  AbstractConcept,
-  Concept,
-  Genre,
-  MaybeDisplayable,
-  Unidentifiable
-}
+import uk.ac.wellcome.models.work.internal._
 import uk.ac.wellcome.transformer.source.MiroTransformableData
 
 trait MiroGenres {
   def getGenres(miroData: MiroTransformableData)
-    : List[Genre[MaybeDisplayable[AbstractConcept]]] = {
+    : List[Genre[Unidentifiable[Concept]]] = {
     // Populate the genres field.  This is based on two fields in the XML,
     // <image_phys_format> and <image_lc_genre>.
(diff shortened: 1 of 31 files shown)


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/miro/MiroGenres.scala
```diff
@@ -1,17 +1,23 @@
...
 
 trait MiroGenres {
   def getGenres(
-    miroData: MiroTransformableData): List[Genre[AbstractConcept]] = {
+    miroData: MiroTransformableData): List[Genre[MaybeDisplayable[AbstractConcept]]] = {
     // Populate the genres field.  This is based on two fields in the XML,
     // <image_phys_format> and <image_lc_genre>.
     (miroData.physFormat.toList ++ miroData.lcGenre.toList).map { label =>
-      Genre[AbstractConcept](
+      Genre[MaybeDisplayable[AbstractConcept]](
```
alicefuzier: Since it's never going to be anything other than `Unidentifiable` for miro, can we not type it as `Genre[Unidentifiable[AbstractConcept]]` here and above?

alicefuzier: Actually it could probably be `Genre[Unidentifiable[Concept]]` cause it's never going to be anything other than `Concept` either

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/transformer/transformers/miro/MiroSubjects.scala
```diff
@@ -16,32 +22,16 @@ trait MiroSubjects {
...
-            )
-          }
-        case None =>
-          List()
-      }
+    val keywordsUnauth: List[String] = miroData.keywordsUnauth.getOrElse(List())
 
-    keywords ++ keywordsUnauth
+    (keywords ++ keywordsUnauth).map { keyword =>
+      Subject[MaybeDisplayable[AbstractConcept]](
```
alicefuzier: Same here

alexwlchan:
Labelling this as WIP, it turns out this wasn't nearly as complete as I thought it was.

PR #1998, opened 2018-05-08
[RFC] Authentication for restricted assets

kenoir:

Provide a mechanism to restrict access to certain images in the collection based on their rights status.

### Who is this change for?

📖 🏹


jtweed:
I wonder if it makes sense to broaden this out beyond images. We'll have to also serve pdfs, videos and other binary file types in addition to images through Loris.

The origin for that could be the derivatives S3 bucket?

PR #2001, opened 2018-05-08
[RFC] METS Adapter

kenoir:

In order to provide information about the majority of digitised assets. We need to ingest METS files created by Goobi and merge their information into works. 

### Who is this change for?

Anyone who wants to see a digital representation of Works in the catalogue.


kenoir:
@jtweed yes, certainly.

Issue #1807, opened 2018-04-05
Implement a service to enable large scale analysis of source data


jtweed:

jtweed:
Prototype done, production version next quarter.

PR #2002, opened 2018-05-08
[RFC] Reporting & analytics pipeline

kenoir:

In order to do some analysis on the many sources ingested into the catalogue we need to provide the source data in a form that is easily searchable.

### Who is this change for?

@jtweed  and @harrisonpim, and anyone who wants to do "big data" analysis on Wellcome Collection data.


kenoir:
Part of https://github.com/wellcometrust/platform/issues/1807

docs/rfcs/analytics_pipeline.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Problem Statement
+
+In order to run analytics and reporting on data from various library sources we need to make source data available in a searchable format.
+
+## Proposed Solution
+
+We propose to add a simple analytics pipeline powered by lambda functions feeding an ElasticSearch cluster.
+
+![analytics pipeline - page 1 2](https://user-images.githubusercontent.com/953792/39773143-0fd21ed8-52ef-11e8-8c53-821f3358b761.png)
```
kenoir: Why not kinesis firehose?

jtweed: I think Kinesis Data Firehose is definitely worth considering for this. I've just had a look and I think it would require one stream per source, plus using Amazon ES rather than Elastic Cloud.

But it would also better align with other potential Kinesis uses in the future, e.g. clickstream analytics and more involved processing managed using Kinesis Data Streams.

PR #2008, opened 2018-05-09
[WIP] VHS to SNS

alicefuzier:
Write a scala lambda to transform `HybridRecord` messages from the dynamo event stream into `MessagePointer`, so that they can be read by the `MessageStream`
### Who is this change for?
💡 🤳

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -25,6 +25,7 @@ def doLambdaSetup(project: Project, folder: String) =
     .in(new File(folder))
     .settings(Common.settings: _*)
     .settings(DockerCompose.settings: _*)
+    .settings(retrieveManaged := true)
     .enablePlugins(DockerComposePlugin)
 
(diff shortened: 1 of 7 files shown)


kenoir:
**Closing as "maybe later":**

This PR represents an experiment to see if running Lambdas in AWS using Scala would be feasible for us. If successful it would allow us to look at moving other work into lambdas to reduce cost and complexity.

The issue with writing lambdas in Scala this way is that in order to run these as a JVM lambda in AWS the Scala Standard Library must be packaged alongside the application code in a ["fat JAR"](https://stackoverflow.com/questions/19150811/what-is-a-fat-jar). This results in an artefact that is (at time of writing) close to or exceeding the AWS lambda limit for the maximum size of an uncompressed JAR. 

See https://docs.aws.amazon.com/lambda/latest/dg/limits.html

PR #2018, opened 2018-05-10
Remove optics from the sbt-common classpath

alexwlchan:

Per the commit messages:

> **Move SierraRecord into the sierra_adapter common lib**
>
> This model is internal to the Sierra adapter pipeline -- by the time it reaches the VHS, it's been coerced into a BibRecord or an ItemRecord. It doesn't need to be available to the transformer or other applications.

> **Move optics out of the sbt-common classpath**
>
> Now that SierraRecord is out of sbt-common, we don't need optics everywhere.  Since it brings in a dependency on Scalaz and cats, there's likely a small benefit to getting it out of some of our applications.

Will this make a massive difference to build times? Unlikely. But it continues to reduce the footprint of sbt-common.

### Who is this change for?

🚢 🤖

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -53,7 +53,6 @@ object Dependencies {
     "io.circe" %% "circe-generic"% versions.circeVersion,
     "io.circe" %% "circe-generic-extras"% versions.circeVersion,
     "io.circe" %% "circe-parser"% versions.circeVersion,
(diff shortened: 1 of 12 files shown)


alexwlchan:
Okay, maybe circe depends on scalaz and cats in other ways. Still good tho!

PR #2020, opened 2018-05-10
[RFC] Matcher proposed architecture

alicefuzier:
Propose how the matching service should work


docs/rfcs/merger_architecture.md
```diff
@@ -71,29 +71,31 @@ This is our current schema:
...
 0.  The existing DB is as follows:
 
         src   | tail_v | component_v | is_redirect | redirected_target
-        A     | B      | ABC         | -           | -
-        B     | A      | ABC         | -           | -
-        C     | B      | ABC         | -           | -
-        D     | F      | DEF         | -           | -
-        E     | D      | DEF         | -           | -
-        F     | -      | DEF         | -           | -
+        A     | B      | ABC         | true        | ABC
```
alicefuzier: I took the liberty to modify some of the things that I think we agreed on a whiteboard session after this document was created, but feel free to scratch that if it doesn't make sense

alicefuzier:
Pff. Too many changes so it show the file as entirely new...How annoying

docs/merger_architecture.md
```diff
@@ -1,172 +0,0 @@
...
-![](relater_component_ids.png)
-
-This is our current schema:
-
--   `source_component` -- contains the identifier of a component.
-    This may be a single vertex.
-
--   `tail_vertices` -- if the source component is a single vertex, a list of
-    vertices which are the "tail" of a directed edge from this vertex.
-
```
mklander: Perhaps call these vertex head and tail, rather than source and tail.  Also, I'm not sure why this is a list of tail_vertices I thought this table was storing an edge per row?

docs/merger_architecture.md
```diff
@@ -1,172 +0,0 @@
...
-We decided to use SQL, specifically RDS, as we already use it in the ID minter, and we know how it scales.
-
-The idea is to store the structure of the graph in SQL, but do all of the graph theory logic outside the database.
-Trying to do graph operations in SQL queries feels a step too far!
-
-## Database schema
-
-We can identify subgraphs/components by concatenating the sorted IDs of the vertices.
-For example, the graph below has three components: `A-B-C-D`, `E-F-G` and `H`.
```
mklander: Is the fact that sorted vertex IDs are used to identify a graph used?  Otherwise perhaps a unique graph id could be easier.

docs/merger_architecture.md
alicefuzier: Uhm, yeah in hindsight that's not a correct description. It stores the single nodes and all the subgraphs that at some point have been advertised as a single work. I think it could work by storing an edge per row but it would just mean more operations when acquiring the locks, I'm not sure whether it would have any advantage

docs/merger_architecture.md
kenoir: I prefer the deterministic method here. Generating unique ids and keeping track of them is I think a harder problem.

PR #2021, opened 2018-05-10
Separate an SQSStream from MessageStream

alexwlchan:

Currently we have a MessageStream class, which streams message pointers from SQS and fetches the associated object from S3.

This PR pulls out the SQS streaming logic, so we can reuse it in other applications where we want to read from SQS with backpressure. It's mostly a copy-and-paste of code from MessageStream into a new class, with tests likewise copied and tweaked only slightly.

### Who is this change for?

Me, as I want to make the Sierra readers backpressure-aware.

### DIFF
--- a/sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/message/MessageStream.scala
+++ b/sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/message/MessageStream.scala
@@ -1,23 +1,16 @@
 package uk.ac.wellcome.messaging.message
 
-import akka.Done
 import akka.actor.ActorSystem
-import akka.stream.{ActorMaterializer, ActorMaterializerSettings, Supervision}
-import akka.stream.alpakka.sqs.MessageAction
-import akka.stream.alpakka.sqs.scaladsl.{SqsAckSink, SqsSource}
 import com.amazonaws.services.s3.AmazonS3
 import com.amazonaws.services.sqs
 import com.amazonaws.services.sqs.AmazonSQSAsync
-import com.amazonaws.services.sqs.model.Message
 import com.google.inject.Inject
-import grizzled.slf4j.Logging
 import io.circe.Decoder
-import uk.ac.wellcome.exceptions.GracefulFailureException
 import uk.ac.wellcome.messaging.sns.NotificationMessage
+import uk.ac.wellcome.messaging.sqs.SQSStream
 import uk.ac.wellcome.monitoring.MetricsSender
 import uk.ac.wellcome.storage.s3.S3ObjectStore
-import uk.ac.wellcome.utils.JsonUtil.fromJson
-import uk.ac.wellcome.utils.JsonUtil._
+import uk.ac.wellcome.utils.JsonUtil.{fromJson, _}
 
 import scala.concurrent.Future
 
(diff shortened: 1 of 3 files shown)


sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/sqs/SQSStream.scala
```diff
@@ -0,0 +1,73 @@
...
+        metricsSender.incrementCount(
+          metricName = s"${streamName}_MessageProcessingFailure",
+          count = 1.0
+        )
+    }
+    processMessageFuture
+  }
+
+  protected def read(message: sqs.model.Message)(implicit decoderT: Decoder[T]): Future[T] =
+    Future.fromTry(fromJson[T](message.getBody))
```
alexwlchan: This is a slight case of the tail wagging the dog – I know we need to override this with a method that returns a Future below, so it’s a Future here to avoid the complexity of modifying readAndProcess() to support a Future or a Try. In an ideal world it would just return a Try, but I don't think it's that important.

sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/message/MessageStream.scala
```diff
@@ -26,60 +19,19 @@ class MessageStream[T] @Inject()(actorSystem: ActorSystem,
...
-          1.0)
-    }
-    processMessageFuture
-  }
-
-  private def read(message: sqs.model.Message)(
-    implicit decoderN: Decoder[NotificationMessage],
-    decoderT: Decoder[T]
-  ): Future[T] = {
+  override def read(message: sqs.model.Message)(implicit decoderT: Decoder[T]): Future[T] = {
```
alicefuzier: Uhm, one of the things I really don't like in `MessageWorker` and `SqsWorker` is using polymorphism rather than composition. It makes it hard to follow how those functions are used. I'd rather not introduce the same pattern here.
Because there is no mutable state there should be a way of just passing functions around.

alexwlchan: So instead you have to pass a `read()` method into `foreach`? Is that the idea?

alicefuzier: To further extend what I mean, I think `MessageStream` should be using `SQSStream` rather than extending it. Its `foreach` function should internally call `SQSStream.foreach` passing it a function that goes from `NotificationMessage` to `Future[Unit]`. That function should be in turn be made of:
- decoding the `MessagePointer` and getting the object from s3
- decoding the object from s3 into T
- apply the `process` function supplied by the user in `MessageStream.foreach`

alexwlchan: Ah, got it. Yeah, that makes sense.

sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/sqs/SQSStream.scala
```diff
@@ -0,0 +1,74 @@
...
+        logger.error(s"Failure while processing message.", exception)
+        metricsSender.incrementCount(
+          metricName = s"${streamName}_MessageProcessingFailure",
+          count = 1.0
+        )
+    }
+    processMessageFuture
+  }
+
+  protected def read(message: sqs.model.Message)(implicit decoderT: Decoder[T]): Try[T] =
```
alicefuzier: This can be private now

alicefuzier:
Yep, I like this 👍

PR #2023, opened 2018-05-10
Use first version of finatra that supports scala 2.12

alicefuzier:
Take us a step closer to upgrade scala
### Who is this change for?
💅 🎪


kenoir:
🐳 🎉

PR #2024, opened 2018-05-10
Share AWS builder methods between client modules and fixtures; take another step away from Finatra

alexwlchan:

All of our AWS client modules now have a `builds()` and a `provides()` method, the former of which can be reused in fixtures for consistent behaviour and configuration.

This means we simplify our fixture code, and take one more step away from relying on Finatra everywhere.

### Who is this change for?

🌐 🌏


kenoir:
I think this is at odds with https://github.com/wellcometrust/platform/pull/2027 (but shares a similar motivation).

Let's wait until we've decided which we like better before merging this.

kenoir:
After our discussion I think this can go in.

PR #2025, opened 2018-05-10
Update finatra swagger

alicefuzier:
Another step towards upgrading scala. The swagger plugin for finatra needs updating too
### Who is this change for?
⛰ 🚦


alicefuzier:
Closing as the new finatra-swagger plugin only supports scala 2.12 so I'll upgrade it in the upgrade scala branch

PR #2027, opened 2018-05-10
[POC] Creates finatra-common library

kenoir:

In order to extract the Messaging and Storage libraries, they must no longer depend on Finatra libraries.  

This change creates a finatra-common library and rearranges dependencies to suit. 

### Who is this change for?

🔍 🗡 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


alexwlchan:
I’m not a fan of this approach, sorry.

If all the Finatra modules move into a common library, that library has to depend on all the bits of the AWS SDK we use anywhere, which unpicks a bunch of the dependency trimming we’ve just done. It also means the AWS builder logic starts getting split across multiple packages (unless we do further splitting a la #2024).

I deliberately dropped the `finatra.modules` structure when I created the storage/messaging libs, because I prefer organising along service lines.

I guess I’m still not sold on the idea of splitting stuff out into separate libraries? We should talk about it tomorrow.

catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/finatra/modules/ElasticSearchServiceModule/ElasticSearchServiceModule.scala
```diff
@@ -3,11 +3,12 @@ package uk.ac.wellcome.platform.api.finatra.modules.ElasticSearchServiceModule
...
 import com.sksamuel.elastic4s.http.HttpClient
 import com.twitter.inject.TwitterModule
-import uk.ac.wellcome.elasticsearch.finatra.modules.ElasticClientModule
+import uk.ac.wellcome.elasticsearch.{ElasticClientBuilder, ElasticClientConfig}
+import uk.ac.wellcome.finatra.modules.ElasticClientConfigModule
 import uk.ac.wellcome.platform.api.services.ElasticSearchService
 
 object ElasticSearchServiceModule extends TwitterModule {
-  override val modules = Seq(ElasticClientModule)
+  override val modules = Seq(ElasticClientConfigModule)
```
kenoir: Not client pls.

kenoir:
Closing as "we learned a lot but don't put so much in a single PR"

PR #2028, opened 2018-05-11
travis-format: skip format tasks we know aren't significant

### LINKED ISSUE #2004: travistooling: skip format tasks we know aren't significant
e.g. don't run turtle-linting if there aren't any changes to .ttl files

The travis-format task is the first build task we run, so any improvement would be welcome.

alexwlchan:
Resolves #2004.

We save the cost of pulling a couple of Docker containers, so I’d expect this to shave at least a minute or two of every build.

### DIFF
--- a/run_autoformat.py
+++ b/run_autoformat.py
@@ -6,6 +6,7 @@ In particular, it runs the 'make format' task, and if there are any changes,
 it pushes a new commit to your pull request and aborts the current build.
 """
 
+import os
 import sys
 
 from travistooling import branch_name, get_changed_paths, git, make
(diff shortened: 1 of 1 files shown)


alexwlchan:
Just for comparison, travis-format took 2m26 in yesterday's overnight build.

For this patch, it took 43s. I think that’s a worthwhile saving when you consider we run these tests for every single PR, even tiny Markdown changes.

run_autoformat.py
```diff
@@ -6,14 +6,51 @@
 it pushes a new commit to your pull request and aborts the current build.
 """
 
+import os
 import sys
 
 from travistooling import branch_name, get_changed_paths, git, make
 
 
+def _run_task_for_extension(extension, task):
```
kenoir: Really nice.

I don't like that this references travis env vars from a script in the root of the project but not blocking for merge.

PR #2029, opened 2018-05-11
Split out concept identifiers as a separate object

alexwlchan:

The logic for identifiers on concepts is about to get even more complicated, as I need to start looking at ind2=7 and then subfield $2.

Rather than test it through the lenses of Genres/Subject, I'm putting it in its own object with dedicated tests.

This PR just rearranges the existing logic; adding new logic will come separately.

### Who is this change for?

🆔


alexwlchan:
Ah, I changed the expected log message in one test, but not the other. Oops. 🙈

alexwlchan:
Closing until we know what identifiers will look like, longer-term.

alexwlchan:
Related: #2030.

PR #2031, opened 2018-05-11
Add some logging to SQSStream

alexwlchan:
🎹 🌴

### DIFF
--- a/sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/sqs/SQSStream.scala
+++ b/sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/sqs/SQSStream.scala
@@ -43,6 +43,7 @@ class SQSStream[T] @Inject()(actorSystem: ActorSystem,
           () => readAndProcess(streamName, message, process))
       }
       .map { m =>
+        println(s"Deleting message ${m.getMessageId}")
         (m, MessageAction.Delete)
       }
       .runWith(SqsAckSink(sqsConfig.queueUrl)(sqsClient))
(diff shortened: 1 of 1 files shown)


sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/sqs/SQSStream.scala
```diff
@@ -43,6 +43,7 @@ class SQSStream[T] @Inject()(actorSystem: ActorSystem,
           () => readAndProcess(streamName, message, process))
       }
       .map { m =>
+        println(s"Deleting message ${m.getMessageId}")
```
alicefuzier: I don't think we should be using println. It gets ordered in a different way than logged strings and printed strings give you less information.
Is there a reason to not use sl4j logging here?

sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/sqs/SQSStream.scala
```diff
@@ -51,16 +52,21 @@ class SQSStream[T] @Inject()(actorSystem: ActorSystem,
     streamName: String,
     message: Message,
     process: T => Future[Unit])(implicit decoderT: Decoder[T]) = {
+    println(s"Processing message ${message.getMessageId}")
```
alicefuzier: Same

kenoir:
What @alicefuzier said.

sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/sqs/SQSStream.scala
alexwlchan: I’d forgotten we had it! Will fix.

PR #2032, opened 2018-05-11
Fix scala 2.12 compile warnings

alicefuzier:
scala 2.12  compiler warns about unused imports, unused variables and usage of deprecated functions/classes (JavaConversions in most cases for us). This is to fix those warnings so that we can re-enable filing the build on warnings
### Who is this change for?
🚜 🎇


catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/SwaggerController.scala
```diff
@@ -12,7 +12,7 @@ object ApiV1Swagger extends Swagger
 object ApiV2Swagger extends Swagger
 
 @Singleton
-class SwaggerController @Inject()(
+class DocsController @Inject()(
```
alexwlchan: This file needs renaming, doesn’t it?

.travis.yml
```diff
@@ -65,9 +65,9 @@ env:
 
 jobs:
   include:
-    - stage: quicktest
-      env: TASK=travis-format
-    - env: TASK=travistooling-test
+#    - stage: quicktest
```
kenoir: Should these still be commented out?

catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/SwaggerController.scala
alicefuzier: Yep, done, thanks

kenoir:
Amazing work. Upgraded Scala and Finatra in 2 days ... pretty good.

PR #2041, opened 2018-05-14
Fix swagger docs for Subjects and Genres

alicefuzier:
🐝 🔊


alicefuzier:
example output for Genre definition:
```[json]
"Genre" : {
      "type" : "object",
      "properties" : {
        "label" : {
          "type" : "string",
          "description" : "A label given to a thing."
        },
        "concepts" : {
          "type" : "array",
          "description" : "Relates a genre to a list of concepts.",
          "items" : {
            "$ref" : "#/definitions/Concept"
          }
        },
        "type" : {
          "type" : "string",
          "readOnly" : true
        }
      }
```

PR #2042, opened 2018-05-14
Create finatra-messaging package

kenoir:

Complementing the approach of moving to extract finatra providers into apps consuming them over libraries providing them, the number of applications that use the messaging modules justifies a `finatra-messaging` module containing only messaging related providers for finatra.

Further this will allow us to remove the finatra dependency from the messaging library.

This PR duplicates the messaging modules in a `finatra-messaging` package and moves the `id_minter` to use that package as an example of future us.

- [ ] Reviewed by @jtweed
- [ ] Updated the JSON-LD context

- [ ] Deployed new versions

- [ ] Run `terraform apply`.

Transformation checklist:

- [ ] Added any new fields to Work/Item (and the unidentified/identified variants)
- [ ] Update the JSON serialisation tests in [Un]IdentifiedWorkTest
- [ ] Implemented the new transformation
- [ ] Added any new fields to the Elasticsearch mapping
- [ ] Updated the Display models in the API


sbt_common/finatra-messaging/src/main/scala/uk/ac/wellcome/finatra/messaging/SNSClientModule.scala
```diff
@@ -0,0 +1,50 @@
...
+      endpoint = snsEndpoint(),
+      accessKey = accessKey(),
+      secretKey = secretKey()
+    )
+
+  def buildSNSClient(awsConfig: AWSConfig,
+                     endpoint: String,
+                     accessKey: String,
+                     secretKey: String): AmazonSNS = {
+    val standardClient = AmazonSNSClientBuilder.standard
```
alexwlchan: I’d rather see this builder logic pushed down into the messaging library, and the module just call it with appropriate flags.

kenoir: 👍

sbt_common/finatra-messaging/src/main/scala/uk/ac/wellcome/finatra/messaging/SNSClientModule.scala
```diff
@@ -22,29 +21,11 @@ object SNSClientModule extends TwitterModule {
   @Singleton
   @Provides
   def providesSNSClient(awsConfig: AWSConfig): AmazonSNS =
-    buildSNSClient(
-      awsConfig = awsConfig,
+    SNSClientFactory.create(
+      region = awsConfig.region,
```
kenoir: Tempted to drop `AWSConfig` here entirely and just add flags here.

build.sbt
```diff
@@ -60,6 +60,10 @@ lazy val common_storage = doSharedLibrarySetup(project, "sbt_common/storage")
   .dependsOn(common % "compile->compile;test->test")
   .settings(libraryDependencies ++= Dependencies.commonStorageDependencies)
 
+lazy val finatra_messaging = doSharedLibrarySetup(project, "sbt_common/finatra-messaging")
```
alexwlchan: Does this need to be tested in Travis?

kenoir: No, it doesn't have any tests.

kenoir:
@alexwlchan The old modules will be removed as we transition the apps to use `finatra-messaging`

Issue #2039, opened 2018-05-14
ID minter tests are flaky (and spammy)


alexwlchan:
Two of the last three cron jobs on master have failed due to ID minter tests. Log from last night’s run attached, and this is the failure message:

```
[info] - continues if something fails processing a message *** FAILED ***
[info]   The code passed to eventually never returned normally. Attempted 78 times over 1.0073198112333335 minutes. Last failure message: The code passed to eventually never returned normally. Attempted 248 times over 40.073747148 seconds. Last failure message: "[2]" was not equal to "[1]".. (IdMinterFeatureTest.scala:127)
```

[log.txt](https://github.com/wellcometrust/platform/files/1999525/log.txt)

alexwlchan:
Drilling down a bit, I think the failing assertion is actually here:

```scala
sqsClient
  .getQueueAttributes(
    queue.url,
    List("ApproximateNumberOfMessagesNotVisible").asJava
  )
  .getAttributes
  .get("ApproximateNumberOfMessagesNotVisible") shouldBe "1"
```

alexwlchan:
We haven’t seen failures for a few days, I think we can close this for now.

PR #2049, opened 2018-05-15
Less repetition in the Server test fixtures, plus a cleanup (maybe?)

alexwlchan:

While poking at #2039, I noticed that the test logs are spammed with messages of the form:

```
com.amazonaws.services.sqs.model.QueueDoesNotExistException: AWS.SimpleQueueService.NonExistentQueue; see the SQS docs. (Service: AmazonSQS; Status Code: 400; Error Code: AWS.SimpleQueueService.NonExistentQueue; Request ID: 00000000-0000-0000-0000-000000000000)
```

This appears in the logs of several applications.

My guess is that the server created by `withServer(flags)` is taking longer to shut down than the queue does to be deleted – so the `destroy` method in the fixture returns while the server is still running. It does a few more polls, and hits this error.

This patch moves `server.stop()` into an `eventually`, so the server should really be stopped before we clean up the local AWS resources. I still see a bunch of errors about non-existent queues, but I don’t think this makes the problem _worse_.

While I was at it, I decided to tidy up all our server fixtures, as they’re pretty repetitive – now they all benefit from this change!

### Who is this change for?

Folks who want tidy test logs.

### DIFF
--- a/sbt_common/messaging/src/test/scala/uk/ac/wellcome/messaging/test/fixtures/Messaging.scala
+++ b/sbt_common/messaging/src/test/scala/uk/ac/wellcome/messaging/test/fixtures/Messaging.scala
@@ -63,7 +63,8 @@ trait Messaging
   def messageReaderLocalFlags(bucket: Bucket, queue: Queue) =
     Map(
       "aws.message.s3.bucketName" -> bucket.name,
-      "aws.message.sqs.queue.url" -> queue.url
+      "aws.message.sqs.queue.url" -> queue.url,
+      "aws.message.sqs.waitTime" -> "1",
     ) ++ s3ClientLocalFlags ++ sqsLocalClientFlags
 
   def messageWriterLocalFlags(bucket: Bucket, topic: Topic) =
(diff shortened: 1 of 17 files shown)


sbt_common/messaging/src/test/scala/uk/ac/wellcome/messaging/test/fixtures/Messaging.scala
```diff
@@ -63,7 +63,8 @@ trait Messaging
   def messageReaderLocalFlags(bucket: Bucket, queue: Queue) =
     Map(
       "aws.message.s3.bucketName" -> bucket.name,
-      "aws.message.sqs.queue.url" -> queue.url
+      "aws.message.sqs.queue.url" -> queue.url,
+      "aws.message.sqs.waitTime" -> "1",
```
alexwlchan: This is the other big change in this patch, and the one that substantially reduces the log size – we get significantly fewer timeouts from our elasticmq container, and thus less log spam.

alexwlchan:
Before I started, I got several hundred QueueDoesNotExistException’s per test run. Now it’s down to ~40.

Notably, they all occur ~20s after the queue in question has been cleaned up, which makes me think a rogue process on a timeout is stuck somewhere (our default wait time for SQS is 20 seconds).

I don’t want to dig into this any more right now, but I think this makes these tests a lot less flaky. Or if not, the logs are at least more manageable.

alexwlchan:
BOOM.

PR #2056, opened 2018-05-15
Matcher redirects with no history

alicefuzier:
Redirect! But not knowing the graph still
### Who is this change for?
🔩 🚲


alicefuzier:
I'm not sure whether this PR still makes sense on its own. The logic here doesn't make much sense anymore and all of this is rejigged in #2067. I don't think merging this first will make #2067 diff smaller

kenoir:
Closing as superseded.

PR #2057, opened 2018-05-15
Sending medium agnostic messaging

kenoir:

We need to be able to send messages that are too large for SNS/SQS and also account for the case where we know the messages are small enough to send without S3.

Creates a `MessageRetriever` and `MessageSender` trait, with `MessageStream` and `MessageWriter` accepting those as constructor params respectively.

The implementer can write their own retrieval and sending logic.

This PR implements:
-  `TypeMessageSender`: Send strongly typed objects using SNS only.
- `TypeMessageRetriever`: Send strongly typed objects using SNS and S3, for when your objects are too large to send.
- `S3TypeMessageSender`: Gets strongly typed objects using SQS only.
- `S3TypeMessageRetriever` : Gets strongly typed objects using SQS and S3, for when your objects are too large to retrieve.

### Who is this change for?

📹 🐕

### DIFF
--- a/sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/message/MessageRetriever.scala
+++ b/sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/message/MessageRetriever.scala
@@ -10,7 +10,8 @@ import scala.concurrent.{ExecutionContext, Future}
 
 trait MessageRetriever[T] {
   def retrieve(notification: NotificationMessage)(
-    implicit decoderT: Decoder[T], ec: ExecutionContext
+    implicit decoderT: Decoder[T],
+    ec: ExecutionContext
   ): Future[T]
 }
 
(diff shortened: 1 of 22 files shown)


sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/message/MessageRetriever.scala
```diff
@@ -0,0 +1,30 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+trait MessageRetriever[T] {
+  def retrieve(notification: NotificationMessage)(implicit decoderT: Decoder[T],
+                                                  decoderN: Decoder[NotificationMessage], ec: ExecutionContext): Future[T]
+}
+
+class S3TypeMessageRetriever[T](s3TypeStore: S3TypeStore[T])(implicit decoderP: Decoder[MessagePointer]) extends MessageRetriever[T] {
+  def retrieve(notification: NotificationMessage)(implicit decoderT: Decoder[T],
+                                                  decoderN: Decoder[NotificationMessage], ec: ExecutionContext): Future[T] = for {
```
alicefuzier: why does it take a Decoder of NotificationMessage? I don't see it ever decoding a NotificationMessage

kenoir:
Closing as too rabbit-holey.

PR #2060, opened 2018-05-16
Create finatra-storage package

kenoir:

Complementing the approach of moving to extract finatra provider modules into apps consuming them over libraries providing them, the number of applications that use the messaging modules justifies a `finatra-storage` module containing only messaging related providers for finatra.

Further this will allow us to remove the finatra dependency from the messaging library.

This PR duplicates the storage modules in a `finatra-storage` package and moves the `id_minter` to use that package as an example of future us.

### Who is this change for?

Folk who want to remove dependence on finatra, simplify and reduce project dependencies (in the long run) and extract messaging as a separately built library to reduce project build times.

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


sbt_common/finatra-storage/src/main/scala/uk/ac/wellcome/finatra/storage/DynamoClientModule.scala
```diff
@@ -34,18 +35,12 @@ object DynamoClientModule extends TwitterModule {
...
-    else
-      standardClient
-        .withCredentials(
-          new AWSStaticCredentialsProvider(
-            new BasicAWSCredentials(accessKey, secretKey)))
-        .withEndpointConfiguration(
-          new EndpointConfiguration(endpoint, awsConfig.region))
-        .build()
+
+    DynamoClientFactory.create(
```
alexwlchan: There’s no reason to go via a buildDynamoClient method in the module now; just call the factory directly.

PR #2067, opened 2018-05-16
Matcher links directly referenced works

alicefuzier:
* Make the matcher output links for directly connected works without keeping track of them 
* Use new format for the matcher output
* `LinkedWorkGraphUpdater` implements the graph logic to deal with edges added/deleted
### Who is this change for?
🐬 🐓


sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/IdentifierSchemes.scala
```diff
@@ -38,6 +38,12 @@ object IdentifierSchemes {
     override def toString: String = "sierra-system-number"
   }
 
+  // Represents a combined work resulting from merging related works
```
alicefuzier: This can go away now

catalogue_pipeline/matcher/src/test/scala/uk/ac/wellcome/platform/matcher/MatcherMessageReceiverTest.scala
```diff
@@ -57,46 +59,106 @@ class MatcherMessageReceiverTest
...
-          actorSystem)
+          actorSystem,
+          new LinkedWorkMatcher)
         testWith(matcherMessageReceiver)
       }
     }
   }
 
-  it("receives a message with UnidentifiedWork") {
+  it("sends no redirects for a work without identifiers") {
```
alicefuzier: This isn't dealing with redirects anymore. We should avoid that word

PR #2073, opened 2018-05-17
Less is more: improve the coverage of Lambda testing in Travis while saving build jobs

alexwlchan:

Currently we spend 7 build VMs per Travis run on Lambdas – even though we primarily write in Scala. That means we spend a significant amount of time starting build VMs that immediately exit.

This patch puts all our Lambda testing on a single VM, so if a patch only contains Scala changes we save having to boot **six VMs**. Additional niceties:

* Within the travis-lambda-test task, each Lambda is still considered separately. If we’ve changed one Lambda, we should only test that Lambda, not the whole set.
* We get rid of the shared_infra-test task, which was all-or-nothing.
* The monitoring Lambdas are being tested in Travis again.
* Now Lambda dependencies are pinned (see #2063), publishes should be truly deterministic, so we won't have unnecessary deployment churn.

I'm still fine-tuning the details, but I think this is nearly ready.

### Who is this change for?

🏷 Ⓜ️

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,6 +80,21 @@ jobs:
     - env: TASK=common_storage-test
     - env: TASK=common_monitoring-test
 
+    # Lambda tests.
+    #
+    # We test all the Lambdas in a single job to reduce the number of small
+    # build VMs we start.  Because most of our applications are Scala, the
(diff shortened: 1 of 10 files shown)


alexwlchan:
Oh, and as a bonus, we only have to pull the Lambda test container once!

PR #2074, opened 2018-05-17
[RFC] Merged work identifier is one of the original source identifiers

alicefuzier:
This is putting in writing some things that we have discussed.
I think that not creating a new sourceIdentifier for the merged work actually makes our life easier:
* The matcher is truly a matcher, it has no mixed redirection logic
* The matcher needs to store each work with its links and the group it currently belongs to, but it doesn't need to store groups that used to exist and have been broken down.
* This means the matcher table can be entirely recreated with a full reindex. It can be safely deleted.
* The matcher ouput is just made by lists of work identifiers that represent how the works group together

### DIFF
--- a/docs/rfcs/merger_matcher/README.md
+++ b/docs/rfcs/merger_matcher/README.md
@@ -11,23 +11,22 @@ merge them into combined works to aid understanding and searching.
 
 ## Problem Statement
 
-Individual works can be merged to form larger combined works.
+Individual works can be merged to form larger combined works. Combined works can be later split.
+Each individual work has a unique source identifier. The source identifier for the merged works will be one of the source
+identifier of the original works that compose it.
+
 The merging of works is broken into two phases
 
- * a matching/redirect phase that identifies works to be merged and tracks previously identified combined works.
+ * a matching phase that identifies groups of work to be merged.
  * a merging phase that merges works to form new combined works.
    
-The matcher/redirecter receives source identifiers for processed works along with source identifiers for works that should be merged, 
-and determines what changes to published works should be made.  Changes can include combining existing works and redirecting previously 
-published works that have been superseded by merging operations.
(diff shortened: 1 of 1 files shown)


docs/rfcs/merger_matcher/README.md
```diff
@@ -11,47 +11,62 @@ merge them into combined works to aid understanding and searching.
...
-Individual works can be merged to form larger combined works.
+Individual works can be merged to form larger combined works. Combined works can be later split.
+Each individual work has a unique source identifier. The source identifier for the merged works will be one of the source
+identifier of the original works that compose it.
+
 The merging of works is broken into two phases
 
- * a matching/redirect phase that identifies works to be merged and tracks previously identified combined works.
+ * a matching phase that identifies groups of work to be merged.
  * a merging phase that merges works to form new combined works.
```
mklander: We use different phrases for similar things (groups, matcher, links) I suggest sticking to one for clarity.  Linked works feels closest at the moment to me.  So perhaps s/groups of work/linked works/

docs/rfcs/merger_matcher/README.md
```diff
@@ -78,32 +93,25 @@ For example creating a combined work 'A-B' when processing work 'A' could give a
...
+will output:
+```[json]
+{
+  "work-groups": [
+    {
+      "identifiers": [ "sierra-system-number/s12345", "miro-image-number/s67890"]
+    }
+  ]
+}
 ```
```
mklander: Perhaps output:




```[json]
{
  "linked-works": [
    {
      "identifiers": [ "sierra-system-number/s12345", "miro-image-number/s67890"]
    }
  ]
}

alicefuzier: You mean if the merger needs to know which work caused the update? I don't think so, but let's see what the others think

alexwlchan:
(I cancelled the build to get back some Travis time, patch is fine)

PR #2084, opened 2018-05-18
Move GlobalExecutionContext into apps

kenoir:

The GlobalExecutionContext should be global only within an app.

### Who is this change for?

💇‍♂️


kenoir:
Yeah, probably ... maybe not in this PR though. Arguably they are "application global" but I don't mind that much.

PR #2089, opened 2018-05-18
Pass the VHS record ID to KeyPrefixGenerator.generate

alexwlchan:

Include the ID when generating the key prefix. This is required for work on the Goobi reader – because we're passing around streams of data, we can’t inspect the object to be stored to get an ID.

### Who is this change for?

🎹 🤴


alicefuzier:
Uhm, it feels a bit icky that the `MessageWriter` calls `keyPrefixGenerator.generate(id = "", obj = message)` with an empty string. I can't see where else this is going to be used but presumably, it's going to be something like `keyPrefixGenerator.generate(id = id, obj = "")`? I wonder if there's a cleaner way of doing this..

alexwlchan:
Closing in favour of passing around a more generic key generation strategy.

PR #2090, opened 2018-05-19
Remove Finatra from the common lib

alexwlchan:

Remove all the finatra-http pieces – not quite all the injector stuff yet – from the common lib. This isn’t working yet, and has exposed a couple of bits of finatra still lying around elsewhere – but we’re getting there!

### Who is this change for?

🏑 🇫🇯


alexwlchan:
Blocked on #2091 and #2092.

common/src/test/scala/uk/ac/wellcome/test/fixtures/ServerFixtures.scala
```diff
@@ -1,28 +0,0 @@
-package uk.ac.wellcome.test.fixtures
-
-import com.twitter.finatra.http.EmbeddedHttpServer
-import com.twitter.inject.server.Ports
-import org.scalatest.Suite
-
-trait ServerFixtures { this: Suite =>
```
alexwlchan: In the end, I found it easier to split this back out than find somewhere for it to live.

PR #2098, opened 2018-05-21
Extract storage strategy

kenoir:

Allow storage strategy and key generation to be defined outside the storage layer.

### Who is this change for?

👺


sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/type_classes/StorageStrategy.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  private def hash(s: String) =
+    MurmurHash3
+      .stringHash(s, MurmurHash3.stringSeed)
+      .toHexString
+
+  def getKey[T](t: T)(implicit strategy: StorageStrategy[T]) =
+    strategy.get(t)
+
+  implicit val inputStreamKeyGetter: StorageStrategy[InputStream] =
```
kenoir: This is super-dumb I know, but now they are in the same place we can fix it later.

sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/s3/S3Storage.scala
```diff
@@ -3,30 +3,48 @@ package uk.ac.wellcome.storage.s3
...
+    objectMetadata.setUserMetadata(userMetadata.asJava)
+
+    objectMetadata
+  }
+
+  def put[T](s3Client: AmazonS3)(
+    bucketName: String
+  )(t: T,
+    keyPrefix: String = "",
+    userMetadata: Map[String, String] = Map.empty[String, String])(
```
alexwlchan: nit: I think the following has the same effect, and I think it’s a bit cleaner:

```scala
userMetadata: Map[String, String] = Map())(
```

sbt_common/storage/src/test/scala/uk/ac/wellcome/storage/s3/S3TypeStoreTest.scala
```diff
@@ -38,7 +38,7 @@ class S3TypeStoreTest
         whenReady(writtenToS3) { actualKey =>
           val expectedJson = JsonUtil.toJson(testObject).get
 
-          val expectedKey = s"$prefix/$expectedHash.json"
```
alexwlchan: Why have these expected keys changed? This change in external behaviour feels wrong to me.

sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/type_classes/StorageStrategy.scala
```diff
@@ -0,0 +1,56 @@
...
+
+import java.io.{ByteArrayInputStream, InputStream}
+
+import io.circe.Json
+
+import scala.io.Source.fromInputStream
+import scala.util.hashing.MurmurHash3
+
+trait StorageStrategy[T] {
+  def get(t: T): (InputStream, String)
```
alexwlchan: Some comments explaining what this returns would be helpful, because it's super non-obvious.

We've discussed this enough, it might be worth that comment including _why_ we have this signature.

sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/type_classes/StorageStrategy.scala
```diff
@@ -0,0 +1,56 @@
...
+
+object StorageStrategyGenerator {
+
+  private def hash(s: String) =
+    MurmurHash3
+      .stringHash(s, MurmurHash3.stringSeed)
+      .toHexString
+
+  def getKey[T](t: T)(implicit strategy: StorageStrategy[T]) =
+    strategy.get(t)
```
alexwlchan: Is this used?

sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/type_classes/StorageStrategy.scala
```diff
@@ -0,0 +1,56 @@
...
+
+        (input, key)
+      }
+    }
+
+  implicit val jsonKeyGetter: StorageStrategy[Json] =
+    new StorageStrategy[Json] {
+      def get(t: Json): (InputStream, String) = {
+        val key = hash(t.noSpaces)
+        val input = new ByteArrayInputStream(t.noSpaces.getBytes)
```
alexwlchan: Rather than repeating the stringKeyGetter, could you serialise the JSON as a string and then use the strategy above?

alexwlchan:
I think mostly fine, but needs more comments to explain what's going on.

sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/type_classes/StorageStrategy.scala
kenoir: These strategies should probably be independent. `.noSpaces` seems the most convenient way of performing the serialisation (it's provided by circe).

sbt_common/storage/src/test/scala/uk/ac/wellcome/storage/s3/S3StringStoreTest.scala
```diff
@@ -29,7 +29,7 @@ class S3StringStoreTest
           stringStore.put(bucket.name)(content = content, keyPrefix = prefix)
 
         whenReady(writtenToS3) { actualKey =>
-          val expectedKey = s"$prefix/$expectedHash.json"
```
alexwlchan: 👍

PR #2105, opened 2018-05-22
Set the default for the aws region

alicefuzier:


alicefuzier:
set the default for the aws region by alicefuzier pull request 2105 wellcometrust platform
2 hours 🙄

PR #2106, opened 2018-05-22
Use streams in id_minter

kenoir:

Use streams in the id_minter, to get better performance and standardise on streams.

### Who is this change for?

🐐


catalogue_pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/services/IdMinterWorkerService.scala
```diff
@@ -1,39 +1,37 @@
...
+  messageStream: MessageStream[Json],
+  system: ActorSystem
+) {
 
-  override lazy val poll = 100 milliseconds
-  val snsSubject = "identified-item"
+  messageStream.foreach(this.getClass.getSimpleName, processMessage)
 
-  override implicit val decoder: Decoder[Json] = Decoder.decodeJson
+  val snsSubject = "identified-item"
```
alicefuzier: Is thhis used?

PR #2111, opened 2018-05-22
Messaging/Storage/Monitoring Finatra independent

kenoir:

Allow the extraction of Messaging/Storage/Monitoring libraries as appropriate by removing the dependency on Finatra and replacing with common interface alternatives (guice/sl4j).

### Who is this change for?

Folk who want to reduce build times, version projects with many dependencies, and contribute to the open source community.

- [ ] Deployed new versions
- [ ] Run `terraform apply`.

### DIFF
--- a/sbt_common/storage/src/test/scala/uk/ac/wellcome/storage/test/fixtures/S3.scala
+++ b/sbt_common/storage/src/test/scala/uk/ac/wellcome/storage/test/fixtures/S3.scala
@@ -75,9 +75,11 @@ trait S3 extends Logging with Eventually {
     )
 
   def getContentFromS3(bucket: Bucket, key: String): String =
-    scala.io.Source.fromInputStream(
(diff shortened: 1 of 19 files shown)


sbt_common/storage/src/test/scala/uk/ac/wellcome/storage/test/fixtures/S3.scala
```diff
@@ -75,9 +74,12 @@ trait S3 extends Logging with Eventually {
       }
     )
 
-  def getContentFromS3(bucket: Bucket, key: String): String = {
-    IOUtils.toString(s3Client.getObject(bucket.name, key).getObjectContent)
-  }
+  def getContentFromS3(bucket: Bucket, key: String): String =
+    scala.io.Source
```
alexwlchan: Gentle ding for mixing unrelated changes into a PR.

PR #2116, opened 2018-05-23
Increase redrive attempts for items_to_dynamo

alicefuzier:
Errors in items_to_dynamo seem to be due mostly to throughput exceeded or conditional check failures. Increasing the number of times messages are resent before ending in the dlq should help

### DIFF
--- a/sierra_adapter/terraform/items_to_dynamo/queues.tf
+++ b/sierra_adapter/terraform/items_to_dynamo/queues.tf
@@ -8,5 +8,4 @@ module "demultiplexer_queue" {
   alarm_topic_arn = "${var.dlq_alarm_arn}"
 
   max_receive_count = 10
-
 }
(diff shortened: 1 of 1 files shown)


sierra_adapter/terraform/items_to_dynamo/queues.tf
```diff
@@ -6,4 +6,7 @@ module "demultiplexer_queue" {
   topic_names = ["${var.demultiplexer_topic_name}"]
 
   alarm_topic_arn = "${var.dlq_alarm_arn}"
+
+  max_receive_count = 10
```
alexwlchan: There should be a comment accompanying this line explaining why we’ve increased the max_receive_count (use what you wrote in the PR description).

kenoir: And it'd be useful to mark from what it was increased.

alicefuzier: 👍

PR #2117, opened 2018-05-23
Move common to sbt_common

kenoir:

All the Scala libraries sit in the same place so all make tasks take for them take the same form and are easily discoverable.

### Who is this change for?

🗃


sbt_common/Makefile
```diff
@@ -3,7 +3,7 @@ include $(ROOT)/functions.Makefile
 
 STACK_ROOT 	= sbt_common
 
-SBT_DOCKER_LIBRARIES = common_elasticsearch common_messaging common_monitoring common_storage
+SBT_DOCKER_LIBRARIES = common common_elasticsearch common_messaging common_monitoring common_storage
```
alexwlchan: Isn’t this a NO_DOCKER library? We don’t run Docker for any common tests (we barely run any common tests, period!).

PR #2119, opened 2018-05-23
Update Finatra to 18.4.0

alicefuzier:
Couple of comments:
* Couldn't migrate to finatra 18.5.0 because finatra-swagger brings in jackson-databind 9.2 which is incompatible with the jackson-scala-module that it brings in. We could work around that with exclusions or upgrading the jackson-scala-module in our dependencies, but that means polluting our dependencies file
* finatra-swagger now always adds the full path relative to the host, not the prefix thanks to [this "improvement"](https://github.com/jakehschwartz/finatra-swagger/pull/27), so had to remove the basePath to keep a valid swagger.json. 
* Other changes are mostly minor around deprecated methods and so on

### DIFF
--- a/catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/ApiSwaggerTest.scala
+++ b/catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/ApiSwaggerTest.scala
@@ -91,24 +91,7 @@ class ApiSwaggerTest extends FunSpec with Matchers with fixtures.Server {
       .toString shouldBe "\"#/definitions/Work\""
   }
 
-  // Because the Swagger should be the same on every request, we cache
-  // requests for the Swagger to save the time cost of starting a new
-  // server for every test/request.
-  //
-  var cachedResponses = Map[String, JsonNode]()
-
   def readTree(path: String): JsonNode = {
-    cachedResponses.get(path) match {
-      case Some(node) => node
-      case None => {
-        val newTree = readTreeFromServer(path)
-        cachedResponses += (path -> newTree)
-        newTree
-      }
-    }
-  }
-
-  def readTreeFromServer(path: String): JsonNode = {
     val flags = Map(
       "api.host" -> "test.host",
       "api.scheme" -> "http",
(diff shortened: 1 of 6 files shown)


catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/ApiSwaggerTest.scala
```diff
@@ -91,24 +91,7 @@ class ApiSwaggerTest extends FunSpec with Matchers with fixtures.Server {
       .toString shouldBe "\"#/definitions/Work\""
   }
 
-  // Because the Swagger should be the same on every request, we cache
-  // requests for the Swagger to save the time cost of starting a new
-  // server for every test/request.
-  //
-  var cachedResponses = Map[String, JsonNode]()
```
alicefuzier: this was causing issues in this branch and as a general rule in tests we should avoid mutable variables unless relly necessary

catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/ApiSwaggerTest.scala
```diff
@@ -11,53 +11,52 @@ class ApiSwaggerTest extends FunSpec with Matchers with fixtures.Server {
...
     ApiVersions.values.toList.foreach { version: ApiVersions.Value =>
-      val tree = readTree(s"/test/${version.toString}/swagger.json")
+      val tree = readTree(s"/catalogue/${version.toString}/swagger.json")
       tree.at("/paths").isObject shouldBe true
       tree
         .at("/paths")
         .fieldNames
         .asScala
-        .toList should contain only ("/works", "/works/{id}")
+        .toList should contain only (s"/catalogue/${version.toString}/works", s"/catalogue/${version.toString}/works/{id}")
```
alicefuzier: This is due to the fact that swagger now outputs paths relative to the host. This also forced to test using the real "catalogue" prefix because `ApiV1Swagger` and `ApiV2Swagger` are singleton objects that are instantiated when the JVM starts. Tests run in a single jvm so they are created once. The ApiTest specify the "/catalogue/{version}/works" endpoints, which are visible from then onwards in the swagger objects, so they were leaking into this test. So I had to change this test suite to use the real api prefix 😐

PR #2121, opened 2018-05-23
an optional index parameter added to DynamoConfig

mklander:
Add an optional index parameter to DynamoConfig to support a global secondary index, needed by Matcher graph store

### DIFF
--- a/sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/dynamo/DynamoConfig.scala
+++ b/sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/dynamo/DynamoConfig.scala
@@ -1,3 +1,3 @@
 package uk.ac.wellcome.storage.dynamo
 
-case class DynamoConfig(table: String)
+case class DynamoConfig(table: String, index: Option[String])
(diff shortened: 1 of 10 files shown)


sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/dynamo/DynamoConfig.scala
```diff
@@ -1,3 +1,3 @@
 package uk.ac.wellcome.storage.dynamo
 
-case class DynamoConfig(table: String)
+case class DynamoConfig(table: String, index: Option[String])
```
alexwlchan: Would it be neater to default this new argument to None, and omit it from the constructor unless used? Do any of the changes below actually use the index parameter?

alicefuzier: I think I prefer it to not have a default value because it would only be there for the purpose of tests. It's never going to use this default value in actual production code. Also, not having a default forced us to modify all of the fixtures to pass the table.index around, which makes sense I think

PR #2124, opened 2018-05-23
Refactor the VHS to use Migration when reading from DynamoDB

alexwlchan:

Refactors the VHS to use #2120, but with no change in external functionality or signature. Adding metadata will be a separate patch.

### Who is this change for?

🇫🇲 🇨🇻


alexwlchan:
This is the error I’m seeing when I run the storage tests locally:

```
[error] /repo/sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/vhs/VersionedHybridStore.scala:81: diverging implicit expansion for type com.gu.scanamo.DynamoFormat[T]
[error] starting with method someFormat in object DynamoFormat
[error]     getObject(id).map { maybeObject =>
[error]              ^
[error] one error found
[error] (common_storage/compile:compileIncremental) Compilation failed
[error] Total time: 7 s, completed May 24, 2018 8:40:17 AM
```

alexwlchan:
Adding the dynamo import got me unstuck there, now I have a different implicit failure:

```
[error] /repo/sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/vhs/VersionedHybridStore.scala:125: could not find implicit value for parameter migration: uk.ac.wellcome.storage.type_classes.Migration[DynamoRow,uk.ac.wellcome.storage.vhs.HybridRecord]
[error]         val hybridRecord = dynamoRow.migrateTo[HybridRecord]
[error]                                               ^
[error] /repo/sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/vhs/VersionedHybridStore.scala:125: local val hybridRecord in value $anonfun is never used
[error]         val hybridRecord = dynamoRow.migrateTo[HybridRecord]
[error]             ^
[error] two errors found
[error] (common_storage/compile:compileIncremental) Compilation failed
[error] Total time: 6 s, completed May 24, 2018 8:46:42 AM
```

alicefuzier:
I think the reason it fails compiling now is this:
```
 def getRecord(id: String): Future[Option[T]] =
     getObject(id).map { maybeObject =>	     getObject(id).map { maybeObject =>
       maybeObject.map(_.s3Object)	       maybeObject.map(_.s3Object)
     }	     
}
```
`getObject` is generic now and takes the type of the `DynamoRow`. In theory if you change that snippet with 
```
 def getRecord[DynamoRow](id: String)(implicit dynamoFormat: DynamoFormat[DynamoRow]): Future[Option[T]] =
     getObject[DynamoRow](id).map { maybeObject =>	     getObject(id).map { maybeObject =>
       maybeObject.map(_.s3Object)	       maybeObject.map(_.s3Object)
     }	     
}
```
it should compile. 
Problem is, it becomes unusable because to call `getRecord` you now need to know what `DynamoRow` is and we don't know cause it's automagically generated.

alicefuzier:
I think we have two options:
* Don't use `getObject` in `getRecord` and retrieve just the `HybridRecord` in that case, cause it's the only thing needed to return `T`
* make `getRecord` return `T` and the `Metadata`: something like ```getRecord[T, Medatata](id: String): Future[Option[(T,Metadata)]]```

alexwlchan:
I prefer the first option.

The point of getObject is to return you a VHSObject, if it returns the Dynamo row we’ve pushed he problem around. I’ll have a play.

alexwlchan:
Okay, so I think passing HybridRecord as a type parameter to getObject also works?

The migration implicit is still broken, blergh.

alexwlchan:
Blergh, still having issues. The Migration type class isn’t being instantiated for some reason.

alicefuzier:
Uhm, I'll have a look

alicefuzier:
`uk.ac.wellcome.storage.type_classes.Migration.Aux[DynamoRow,uk.ac.wellcome.storage.vhs.HybridRecord]`
Where does the .Aux come from?? the Migration object doesn't have a Aux type

alexwlchan:
that must be a merge conflict, I tried having an Aux (a la HybridRecordEnricher) but it made no difference

has one snuck into the diff?

alexwlchan:
Ah no, I see it’s in the error. I don’t think it made any significant difference to my errors.

sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/type_classes/Migration.scala
```diff
@@ -34,39 +34,55 @@ trait Migration[Source, Target] {
 }
 
 object Migration {
-  implicit class MigrationOps[Source](src: Source) {
-    def migrateTo[Target](
-      implicit migration: Migration[Source, Target]): Target =
+  implicit class MigrationOps[Source, Target](src: Source) {
+    def migrateTo[T](implicit migration: Migration[Source, T]): T =
```
alexwlchan: What’s the difference between T and Target?

sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/type_classes/Migration.scala
```diff
@@ -34,39 +34,55 @@ trait Migration[Source, Target] {
...
-  ): Migration[Source, Target] = new Migration[Source, Target] {
-    def apply(src: Source): Target = {
-      val sourceHList: SRepr = sourceGen.to(src)
-
+  ): Migration[SRepr, Target] = new Migration[SRepr, Target] {
+    def apply(src: SRepr): Target = {
       // This gets the key-value pairs whose keys are in both `Source` and
       // `Target`, but they may not be in the correct order.
-      val commonFields: Unaligned = intersection.apply(sourceHList)
+      val commonFields: Unaligned = intersection.apply(src)
```
alexwlchan: Let’s keep `sourceHList: SRepr` for easy distinction from `src: Source` below.

PR #2126, opened 2018-05-23
LocalDynamoDb split for reuse

mklander:
Split up the current LocalDynamoDb trait used in testing to be reusable while supporting different databases.  Principally by moving out table creation.

### DIFF
--- a/reindexer/reindex_worker/src/test/scala/uk/ac/wellcome/platform/reindex_worker/ReindexerFeatureTest.scala
+++ b/reindexer/reindex_worker/src/test/scala/uk/ac/wellcome/platform/reindex_worker/ReindexerFeatureTest.scala
@@ -1,8 +1,8 @@
 package uk.ac.wellcome.platform.reindex_worker
 
-import com.gu.scanamo.{DynamoFormat, Scanamo}
-import org.scalatest.{FunSpec, Matchers}
+import com.gu.scanamo.Scanamo
 import org.scalatest.concurrent.{Eventually, ScalaFutures}
+import org.scalatest.{FunSpec, Matchers}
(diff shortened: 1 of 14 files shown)


reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex_worker/services/ReindexService.scala
```diff
@@ -1,37 +1,32 @@
...
 
-import javax.inject.Inject
 import com.amazonaws.services.dynamodbv2.AmazonDynamoDB
 import com.gu.scanamo.error.DynamoReadError
 import com.gu.scanamo.query._
 import com.gu.scanamo.syntax._
 import com.gu.scanamo.{Scanamo, _}
 import com.twitter.inject.Logging
-import com.twitter.inject.annotations.Flag
+import javax.inject.Inject
```
kenoir: I've used Guice `@Inject` elsewhere as i've made the assumption we're using it for DI everywhere. Should We be using `javax.inject.Inject` instead for better compatibility with other DI libraries?

mklander: I don't think javax.inject is introduced in this change, but the import moved.  Guice should be [JSR330 compatible](https://github.com/google/guice/wiki/JSR330), so I'd prefer it except if we are using Guice inject everywhere else, or using it via Scala introduces doubts over compatibility.

Looks like `com.google.inject.Inject` wins 37 to 2 for `javax.inject.Inject` (here and SnapshotService).  Perhaps another PR to make it Guice Inject everywhere?

PR #2128, opened 2018-05-24
Add a way to get reference data into our applications

alexwlchan:

Add a script, a Make task, and a Travis job that automatically:

* Checks files in the reference data folder are formatted correctly (to reduce the risk of runtime errors)
* Copies them into the `resources` folder of the target application

There’s probably a better way to load external resources with sbt, but after four attempts I am neither smart nor patient enough to work it out.

### Who is this change for?

⏺ ♻️


alicefuzier:
Uhm I wonder if simply linking them in the resources folder would work?

alexwlchan:
I had it in my head that symlinks wouldn't work, but let me double-check that!

kenoir:
I think I like this approach over symlinks, so long as the copy operation isn't automated - stepping the resource forward in the consuming library should be a separate step. As such i'd prefer the copying step separate from the lint step and the copy step to be omitted from automated build.

alexwlchan:
Yep, so trying it with symlinks, the compilation complains with "No such file or directory".

alicefuzier:
Actually I've been convinced. They did work in a maven project I've worked with in a previous job. But I see that doing it this way, means that a change in the ontologies effectively automatically triggers a new build for the applications. Might be safer to copy them in the resources folder as a separate commit exiplicilty when we want to.

alexwlchan:
If the copying is a separate step, then I think I can expunge all of this except actually copying the file across? Because then you check it’s correct when you copy it across (and in the application too, for that matter).

run_autoformat.py
```diff
@@ -43,6 +43,7 @@ def _run_task_for_extension(extension, task):
         ('.scala', 'format-scala'),
         ('.py', 'format-python'),
         ('.json', 'format-json'),
+        ('.csv', 'check-reference_data'),
```
alicefuzier: Does this mean that files get copied across as part of the linting phase? It makes sense to check them, but I'm not sure about copying them...

alexwlchan: The current script automatically copies them across, but I think we’re all agreed that we don’t actually want that.

alexwlchan:
I’m going to close this, and just copy the files across manually.

alicefuzier:
Ah ok. It might be good to have a linting step for csv failes though. So that we find out if they're broken before we copy them across. It's going to be annoying if, copying new versions of these csv to the applications also means having to fix them

alexwlchan:
Fine, I’ll put that as a separate TODO.

Issue #2030, opened 2018-05-11
Tweak identifier model for genre/subject identifiers in Sierra


alexwlchan:
The MARC tags we read for identifiers on genre/subject either come from a couple of prefilled headings, or a freeform text field in a different subfield:

> **Second Indicator**
> 0 - Library of Congress Subject Headings
1 - LC subject headings for children's literature
2 - Medical Subject Headings
3 - National Agricultural Library subject authority file
4 - Source not specified
5 - Canadian Subject Headings
6 - Répertoire de vedettes-matière
7 - Source specified in subfield $2
>
> https://www.loc.gov/marc/bibliographic/bd651.html

It's that 7 which is causing us issues.

The current code rejects anything that's not "0" or "2", and so we have ~250k records on the DLQ. Some of these have second indicator "4", which is a small patch.

The remaining genres/subjects come from all over the shop. This is a brief tally from the first 4000 records:

```
Counter({'aiatsiss': 2,
         'bidex': 5,
         'bisacsh': 57,
         'blmlsh': 6,
         'cct': 22,
         'dcs': 2,
         'eclas': 88,
         'eflch': 1,
         'embne': 4,
         'fast': 8107,
         'fssh': 3,
         'gmgpc': 392,
         'gmgpc ': 4,
         'gnd': 2,
         'gsafd': 1,
         'idszbz': 10,
         'idszbzes': 3,
         'inriac': 4,
         'jhpk': 3,
         'jlabsh/4': 1,
         'larpcal': 139,
         'lcgft': 27,
         'local': 198,
         'ram': 174,
         'rasuqam': 22,
         'rbgenr': 444,
         'renib': 16,
         'rero': 8,
         'retrosciences': 1,
         'sao': 1,
         'sigle': 7,
         'unbist': 6})
```

I think we have a couple of options:

1. Ignore all of them
2. Add enumerations for the most common identifier schemes, and ignore the rest
3. This list seems to be (partially) populated from the MARC site https://www.loc.gov/standards/sourcelist/genre-form.html – we could get rid of the enumeration, and load this list as a resource (a la Miro contributors). If we go down this approach, we might need to think about getting rid of the snake-case-strings and replace them with proper sentences (which we might want to do anyway!)

alexwlchan:
If we go down the route of `lc-subjects`, `lc-childrens-subjects`, and so on, I’d like to push for having a human-readable label. Some of the names are quite opaque, and not especially searchable.

[...]

PR #2129, opened 2018-05-24
Create a new IdentifierType type, populated from a reference data CSV

alexwlchan:

This is based on the model proposed in https://github.com/wellcometrust/platform/issues/2030. This PR creates the model and populates it from the CSV, but doesn’t actually use it anywhere – that will be another patch.

### Who is this change for?

🉐 👇


alexwlchan:
My big hangup with this is that we lose our compile-time safety.

I wonder if it’s possible to dynamically create an Enum at compile-time from Shapeless?

alexwlchan:
If I try to run tests in IntelliJ and I mess up one of those constants, I get an immediate error:

```
An exception or error caused a run to abort. 
java.lang.ExceptionInInitializerError
	at uk.ac.wellcome.models.work.internal.IdentifierTypeTest.$anonfun$new$1(IdentifierTypeTest.scala:7)
	at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
	at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
	at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
	at org.scalatest.Transformer.apply(Transformer.scala:22)
	at org.scalatest.Transformer.apply(Transformer.scala:20)
	at org.scalatest.FunSpecLike$$anon$1.apply(FunSpecLike.scala:454)
	at org.scalatest.TestSuite.withFixture(TestSuite.scala:196)
	at org.scalatest.TestSuite.withFixture$(TestSuite.scala:195)
	at org.scalatest.FunSpec.withFixture(FunSpec.scala:1630)
	at org.scalatest.FunSpecLike.invokeWithFixture$1(FunSpecLike.scala:452)
	at org.scalatest.FunSpecLike.$anonfun$runTest$1(FunSpecLike.scala:464)
	at org.scalatest.SuperEngine.runTestImpl(Engine.scala:289)
	at org.scalatest.FunSpecLike.runTest(FunSpecLike.scala:464)
	at org.scalatest.FunSpecLike.runTest$(FunSpecLike.scala:446)
	at org.scalatest.FunSpec.runTest(FunSpec.scala:1630)
	at org.scalatest.FunSpecLike.$anonfun$runTests$1(FunSpecLike.scala:497)
	at org.scalatest.SuperEngine.$anonfun$runTestsInBranch$1(Engine.scala:396)
	at scala.collection.immutable.List.foreach(List.scala:389)
	at org.scalatest.SuperEngine.traverseSubNodes$1(Engine.scala:384)
	at org.scalatest.SuperEngine.runTestsInBranch(Engine.scala:379)
	at org.scalatest.SuperEngine.runTestsImpl(Engine.scala:461)
	at org.scalatest.FunSpecLike.runTests(FunSpecLike.scala:497)
	at org.scalatest.FunSpecLike.runTests$(FunSpecLike.scala:496)
	at org.scalatest.FunSpec.runTests(FunSpec.scala:1630)
	at org.scalatest.Suite.run(Suite.scala:1147)
	at org.scalatest.Suite.run$(Suite.scala:1129)
	at org.scalatest.FunSpec.org$scalatest$FunSpecLike$$super$run(FunSpec.scala:1630)
	at org.scalatest.FunSpecLike.$anonfun$run$1(FunSpecLike.scala:501)
	at org.scalatest.SuperEngine.runImpl(Engine.scala:521)
	at org.scalatest.FunSpecLike.run(FunSpecLike.scala:501)
	at org.scalatest.FunSpecLike.run$(FunSpecLike.scala:500)
	at org.scalatest.FunSpec.run(FunSpec.scala:1630)
	at org.scalatest.tools.SuiteRunner.run(SuiteRunner.scala:45)
	at org.scalatest.tools.Runner$.$anonfun$doRunRunRunDaDoRunRun$13(Runner.scala:1340)
	at org.scalatest.tools.Runner$.$anonfun$doRunRunRunDaDoRunRun$13$adapted(Runner.scala:1334)
	at scala.collection.immutable.List.foreach(List.scala:389)
	at org.scalatest.tools.Runner$.doRunRunRunDaDoRunRun(Runner.scala:1334)
	at org.scalatest.tools.Runner$.$anonfun$runOptionallyWithPassFailReporter$24(Runner.scala:1031)
	at org.scalatest.tools.Runner$.$anonfun$runOptionallyWithPassFailReporter$24$adapted(Runner.scala:1010)
	at org.scalatest.tools.Runner$.withClassLoaderAndDispatchReporter(Runner.scala:1500)
	at org.scalatest.tools.Runner$.runOptionallyWithPassFailReporter(Runner.scala:1010)
	at org.scalatest.tools.Runner$.run(Runner.scala:850)
	at org.scalatest.tools.Runner.run(Runner.scala)
	at org.jetbrains.plugins.scala.testingSupport.scalaTest.ScalaTestRunner.runScalaTest2(ScalaTestRunner.java:138)
	at org.jetbrains.plugins.scala.testingSupport.scalaTest.ScalaTestRunner.main(ScalaTestRunner.java:28)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:147)
Caused by: uk.ac.wellcome.exceptions.GracefulFailureException: Unrecognised identifier type: [foo]
	at uk.ac.wellcome.models.work.internal.IdentifierType$.lookupPlatformId(IdentifierType.scala:52)
	at uk.ac.wellcome.models.work.internal.IdentifierType$.<init>(IdentifierType.scala:64)
	at uk.ac.wellcome.models.work.internal.IdentifierType$.<clinit>(IdentifierType.scala)
	... 51 more
```

It’s not quite as good as a compile-time error, but I think it gives us some protection against trying to look up identifier schemes that don‘t exist.

Open to alternative suggestions, but this was the easiest thing I could find and think to do.

sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/IdentifierType.scala
```diff
@@ -0,0 +1,83 @@
...
+          new RuntimeException(s"Unrecognised identifier type: [$id]")
+        )
+    }
+
+  // This isn't ideal, because we're duplicating the list from the CSV,
+  // but it gives us a bit of type-safety.
+  //
+  // If any of these values don't resolve, we fail immediately upon
+  // startup -- not when we try to load the value!
+  //
```
mklander: I agree with the comment here, I'm not sure what duplicating the entries from the CSV gives us.  If there is a mistake or omission it will fail at runtime, which is the same as if we load the CSV at instantiation and trust it.  For the larger files (e.g., location-types) this seems to be the best option anyway.

alicefuzier: If we are going down the route of reading the csv files, being fully dynamic and trusting that they are correct might be ok. 
This check at startup basically forces us to duplicate the information in the csv file and if we do that we could just as well duplicate it by keeping the enum and updating the list of allowed values in the enum.
I think especially for big csv like the location-types one, this check is not really a viable option (nor is the enum) and in that case, we are forced to trust those files anyway. 
Also, it's very easy to forget to add new things to these lines as they get added to the csv file, so there is actually no real certainty that the application doesn't start if the csv file is somehow invalid.

alexwlchan: It fails at runtime, but it fails as soon as you instantiate this object, which I think happens at import, i.e. startup time?

As opposed to failing when you do the query.

If you forget to add an identifier type from the CSV to this list, you can’t used it elsewhere in the codebase, so I don’t think that’s an issue.

The alternative is:

* Everywhere we use an instance of IdentifierSchemes right now, we have to do `lookupPlatformId("MiroImageNumber")` with a hard-coded string
* Something with Shapeless that generates these values for us (but then IntelliJ might not be very happy). Open to suggestions.

alicefuzier: I don't think you can do it with shapeless. You probably can with macros but I don't think we want to go down that route. 

I don't think this gives us many advantages over just keeping the enum up to date and it's not a viable option for csvs that change often and/or are too big.
I know that this would fail at startup rather than while processing a message. I just don't think that failure while processing the message would be that bad if it give us the ability to just load the content of this file into a map, wihtout having duplicated information.

alexwlchan: So expose `lookupPlatformId` as a public (and renamed) method, use that everywhere, and drop the constants below?

alicefuzier: I think so, but I'd like @kenoir and @mklander opinion

jtweed:
The values in the CSV don't match what we agreed in #2030.

sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/IdentifierType.scala
alexwlchan: I think it’s a Singleton, yes.

alexwlchan:
I’ve rolled back to a model where we look up strings every time. I’ll put up another patch to illustrate what that might look like.

I also pulled out the identifier-type specific stuff from the CSV parsing, because we can reuse that for license/location types.

sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/utils/CSVReferenceTypeManager.scala
```diff
@@ -0,0 +1,53 @@
...
+  //    2,acqi,Info Service acquisitions
+  //    3,acql,Wellcome Library
+  //    4,admi,administration
+  //
+  // The first entry is an identifier we can use to identify a row
+  // within a reference file.  The second and third entries are the ID
+  // and label we show in public ID schemes.
+  //
+  private val csvRows: Array[(String, String, String)] = Source
+    .fromInputStream(stream)
```
mklander: Should we hold the desired resource path, and create the stream where used?
`private val resourcePath = path`
`Try(Source.fromResource(resource)...`

sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/IdentifierType.scala
```diff
@@ -0,0 +1,21 @@
...
+
+case class IdentifierType(
+  id: String,
+  label: String,
+  ontologyType: String = "IdentifierType"
+)
+
+case object IdentifierType {
+  private val manager = new CSVReferenceTypeManager("/identifier-schemes.csv")
+
```
mklander: Should we also have a test for `CSVReferenceTypeManager` for the case where the resource doesn't exist?

sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/IdentifierType.scala
```diff
@@ -0,0 +1,21 @@
...
+import uk.ac.wellcome.models.utils.CSVReferenceTypeManager
+
+case class IdentifierType(
+  id: String,
+  label: String,
+  ontologyType: String = "IdentifierType"
+)
+
+case object IdentifierType {
```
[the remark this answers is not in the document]
alexwlchan: Blargh. That’s annoying.

alexwlchan:
I think there’s a rough consensus that looking up values by string at runtime is the best we can do for now, so I’m going to merge this and work on fixing everything else to use it.

PR #2136, opened 2018-05-24
VHS provides metadata at record update time

kenoir:

For https://github.com/wellcometrust/platform/issues/1810

In order to make decisions about whether to update stored files it is necessary to store metadata about those files. When using the `InputStream` store this is essential as we cannot easily inspect the stored object to make decisions about it.

This change modifies the hybrid store to surface metadata about stored items when they are being modified.

### Who is this change for?

🍖 🗄

### DIFF
--- a/sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/vhs/VersionedHybridStore.scala
+++ b/sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/vhs/VersionedHybridStore.scala
@@ -11,21 +11,22 @@ import uk.ac.wellcome.storage.GlobalExecutionContext.context
 
 import scala.concurrent.Future
 
-class VersionedHybridStore[T, Store <: S3ObjectStore[T]] @Inject()(
+case class EmptyMetadata()
+
+class VersionedHybridStore[T, Metadata, Store <: S3ObjectStore[T]] @Inject()(
   vhsConfig: VHSConfig,
   s3ObjectStore: Store,
   dynamoDbClient: AmazonDynamoDB
 ) {
-
   val versionedDao = new VersionedDao(
     dynamoDbClient = dynamoDbClient,
     dynamoConfig = vhsConfig.dynamoConfig
   )
 
-  case class EmptyMetadata()
   private case class VersionedHybridObject(
     hybridRecord: HybridRecord,
-    s3Object: T
+    s3Object: T,
+    metadata: Metadata
   )
 
   // Store a single record in DynamoDB.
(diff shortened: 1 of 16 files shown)


sbt_common/storage/src/test/scala/uk/ac/wellcome/storage/test/fixtures/LocalVersionedHybridStore.scala
```diff
@@ -39,10 +39,10 @@ trait LocalVersionedHybridStore
       "aws.vhs.dynamo.tableName" -> table.name
     ) ++ s3ClientLocalFlags ++ dynamoClientLocalFlags
 
-  def withTypeVHS[T <: Id, R](bucket: Bucket,
+  def withTypeVHS[T <: Id, M, R](bucket: Bucket,
```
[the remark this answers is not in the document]
kenoir: Descriptive names are good, but I did notice that it makes it harder to tell when types are inferred rather than explicit.

kenoir:
@alexwlchan there is a test at https://github.com/wellcometrust/platform/pull/2136/files#diff-7fceedc009877f1fcdc644863e4ecc8fR150 that replaces what's stored with the metadata with what's in the store to avoid using a `var` and to ensure that the exact metadata that is stored is handed to `ifExisting`.

sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/vhs/VersionedHybridStore.scala
```diff
@@ -77,10 +83,19 @@ class VersionedHybridStore[T, Store <: S3ObjectStore[T]] @Inject()(
...
-  def getRecord(id: String): Future[Option[T]] =
-    getObject[HybridRecord](id).map { maybeObject =>
+  def getRecord[DynamoRow](id: String)(
+    implicit enricher: HybridRecordEnricher.Aux[Metadata, DynamoRow],
+    dynamoFormat: DynamoFormat[DynamoRow],
+    migrationH: Migration[DynamoRow, HybridRecord],
+    migrationM: Migration[DynamoRow, Metadata]
+  ): Future[Option[T]] = {
+
+    identity(enricher)
```
alicefuzier: A comment explaining the reason for this could be helpful?

sbt_common/storage/src/test/scala/uk/ac/wellcome/storage/vhs/StringStoreVersionedHybridStoreTest.scala
```diff
@@ -106,38 +108,67 @@ class StringStoreVersionedHybridStoreTest
...
+          withLocalDynamoDbTable { table =>
+            withStringVHS[ExtraData, Assertion](bucket, table) { hybridStore =>
+              val updatedRecord = Random.nextString(256)
+
+              val future =
+                hybridStore.updateRecord(id)(record)((t, _) => t)(data)
+
+              val updatedFuture = future.flatMap { _ =>
+                hybridStore.updateRecord(id)(updatedRecord)((t, m) =>
+                  m.toString)(data)
```
alexwlchan: Using `storedMetadata` for the variable name here would help to emphasise that we’re getting the metadata which is already stored.

PR #2138, opened 2018-05-25
Changes to the display library don't trigger builds for (some) pipeline applications

alexwlchan:

If the only changes are to a display library, and we're testing an application that has no public-facing bits (i.e. catalogue pipeline, Goobi and Sierra adapters), we can shortcut running the tests.

This is in anticipation of some changes to the display models following #2134, when I modify DisplayIdentifier to match our new model.

### Who is this change for?

🏈 🎫


alexwlchan:
Plus a few tweaks to the travistooling tests so they're easier to run locally.

PR #2140, opened 2018-05-25
Split DisplayIdentifier into V1 and V2 versions

alexwlchan:

Changing the Identifier model means changing DisplayIdentifier in non-backwards compatible ways, so we need to fork the display model.

Because DisplayIdentifier is used in several other places, this means also forking AbstractAgent and Item. This patch does that forking – but with no changes to DisplayIdentifier, they'll be a separate patch. This is quite large enough on its own!

### Who is this change for?

🍴 📛


alexwlchan:
This should be merged after #2134.

PR #2153, opened 2018-05-30
Work on location and data storage model

silveroliver:


silveroliver:
@jtweed this includes an example of the storage model. Your orginal file updated to reflect new model.

PR #2158, opened 2018-05-31
Rename LinkedWorkUpdate to WorkUpdate

alexwlchan:
Continuing the renaming thread from #2155.


alexwlchan:
Yep, I’ll do that in another patch and then trigger a reindex.

PR #2160, opened 2018-05-31
Replace IdentifierList with MatchedIdentifiers + new namespace

alexwlchan:
This model needs to be shared with the merger, so it shouldn't live in the matcher namespace.

Additionally, I think "EquivalentIdentifiers" better express the notion that this isn't just an arbitrary collection, but it's a collection of related identifiers.

More renaming for #2155.

### DIFF
--- a/catalogue_pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/models/LinkedWorksIdentifiersList.scala
+++ b/catalogue_pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/models/LinkedWorksIdentifiersList.scala
@@ -1,3 +1,5 @@
 package uk.ac.wellcome.platform.matcher.models
 
-case class LinkedWorksIdentifiersList(linkedWorks: Set[IdentifierList])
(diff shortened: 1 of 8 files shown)


catalogue_pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/models/LinkedWorksIdentifiersList.scala
```diff
@@ -1,3 +1,5 @@
 package uk.ac.wellcome.platform.matcher.models
 
-case class LinkedWorksIdentifiersList(linkedWorks: Set[IdentifierList])
+import uk.ac.wellcome.models.matcher.MatchedIdentifiers
+
+case class LinkedWorksIdentifiersList(linkedWorks: Set[MatchedIdentifiers])
```
alicefuzier: `linkedWorks`: should this change too?

alexwlchan: yep, that's in the PR which ditches LinkedWorksIdentifiersList

PR #2166, opened 2018-05-31
Add a script for getting the current versions of the API

alexwlchan:
This script tells you which version of the API is running, and the current pins in ECS. Screenshot:

screen shot 2018-05-31 at 15 25 47

The idea is that when you want to switch the API, you can run this script to know what to use for pins on the about-to-be-prod API, saving you the trouble of looking them up in ECS.

It _doesn’t_ edit any of the Terraform files for you, because that feels a bit more fiddly.

### DIFF
--- a/catalogue_api/get_api_versions.py
+++ b/catalogue_api/get_api_versions.py
@@ -70,3 +70,41 @@ if __name__ == '__main__':
     print(f'The staging API is {bold(staging_api)}')
     print(f'- api   = {bold(staging_api_info["api"])}')
     print(f'- nginx = {bold(staging_api_info["nginx_api"])}')
+
+    print('')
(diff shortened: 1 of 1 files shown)


alexwlchan:
Okay, so it doesn’t edit your terraform… but it can tell you what Terraform you should be writing!

screen shot 2018-05-31 at 15 35 22

alexwlchan:
I’m going to merge this as-is, but yes, a manifest endpoint would be a nice future enhancement.

How do you tell the applications what version they’re running? Configure flag?

alexwlchan:
Also, the applications would have to know they’re running nginx, and present that version (or somehow know there were other apps to talk about), which feels like a bit of a boundary violation.

PR #2171, opened 2018-06-01
VHS: update uses metadata in both existing and not existing cases 

mklander:
VHS update takes metadata for both existing and non-existing cases giving the caller control over what metadata should be saved in each case.


sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/vhs/VersionedHybridStore.scala
```diff
@@ -35,8 +36,11 @@ class VersionedHybridStore[T, Metadata, Store <: S3ObjectStore[T]] @Inject()(
   // The HybridRecordEnricher combines this with the HybridRecord, and stores
   // both of them as a single row in DynamoDB.
   //
-  def updateRecord[DynamoRow](id: String)(ifNotExisting: => T)(
-    ifExisting: (T, Metadata) => T)(metadata: Metadata)(
+  // Update and version logic (e.g., do not store this record if a newer record already exists)
+  // is handled through the mappings ifNotExisting and ifExisting
```
alexwlchan: 👍

PR #2176, opened 2018-06-01
Fix the Terraform config for the matcher table

alexwlchan:
Required now that we’ve renamed the parameters on WorkNode.

Also, this puts the table definition in its own file – in general, we prefer to have each database/S3 bucket in a separate file to reduce the risk of making changes to the wrong thing.


kenoir:
Just leaving a note that this is a destructive change, I know we don't care right now but good to know.

PR #2114, opened 2018-05-22
[WIP] Better separate storage concerns

kenoir:
### What is this PR trying to achieve?

Experimenting with restructuring storage a bit, in order to separate concerns more effectively.

### Who is this change for?

🎏

PR #2178, opened 2018-06-02
Storage library patch

kenoir:

Simplify the storage library, removing the need for multiple classes extending `ObjectStore`.

This change uses an `ObjectStore` type class to replace existing `StringStore`, `StreamStore` and `TypeStore` classes.

An `ObjectStore` implicitly requires a `StorageStrategy`:

```scala
trait StorageStrategy[T] {
  def store(t: T): StorageStream
  def retrieve(input: InputStream): Try[T]
}
```

Utilising `circe` we are able to provide an implicit conversion: `T -> Json ->  InputStream`

And a `StorageBackend`:

```scala
trait StorageBackend {
  def put(location: ObjectLocation,
          input: InputStream,
          metadata: Map[String, String]): Future[Unit]
  def get(location: ObjectLocation): Future[InputStream]
}
```

Thus change provides an `S3StorageBackend` implementation, allowing the possibility of using other storage back ends (for example Azure Blob Store).

### Who is this change for?

Developers who want maintainable and extensible code with properly separated concerns.

- [ ] Deployed new versions
- [ ] Run `terraform apply`.


kenoir:
**Large PR Warning:**

This is a large change resulting from refactoring experiments in https://github.com/wellcometrust/platform/pull/2114. Performing this change in smaller increments though preferable is IMHO unlikely to offer great benefit at this point especially given the high build contention.

catalogue_pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/modules/ExecutionContextModule.scala
```diff
@@ -0,0 +1,15 @@
...
+import com.twitter.inject.{Injector, TwitterModule}
+import javax.inject.Singleton
+import uk.ac.wellcome.platform.idminter.GlobalExecutionContext
+
+import scala.concurrent.ExecutionContext
+
+object ExecutionContextModule extends TwitterModule {
+  @Provides
+  @Singleton
+  def provideExecutionContext(injector: Injector): ExecutionContext =
```
alicefuzier: Since we are writing Guice providers, could we make it get the excution context from the `ActorSystem` rather than from the `GlobalExecutionContext`? Somethink like 
```
def provideExecutionContext(actorSystem: ActorSystem): ExecutionContext = actorSystem.dispatcher
```
I think it would be good to deprecate those `GlobalExecutionContext` and this might be a step towards that

catalogue_pipeline/terraform/dynamo.tf
```diff
@@ -5,18 +5,18 @@ resource "aws_dynamodb_table" "matcher_graph_table" {
   hash_key       = "workId"
 
   attribute {
-    name = "id"
+    name = "workId"
```
alicefuzier: mmh, is this supposed to be here?

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/modules/TransformablesModule.scala
```diff
@@ -0,0 +1,48 @@
...
+import uk.ac.wellcome.storage.s3.S3StorageBackend
+
+import uk.ac.wellcome.utils.JsonUtil._
+
+import scala.concurrent.ExecutionContext
+
+object TransformablesModule extends TwitterModule {
+  @Provides
+  @Singleton
+  def provideMiroTransformable(
```
alicefuzier: `provideMiroTrabsformableObjectStore`?

sbt_common/common/src/main/scala/uk/ac/wellcome/utils/JsonUtil.scala
```diff
@@ -5,6 +5,7 @@ import io.circe.generic.extras.{AutoDerivation, Configuration}
 import io.circe.java8.time.TimeInstances
 import io.circe.parser._
 import io.circe.syntax._
+import io.circe._
```
alicefuzier: ??

sbt_common/storage/src/main/scala/uk/ac/wellcome/storage/ObjectStore.scala
```diff
@@ -0,0 +1,67 @@
...
+    implicit storageStrategy: StorageStrategy[T],
+    storageBackend: R,
+    ec: ExecutionContext): ObjectStore[T] = new ObjectStore[T] {
+    def put(namespace: String)(
+      t: T,
+      keyPrefix: KeyPrefix = KeyPrefix(""),
+      keySuffix: KeySuffix = KeySuffix(""),
+      userMetadata: Map[String, String] = Map()
+    ): Future[ObjectLocation] = {
+      val storageStream = storageStrategy.store(t)
```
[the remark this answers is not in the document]
alicefuzier: Yeah, sounds better. Also the method names could be clearer. `store` and `retrieve` are very confusing to me. Maybe `toStream` and `fromStream`?

catalogue_pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/modules/ExecutionContextModule.scala
kenoir: I generally agree, but making that change in this patch might be a step too far. I'd prefer to make this update in a future patch.

alicefuzier: definitely don't get rid of the GlobalExecutionContexts here, but I would avoid adding an extra usage of them in this patch. Making the provider use the actor system will at least not worsen the position we are in now. Also, delegating all scheduling responsibilities to the actor system is probably more reliable

PR #2180, opened 2018-06-03
[WIP] Storage path VHS update

kenoir:
...


kenoir:
Quick experiment to see how easy it would be to merge storage patch after vhs changes. Turns out, not too bad. Closing.

PR #2187, opened 2018-06-04
Use a for comprehension in the reindexer

alexwlchan:
As discussed with @kenoir, this use of a for comprehension means we're not wrapping a Future in a Future and returning early.

### DIFF
--- a/reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex_worker/services/ReindexService.scala
+++ b/reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex_worker/services/ReindexService.scala
@@ -58,9 +58,11 @@ class ReindexService @Inject()(dynamoDbClient: AmazonDynamoDB,
 
     // Then we PUT all the records.  It might be more efficient to do a
     // bulk update, but this will do for now.
-    outdatedRecordsFuture.flatMap {
-      outdatedRecords: List[ReindexRecord] => Future.sequence {
-        outdatedRecords.map { updateIndividualRecord(_, desiredVersion = reindexJob.desiredVersion) }
+    outdatedRecordsFuture.flatMap { outdatedRecords: List[ReindexRecord] =>
+      Future.sequence {
+        outdatedRecords.map {
+          updateIndividualRecord(_, desiredVersion = reindexJob.desiredVersion)
+        }
       }
     }
   }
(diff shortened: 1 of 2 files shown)


reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex_worker/services/ReindexService.scala
```diff
@@ -28,47 +30,53 @@ class ReindexService @Inject()(dynamoDbClient: AmazonDynamoDB,
...
-      outdatedRecords
-        .map { record: ReindexRecord =>
-          val updatedRecord =
-            record.copy(reindexVersion = reindexJob.desiredVersion)
-          versionedDao.updateRecord[ReindexRecord](updatedRecord)
-        }
-        .map { _ =>
-          ()
-        }
+      _ <- outdatedRecords.map {
```
alicefuzier: Don't you need a Future.sequence here? This creates a List[Future]], so I don't think this waits on the completion of individual updates. Are there any tests that check what happens if an individual update fails? If not we should add one

alexwlchan: Yep, yep. Test added, and I’ve got the Future.sequence stuff working as well.

reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex_worker/services/ReindexService.scala
```diff
@@ -28,47 +30,58 @@ class ReindexService @Inject()(dynamoDbClient: AmazonDynamoDB,
...
-          warn(s"Failed to read Dynamo records: $err")
-          throw GracefulFailureException(
-            new RuntimeException(s"Error in the DynamoDB query: $err")
-          )
+      outdatedRecords: List[ReindexRecord] = results.map(extractRecord)
+    } yield outdatedRecords
+
+    // Then we PUT all the records.  It might be more efficient to do a
+    // bulk update, but this will do for now.
+    outdatedRecordsFuture.flatMap { outdatedRecords: List[ReindexRecord] =>
```
kenoir: What happened to the for comprehension?

alexwlchan: When I tried this earlier, I couldn’t find a way to put this inside the for comprehension that didn't cause the compiler to think I was returning Future[Nothing] or Future[Unit], rather than Future[List[Unit]].

kenoir: Let's go through it tomorrow, I think we should be able to use a for comprehension here. For now we can merge.

Issue #2177, opened 2018-06-01
Reduce our CloudWatch costs


alexwlchan:
We spent a bit much on CloudWatch last month; it would be nice to make these numbers more sensible. This is a tracking ticket for ways to reduce our bills:

- [x] Only log the Lambda event when we have an error: https://github.com/wellcometrust/aws_utils/pull/18
- [x] Only log SNS responses in Lambdas when we have an error: https://github.com/wellcometrust/aws_utils/pull/19
- [x] Don’t log an enormous stack trace for ConditionalCheckFailed exceptions:

    ```
    java.lang.RuntimeException: ConditionNotMet(com.amazonaws.services.dynamodbv2.model.ConditionalCheckFailedException: The conditional request failed (Service: AmazonDynamoDBv2; Status Code: 400; Error Code: ConditionalCheckFailedException; Request ID: 0FEMQRMOTFA1F4BV2G021JS9V7VV4KQNSO5AEMVJF66Q9ASUAAJG))
        at uk.ac.wellcome.storage.dynamo.VersionedDao.$anonfun$updateRecord$3(VersionedDao.scala:65)
        at uk.ac.wellcome.storage.dynamo.VersionedDao.$anonfun$updateRecord$3$adapted(VersionedDao.scala:62)
        at scala.Option.map(Option.scala:146)
        at uk.ac.wellcome.storage.dynamo.VersionedDao.$anonfun$updateRecord$1(VersionedDao.scala:62)
        at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
        at scala.concurrent.Future$.$anonfun$apply$1(Future.scala:654)
        at scala.util.Success.$anonfun$map$1(Try.scala:251)
    00:13:32.118 [ForkJoinPool-4-worker-14] WARN  u.a.w.storage.dynamo.VersionedDao - Failed to updating Dynamo record: sierra/1832158
        at scala.concurrent.Future.$anonfun$map$1(Future.scala:288)
        at scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:29)
        at scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:29)
        at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:60)
        at java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1402)
[...]

PR #2194, opened 2018-06-05
timeAndCount -> count & reduce use

kenoir:

Reduce the use of CloudWatch metrics where they are not useful and achieve cost saving from this.

### Who is this change for?

🤑 

- [ ] Deployed new versions
- [ ] Run `terraform apply`.

### DIFF
--- a/catalogue_pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/IdEmbedder.scala
+++ b/catalogue_pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/IdEmbedder.scala
@@ -12,8 +12,8 @@ import scala.annotation.tailrec
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success}
 
-class IdEmbedder @Inject()(
-  identifierGenerator: IdentifierGenerator)(implicit ec: ExecutionContext)
+class IdEmbedder @Inject()(identifierGenerator: IdentifierGenerator)(
+  implicit ec: ExecutionContext)
     extends Logging {
 
   def embedId(json: Json): Future[Json] = Future {
--- a/catalogue_pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/IdEmbedder.scala
+++ b/catalogue_pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/IdEmbedder.scala
@@ -6,7 +6,6 @@ import io.circe.optics.JsonPath.root
 import io.circe.optics.JsonTraversalPath
 import io.circe.{Json, _}
 import uk.ac.wellcome.models.work.internal.SourceIdentifier
-import uk.ac.wellcome.monitoring.MetricsSender
 import uk.ac.wellcome.utils.JsonUtil._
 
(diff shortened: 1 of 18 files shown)


catalogue_pipeline/id_minter/src/main/scala/uk/ac/wellcome/platform/idminter/steps/IdEmbedder.scala
```diff
@@ -18,14 +18,8 @@ class IdEmbedder @Inject()(
...
-  def embedId(json: Json): Future[Json] = {
-    metricsSender.timeAndCount(
-      "generate-id",
-      () =>
-        Future {
-          iterate(root.each, addIdentifierToJson(json))
-      }
-    )
+  def embedId(json: Json): Future[Json] = Future {
+    iterate(root.each, addIdentifierToJson(json))
```
alexwlchan: You can remove metricsSender from IdEmbedder entirely now.

catalogue_pipeline/ingestor/src/main/scala/uk/ac/wellcome/platform/ingestor/services/WorkIndexer.scala
```diff
@@ -30,44 +30,35 @@ class WorkIndexer @Inject()(
...
-  def indexWork(work: IdentifiedWork,
-                esIndex: String,
-                esType: String): Future[Any] = {
+  def indexWork(work: IdentifiedWork, esIndex: String, esType: String) = {
 
-    metricsSender.timeAndCount[Any](
-      "ingestor-index-work",
-      () => {
-        info(s"Indexing work ${work.canonicalId}")
+    info(s"Indexing work ${work.canonicalId}")
```
alexwlchan: And WorksIndexer. (And probably others, I won’t keep checking.)

sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/sqs/SQSStream.scala
```diff
@@ -67,10 +67,6 @@ class SQSStream[T] @Inject()(actorSystem: ActorSystem,
         logger.error(
           s"Unrecognised failure while processing message ${message.getMessageId}",
           exception)
-        metricsSender.incrementCount(
-          metricName = s"${streamName}_MessageProcessingFailure",
-          count = 1.0
-        )
```
alexwlchan: Okay, maybe we should keep this, oops.

alicefuzier: If this goes away, We can get rid of GracefulFailureException cause this is the only difference in behaviour between the two types of exception

alexwlchan: Ah no, it's covered above.

alexwlchan: There’s also the difference in logging – a GracefulFailureException logs a WARN and a message, whereas an unrecognised exception includes the full stack trace for extra debugging.

alexwlchan:
Part of #2177.

PR #2198, opened 2018-06-06
Bulk insert in the ingestor

alicefuzier:
🕋 💉


alicefuzier:
Superseeded by #2228 and #2229

PR #2199, opened 2018-06-06
Make IngestorWorkerServiceTest test the whole class

alicefuzier:
⌛️ 🤥


catalogue_pipeline/ingestor/src/test/scala/uk/ac/wellcome/platform/ingestor/services/IngestorWorkerServiceTest.scala
```diff
@@ -184,30 +219,31 @@ class IngestorWorkerServiceTest
...
-            messageStream = messageStream,
-            system = actorSystem
-          )
-
-          testWith(service)
+    testWith: TestWith[(QueuePair, Bucket), R]): R = {
+    withLocalSqsQueueAndDlq {
+      case queuePair @ QueuePair(queue, dlq) =>
+        withLocalS3Bucket { bucket =>
+          withMessageStream[IdentifiedWork, R](
```
alexwlchan: This fixture feels like a weird hybrid – some things get passed in that are constructor by external fixtures (ActorSystem, MetricsSender), others are defined internally (queuePair, S3 bucket).

We should be consistent – does this fixture do everything, or nothing?

PR #2201, opened 2018-06-06
Reduce failures in the Miro transformer – check for nulls in the image_keywords_unauth field

alexwlchan:
10 images are failing in the transformer because the field contains nulls, and we expect List[String]. This patch makes the transformer more resilient to these fields.

### DIFF
--- a/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/miro/MiroSubjects.scala
+++ b/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/miro/MiroSubjects.scala
@@ -20,6 +20,10 @@ trait MiroSubjects {
     val keywords: List[String] = miroData.keywords.getOrElse(List())
 
     val keywordsUnauth: List[String] =
+      miroData.keywordsUnauth match {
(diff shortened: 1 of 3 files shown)


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/miro/MiroSubjects.scala
```diff
@@ -20,7 +20,11 @@ trait MiroSubjects {
     val keywords: List[String] = miroData.keywords.getOrElse(List())
 
     val keywordsUnauth: List[String] =
-      miroData.keywordsUnauth.getOrElse(List())
+      miroData.keywordsUnauth match {
+        case Some(maybeKeywords) => maybeKeywords.flatten
+        case None => List()
+      }
+    miroData.keywordsUnauth.getOrElse(List())
```
alicefuzier: do you still need this line?

alexwlchan: Nope, should be gone. Thanks!

PR #2203, opened 2018-06-06
Simplify the interface of the reindexer script

alexwlchan:
The script for triggering a reindexer is a little wonky: you need to know what the _current_ best desiredVersion is before you can trigger your index. Which means you need to go look at the DynamoDB table, or remember what it was on a previous run. LIKE AN ANIMAL.

Now the script looks in the table at the current desiredVersion in the records you’re about to reindex, finds the maximum and adds one.

The interface goes from:

```console
$ python reindexer/manage_reindex.py update-shards --prefix=miro --desired_version=36
```

to:

```console
$ python reindexer/trigger_reindex.py --prefix=miro
```

### DIFF
--- a/reindexer/manage_reindex.py
+++ b/reindexer/manage_reindex.py
@@ -56,22 +56,19 @@ def _update_shard(client, table_name, shard):
     )
 
 
+
+def _all_records_in_shard(client, table_name):
+    """Generates every row in a particular shard."""
+    paginator = client.get_paginator('scan')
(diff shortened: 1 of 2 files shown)


alexwlchan:
You are, or am I?

What do you do about the five minute execution timeout? I’m not entirely sure how long it takes to trigger a full Sierra reindex, but when we have all the Goobi METS, it'll definitely be too big.

PR #2210, opened 2018-06-07
[WIP] Workaround for a JSON decoding bug in the Miro transformer

alexwlchan:

Nine records are falling out of the Miro transformer because we screwed up the XML-to-JSON conversion in a way that makes Circe miserable. Specifically, we have a mixture of:

    "image_source_code": "ABC"

and

    "image_source_code":["GUS","TO-DELETE"]

which appears in that exact form.

We need to address it (so we don't get stuff on the DLQs), but this may not be the final fix. I’ve emailed Christy to find out if we want to just drop these images entirely, or pass them through with GUS as the sole contributor.

### DIFF
--- a/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/source/MiroTransformableData.scala
+++ b/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/source/MiroTransformableData.scala
@@ -68,7 +68,7 @@ case object MiroTransformableData {
         "image_source_code":"GUS"
       """.trim
     )
-    
+
     val tryMiroTransformableData =
       createMiroTransformableData(tidiedData)
 
--- a/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/source/MiroTransformableData.scala
+++ b/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/source/MiroTransformableData.scala
@@ -54,8 +54,25 @@ case object MiroTransformableData {
   def create(data: String): MiroTransformableData = {
     val unescapedData = unescapeHtml(data)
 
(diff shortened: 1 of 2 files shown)


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/source/MiroTransformableData.scala
```diff
@@ -54,8 +54,23 @@ case object MiroTransformableData {
...
 
+    // We have a very small number of records where the image_source_code
+    // field is a list, not a string.  They all contain the value
+    // ["GUS", "TO-DELETE"], which causes a JSON decoding error.  We could
+    // define a custom Circe decoder for this, but that's more complicated
+    // than it's worth!  A string substitution is a good enough workaround.
+    //
+    val tidiedData = unescapedData.replaceAllLiterally(
+      """
+        "image_source_code":["GUS","TO-DELETE"]
```
alicefuzier: oh wait, I didn't realise those were the ones causing issues. I modified the source data because Hannah asked us to remove them and I didn't want them to keep reappearing with reindexes. 
I modified them so that they deliberately didn't match our schema. Maybe that wasn't such a good idea..

alexwlchan: Since the change wasn’t documented anywhere – yeah, I think that was a mistake.

If you’re going to edit the source data (which is a bad idea for other reasons, but) make sure it’s super-duper clear when looking at the data why it’s changed – we don’t have commit messages for that stuff. _“This is deliberately broken because we don’t want to expose these images”_ is better than _“TO-DELETE”_, which could be a source artefact or could be something we put in.

alexwlchan:
Superseded by #2212.

PR #2212, opened 2018-06-08
Dont transform Miro records with the GUS contributor code

alexwlchan:
Replaces #2210.

This is better than putting a hand-coded fix in the source JSON because here the change is documented, visible, and permanent. If we ever re-run the Miro preprocessor pipeline (accidentally or on purpose), we’d lose our hand-coded fix.

Once this change is merged, I’ll go back and fix what’s in the VHS, so we have consistent data and these records don’t just land on the transformer DLQs.


alexwlchan:
I haven’t done it yet, I was planning to do it when I get into the office.

Issue #2213, opened 2018-06-08
Get Miro reindexes to pass through the transformer cleanly


alexwlchan:
We’ve done a bunch of reindexes to change the internal model, we should really get this just passing.

So far includes: #2200, #2201, #2210, #2212.

alexwlchan:
I ran another Miro reindex this morning, which completed successfully. The only item on the transformer DLQ is a Sierra record, so I think this is done!

PR #2220, opened 2018-06-08
Handle a malformed INNOPAC ID in the Miro data

alexwlchan:
One more step towards completing #2213.


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/miro/MiroIdentifiers.scala
```diff
@@ -0,0 +1,89 @@
...
+        value = miroId)
+    )
+
+    // Add the Sierra system number from the INNOPAC ID, if it's present.
+    //
+    // We add a b-prefix because everything in Miro is a bibliographic record,
+    // but there are other types in Sierra (e.g. item, holding) with matching
+    // IDs but different prefixes.
+
+    // First, a note: one of the Miro records has a bit of weird data in
```
alexwlchan: This is the interesting bit – the rest is just copied from the old transformer code.

PR #2222, opened 2018-06-11
Don’t log a complete traceback for expected DynamoDB errors

alexwlchan:
Related to #2177.

Running a reindex on Sunday cost ~£80, most of which was tracebacks from DynamoDB errors inside the VersionedDao. This patch modifies the Dao to only emit GracefulFailureExceptions in this case, which SQSStream logs as a single line.

Additionally, it removes SQSToDynamoStream, which was only being used to suppress these errors.


alexwlchan:
This is going to be migrated to scala-storage, but porting it is blocked on #2227.

alexwlchan:
Replaced by #2249.

PR #2227, opened 2018-06-12
Migrate all our applications to use scala-storage

alexwlchan:
A new and terrifying age of library management is upon us.

We now have scala-storage outside the main repo, on an independent CI and release cycle. This patch takes the existing storage code from the platform, and unceremoniously evicts it with a large 👢. This allows us to remove a significant portion of code from the repo, and save a build job on every Travis run.

Next up is probably monitoring, but let’s see how this goes first.


alexwlchan:
This is now green; merging by administrative fiat to get it into master and avoid another round of merge conflicts.

PR #2233, opened 2018-06-14
Fix problems with spaces in the s3 key

alicefuzier:
For #2230


goobi_adapter/goobi_reader/src/main/scala/uk/ac/wellcome/platform/goobi_reader/services/GoobiReaderWorkerService.scala
```diff
@@ -52,11 +52,12 @@ class GoobiReaderWorkerService @Inject()(
 
   private def updateRecord(r: S3Record) = {
     val bucketName = r.s3.bucket.name
-    val objectKey = r.s3.`object`.key
+    val objectKey = java.net.URLDecoder.decode(r.s3.`object`.key, "utf-8")
```
alexwlchan: A brief explanation of why we do the URL decoding would be helpful, ideally including an example of the sort of key that broke it the first time.

alexwlchan: Hmm. I also wonder if this is in the right place? Wouldn’t it be better to do the URL decoding as soon as we get the event from S3?

PR #2236, opened 2018-06-14
Merger skeleton

alicefuzier:
🏛 🎲

### DIFF
--- a/sbt_common/messaging/src/test/scala/uk/ac/wellcome/messaging/test/fixtures/Messaging.scala
+++ b/sbt_common/messaging/src/test/scala/uk/ac/wellcome/messaging/test/fixtures/Messaging.scala
@@ -112,25 +112,25 @@ trait Messaging
     testWith(stream)
   }
 
-  def withMessageStreamFixtures[R](
-    testWith: TestWith[(Bucket,
-    MessageStream[ExampleObject],
+  def withMessageStreamFixtures[T,R](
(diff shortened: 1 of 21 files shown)


sbt_common/messaging/src/test/scala/uk/ac/wellcome/messaging/test/fixtures/Messaging.scala
```diff
@@ -112,25 +112,25 @@ trait Messaging
     testWith(stream)
   }
 
-  def withMessageStreamFixtures[R](
-    testWith: TestWith[(Bucket,
-    MessageStream[ExampleObject],
+  def withMessageStreamFixtures[T,R](
+    testWith: TestWith[(ActorSystem, Bucket,
```
alexwlchan: Why are we adding the ActorSystem to the returned outputs of the fixture? I can’t see it being used anywhere.

alicefuzier: ah right, it was used at some point, but it's not anymore. Thanks, I'll remove that

alexwlchan: Cool. Happy for you to merge once that’s done, I don’t need another review.

PR #2242, opened 2018-06-15
[RFC] New reindexer architecture

alexwlchan:
Because the current reindexer keeps flaking out and falling over.

### DIFF
--- a/docs/rfcs/006-reindexer-architecture/README.md
+++ b/docs/rfcs/006-reindexer-architecture/README.md
@@ -1,0 +1,105 @@
+# RFC 006: Reindexer architecture
+
+**Last updated: 15 June 2018.**
+
+## Problem statement
+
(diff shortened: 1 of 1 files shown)


docs/rfcs/006-reindexer-architecture/README.md
```diff
@@ -0,0 +1,105 @@
...
+    This message is of the form:
+
+        {
+            "id": "miro/123",
+            "desiredVersion": 456
+        }
+
+4.  A new ECS service (the "reindex worker") receives this message.
+    It queries DynamoDB to get the current row with this ID, and if it still needs reindexing, it makes the PutItem call to update the row.
+
```
mklander: numbers 4 and maybe 3 could be a lambda.  I think ECS is probably best for all of this though.

PR #2247, opened 2018-06-15
Be consistent about importing Inject/Singleton

alexwlchan:
We had a mixture of imports from javax.inject and com.google.inject. I don’t know what the differences are – we should be consistent to reduce the risk of hard-to-spot bugs.


alexwlchan:
@mklander Forgive my ignorance, but what’s JSR330, and why would it be better? (Google is not immediately helpful in this regard.)

PR #2256, opened 2018-06-18
Matcher locktable terraform provision

alicefuzier:
🐎 🎻

### DIFF
--- a/catalogue_pipeline/terraform/dynamo_matcher_locktable.tf
+++ b/catalogue_pipeline/terraform/dynamo_matcher_locktable.tf
@@ -1,5 +1,5 @@
-resource "aws_dynamodb_table" "matcher_graph_table" {
-  name           = "works-graph"
+resource "aws_dynamodb_table" "matcher_lock_table" {
+  name           = "matcher-lock-table"
   read_capacity  = 1
   write_capacity = 1
   hash_key       = "id"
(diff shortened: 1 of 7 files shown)


catalogue_pipeline/terraform/dynamo_matcher_locktable.tf
```diff
@@ -0,0 +1,50 @@
...
+
+  enable_read_scaling     = true
+  read_target_utilization = 30
+  read_min_capacity       = 1
+  read_max_capacity       = 300
+
+  enable_write_scaling     = true
+  write_target_utilization = 30
+  write_min_capacity       = 1
+  write_max_capacity       = 300
```
alexwlchan: How did you decide these autoscaling limits? Naively, my guess would be that a lock table uses less capacity than a table with lots of data.

alicefuzier: I copied and pasted, to be honest. Anyway, the `write_target_utilization` should probably stay low 'cause we want the table to scale fast. `write_max_capacity` could be lower but I'm not sure it matters cause it will only scale as much as it needs to. Also, do failed conditional puts consume write capacity? The amount of write attempts will be more than on the graph table, it's just that we expect some of them to fail

alicefuzier: From [the aws documentation](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/CapacityUnitCalculations.html) 
>PutItem, UpdateItem, and DeleteItem allow conditional writes, where you specify an expression that must evaluate to true in order for the operation to succeed. If the expression evaluates to false, DynamoDB will still consume write capacity units from the table

PR #2258, opened 2018-06-19
Matcher concurrency and logging/metrics improvements

mklander:
Additional tests for the matcher to cover concurrency and failure handling


catalogue_pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/locking/DynamoRowLockDao.scala
```diff
@@ -37,46 +40,60 @@ class DynamoRowLockDao @Inject()(
...
+        case Right(rowLock) => rowLock.id
+        case Left(error) =>
+          info(s"Error $error when unlocking $contextId")
+          throw FailedLockException(s"Failed to unlock [$contextId] $error")
+      }.toSet
+
+      debug(s"Unlocking rows: $rowLockIds")
+      Scanamo.deleteAll(dynamoDBClient)(table.name)('id -> rowLockIds)
+      val deleteAllResults =
+        Scanamo.deleteAll(dynamoDBClient)(table.name)('id -> rowLockIds)
```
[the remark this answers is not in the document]
mklander: Thanks, think this got mixed up from when the result wasn't assigned

catalogue_pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/locking/DynamoRowLockDao.scala
```diff
@@ -37,46 +40,59 @@ class DynamoRowLockDao @Inject()(
...
+      deleteAllResults.foreach { result: BatchWriteItemResult =>
+        if (result.getUnprocessedItems.size() > 0) {
+          val error =
+            s"Batch delete failed to delete ${result.getUnprocessedItems}"
+          info(error)
+          throw FailedLockException(error)
+        }
       }
+    } catch {
```
alicefuzier: It's a bit weird to have a try catch block inside a Future. You can achieve the same thing with a recover on the future. It's a bit more scalary, I think

catalogue_pipeline/matcher/src/test/scala/uk/ac/wellcome/platform/matcher/fixtures/MatcherFixtures.scala
```diff
@@ -70,6 +75,13 @@ trait MatcherFixtures
     }
   }
 
+  implicit val instantLongFormat: AnyRef with DynamoFormat[Instant] =
```
alicefuzier: You should be able to get rid of this block by importing `uk.ac.wellcome.storage.dynamo._` I think

catalogue_pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/locking/DynamoRowLockDao.scala
kenoir: 👍

catalogue_pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/locking/DynamoRowLockDao.scala
```diff
@@ -37,46 +40,59 @@ class DynamoRowLockDao @Inject()(
...
+    debug(s"Got $result for $rowLock")
 
     result match {
       case Right(_) => rowLock
-      case Left(error) =>
-        throw new FailedLockException(s"Failed to lock $id", error)
+      case Left(error) => {
+        debug(s"Failed to lock $rowLock $error")
+        throw FailedLockException(s"Failed to lock $id", error)
+      }
```
alicefuzier: What happens if the call to `Scanamo.exec` throws an exception? I think, `lockRow` will return a failed future, but the wrapped exception will not be a `FailedLockException`.If we want metrics around failed locks and unlocks, we probably want it to be a `FailedLockException`. It could be useful to add a test around this scenario, perhaps? I think practically this is what happens with ThroughputEceededExceptions

catalogue_pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/locking/DynamoRowLockDao.scala
```diff
@@ -37,46 +40,59 @@ class DynamoRowLockDao @Inject()(
...
+
+      debug(s"Unlocking rows: $rowLockIds")
+      val deleteAllResults =
+        Scanamo.deleteAll(dynamoDBClient)(table.name)('id -> rowLockIds)
+      deleteAllResults.foreach { result: BatchWriteItemResult =>
+        if (result.getUnprocessedItems.size() > 0) {
+          val error =
+            s"Batch delete failed to delete ${result.getUnprocessedItems}"
+          info(error)
+          throw FailedLockException(error)
```
alicefuzier: Would it be useful to reintroduce `FailedUnlockException`? If we want to add metrics it might be useful to distinguish them. A failure locking is fine, but a failure unlocking is something that is a bit harder to recover from, so we might want to know how often it happens

catalogue_pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/messages/MatcherMessageReceiver.scala
```diff
@@ -40,7 +40,8 @@ class MatcherMessageReceiver @Inject()(
         subject = s"source: ${this.getClass.getSimpleName}.processMessage"
       )
     } yield ()).recover {
-      case e: VersionConflictException => warn(e.getMessage)
+      case e: VersionConflictException =>
+        debug(s"Not matching work due to version: ${e.getMessage}")
```
kenoir: 👍

catalogue_pipeline/terraform/queues.tf
```diff
@@ -57,8 +57,12 @@ module "matcher_queue" {
   account_id  = "${data.aws_caller_identity.current.account_id}"
   topic_names = ["${module.recorded_works_topic.name}"]
 
-  visibility_timeout_seconds = 60
-  max_receive_count          = 8
+  // The records in the locktable expire after 3 minutes
+  // The matcher is abkle to override locks that have expired
+  // Wait slightly longer to make sure locks are axpired
+  visibility_timeout_seconds = 210
```
kenoir: 👍 ⏲

alicefuzier: If only I'd spelled it correctly

PR #2263, opened 2018-06-21
Split the identifiers field to take advantage of the type system

alexwlchan:
Finishing off an idea from yesterday, and building on something mentioned this morning.

We have an "identifiers" field in the API that should contain all the identifiers on a Work/Item/Agent, including the one we stick in the "sourceIdentifier" field of the internal model – but there’s nothing that enforces we actually put the source identifier there. We have to remember to duplicate the information (to which: yuck).

This PR replaces the "identifiers" field with:

* an "otherIdentifiers" field, which is populated with any extra identifiers we want to add
* an "identifiers" property, which contains both the sourceIdentifier and the extra identifiers

This means we are *guaranteed* to include the source identifier in the "identifiers" field in the API, even if we forget to set it (which it turns out we do in a bunch of tests).

This is **not** the new mergeCandidates field – that will come separately.


alexwlchan:
This will need a reindex, so I’d like to hold merging this until I’m ready to do a deploy and a reindex.

PR #2265, opened 2018-06-21
Add mergeCandidates to our internal Work model

alexwlchan:
As discussed in standup, this adds a new field "mergeCandidates" to Work. This is distinct from "identifiers" – it’s a stronger statement that these Works should eventually exist in the pipeline, and the merger should look for them.

I also added a "reason" field, so the merger can know why we thought these two works might need merging.

### DIFF
--- a/sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
+++ b/sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
@@ -2,32 +2,35 @@ package uk.ac.wellcome.models.work.internal
 
 /** A representation of a work in our ontology */
 trait Work {
-  val title: Option[String]
   val sourceIdentifier: SourceIdentifier
-  val version: Int
   val identifiers: List[SourceIdentifier]
+
   val workType: Option[WorkType]
+  val title: Option[String]
   val description: Option[String]
   val physicalDescription: Option[String]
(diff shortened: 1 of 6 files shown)


sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
```diff
@@ -59,13 +64,14 @@ case class IdentifiedWork(
...
   subjects: List[Subject[Displayable[AbstractConcept]]] = Nil,
-  contributors: List[Contributor[Displayable[AbstractAgent]]] = Nil,
   genres: List[Genre[Displayable[AbstractConcept]]] = Nil,
+  contributors: List[Contributor[Displayable[AbstractAgent]]] = Nil,
   thumbnail: Option[Location] = None,
   items: List[IdentifiedItem] = Nil,
   production: List[ProductionEvent[Displayable[AbstractAgent]]] = Nil,
   language: Option[Language] = None,
   dimensions: Option[String] = None,
+  version: Int,
```
alicefuzier: Minor, but I tend to prefer having all members with no default value at the beginning of the list so that it's easier to spot what's the minimum amount of stuff that must be provided

alexwlchan: It’s getting quite

PR #2268, opened 2018-06-22
Don’t log health check requests from the ALB

### LINKED ISSUE #2238: Don't log ALB healthchecks in CloudWatch
I don’t think they’re a big cost, but they do spam the logs and slow down search:

```
10:14:09.578 [finagle/netty4-7] INFO c.t.f.h.filters.AccessLoggingFilter - 172.17.0.5 - - [14/Jun/2018:10:14:09 +0000] "GET /management/healthcheck HTTP/1.0" 200 16 0 "ELB-HealthChecker/2.0"
```

alexwlchan:
Resolves #2238. We no longer get the following logs in CloudWatch unless we have debug logging enabled:

```
10:14:09.578 [finagle/netty4-7] INFO c.t.f.h.filters.AccessLoggingFilter - 172.17.0.5 - - [14/Jun/2018:10:14:09 +0000] "GET /management/healthcheck HTTP/1.0" 200 16 0 "ELB-HealthChecker/2.0"
```

The motivation isn’t cost (although it doesn’t hurt!) – it’s to keep the logs clean. We know healthchecks are succeeding, or our applications would be killed by the ALB. By the rule of “don’t log the happy path”, let’s suppress these as well. This has a dramatic improvement on the readability of CloudWatch logs.

Tested by running a container locally and using curl.

### DIFF
--- a/catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/Server.scala
+++ b/catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/Server.scala
@@ -13,6 +13,7 @@ import uk.ac.wellcome.finatra.elasticsearch.{
   ElasticClientModule,
   ElasticConfigModule
 }
+import uk.ac.wellcome.finatra.modules.AccessLoggingFilterModule
 import uk.ac.wellcome.platform.api.controllers._
 import uk.ac.wellcome.platform.api.finatra.exceptions.{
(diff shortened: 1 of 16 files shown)


alicefuzier:
As mentioned to @alexwlchan ,I think there is a way that involves just changing the logback config and no actual code, so let's wait merging this

alexwlchan:
Replaced by #2269.

PR #2235, opened 2018-06-14
Palette search

harrisonpim:
🤖👉🎨

PR #2271, opened 2018-06-22
Update data_science stack to use new infra

kenoir:

Provide a space to make Wellcome Collection data science experiments public.

Also updates data science notebook infra.

### Who is this change for?

The public and @harrisonpim 

 - [X] Deployed new versions
- [X] Run `terraform apply`.

**Note:** This change depends on:
- https://github.com/wellcometrust/platform/pull/2235
- https://github.com/wellcometrust/terraform-modules/pull/76


data_science/terraform/notebooks/iam.tf
```diff
@@ -1,3 +1,38 @@
...
+    actions = [
+      "s3:ListAllMyBuckets",
+    ]
+
+    resources = [
+      "arn:aws:s3:::*",
+    ]
+  }
+}
+
```
mklander: Should we restrict these permissions a bit more?

data_science/terraform/variables.tf
```diff
@@ -1,15 +1,13 @@
...
-  # The default password is 'password'
-  # To generate a new password run the following Python code:
-  #
-  # from notebook.auth import passwd
-  #   passwd()
-  #
-  default = "sha1:5310f21e370d:a4d66e725c179218638c21c03d83933aa066db2d"
+variable "vpc_cidr_block" {
+  default = "18.0.0.0/16"
 }
```
mklander: I know it's possible, but perhaps we should be a bit careful about using 18.0.0.0/16?

data_science/terraform/notebooks/iam.tf
kenoir: Perhaps. My reading of the permissions here is that @harrisonpim can do anything to the data science bucket, and read everything else which seems ok to me?

data_science/terraform/variables.tf
kenoir: What issues might there be with using that block?

kenoir:
Merging, will address comments in future PR.

PR #2275, opened 2018-06-23
Tweak the max capacity on the SourceData VHS

alexwlchan:
We saturated it with the VHS reindex last night.

### DIFF
--- a/catalogue_pipeline/terraform/vhs_sourcedata.tf
+++ b/catalogue_pipeline/terraform/vhs_sourcedata.tf
@@ -3,4 +3,6 @@ module "vhs_sourcedata" {
 
   name              = "SourceData"
   table_name_prefix = ""
+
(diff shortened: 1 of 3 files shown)


alexwlchan:
This is ready to merge.

catalogue_pipeline/terraform/vhs_sourcedata.tf
```diff
@@ -3,4 +3,6 @@ module "vhs_sourcedata" {
 
   name              = "SourceData"
   table_name_prefix = ""
+
+  table_write_max_capacity = 750
```
mklander: 750 seems a good idea if we've maxed out. I think it's ~ 3 million writes per hour, so we should be able to reindex in about an hour?  SourceData currently has about 2 million entries -- does that seem about right?

PR #2288, opened 2018-06-25
Apply new formatting rules

alicefuzier:
As part of #2272, the scalafmt sbt plugin had to be updated cause the version we were using was not available for sbt 1.1.6. This has caused a slight change in formatting rules apparently wich where not applied. Checking the log for the auto formatting task on the upgrade sbt PR, I saw this ```*** Skipping format-scala as there are no affected files```. Apparently, scala format is only applied when there are changes to.scala files, not .sbt files. We might want to change that.

In the meantime, pushing the formatting changes as this is polluting all PRs


alicefuzier:
I'm going to just merge this to get clean PRs again. Please rebase against master in any PR affected

PR #2308, opened 2018-06-26
Modify work model to contain redirect information

alicefuzier:
In a way that allows the idminter to add an identifier for it that can be used by the API

### DIFF
--- a/sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
+++ b/sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
@@ -57,7 +57,7 @@ case class UnidentifiedWork(
   items: List[UnidentifiedItem] = Nil,
   visible: Boolean = true,
   ontologyType: String = "Work")
-    extends Work with IdentifiedBaseWork
+    extends Work
 
 case class IdentifiedWork(
   canonicalId: String,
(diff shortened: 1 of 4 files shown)


sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
```diff
@@ -48,9 +50,9 @@ case class UnidentifiedWork(
   language: Option[Language] = None,
   dimensions: Option[String] = None,
   items: List[UnidentifiedItem] = Nil,
-  version: Int,
   visible: Boolean = true,
-  ontologyType: String = "Work")
+  ontologyType: String = "Work",
+  redirect: Option[Identifiable[None.type]] = None)
```
alicefuzier: Obviously the None.type thing is not right. Unfortunately the IdentityState trait is designed to always wrap a value of type T and we need to rejig it to make it just wrap a sourceIdentifier field. I'm going to try and do that next but I want to see if things break with this in

sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
```diff
@@ -27,13 +32,15 @@ trait Work extends MultipleSourceIdentifiers {
   val visible: Boolean
 
   val ontologyType: String
+  val redirect: Option[Redirect]
```
kenoir: Instead of putting redirect on work could we create a `RedirectedWork` that shared a sealed trait with `Work`?

alexwlchan: That would allow us to enforce that visible == false if redirect is not-None.

alicefuzier: Yep, I think we could extend that idea for non visible works as well and avoid the horrible error when a work is visible but has no title

PR #2312, opened 2018-06-27
Adds catalogue_vpc to shared_infra

kenoir:

Reduce the number of VPCs we have by consolidating all adapters and APIs into the same VPC using public/private subnets to ensure private services are not user facing.

### Who is this change for?

Folk who want simpler more sensible infrastructure.


alexwlchan:
Is anybody going to use this new VPC?

kenoir:
@alexwlchan everybody! It's the new hot VPC to be placed in!

alexwlchan:
So are the changes to use it coming in a separate patch?

kenoir:
Yup, first thing is to move `sierra_adapter` to it and then carry on moving things to the new infra while using this VPC.

PR #2318, opened 2018-06-28
Improvements to the “trigger reindex” script

alexwlchan:
* Don’t set SourceData capacity manually, this was an ugly hack
* Use the current timestamp as the reindex version
* Don't print a massive traceback if you only run a partial reindex
* Drop a message in Slack when the reindex starts, so we can track them:

    
screen shot 2018-06-28 at 14 34 42

### DIFF
--- a/reindexer/trigger_reindex.py
+++ b/reindexer/trigger_reindex.py
@@ -3,11 +3,13 @@
 """
 Create/update reindex shards in the reindex shard tracker table.
 
-Usage: trigger_reindex.py --prefix=<PREFIX> [--count=<COUNT>]
+Usage: trigger_reindex.py --prefix=<PREFIX> --reason=<REASON> [--count=<COUNT>]
        trigger_reindex.py -h | --help
 
 Options:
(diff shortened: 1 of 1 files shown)


reindexer/trigger_reindex.py
```diff
@@ -149,30 +184,10 @@ def create_shards(client, prefix, desired_version, count, table_name):
...
-                'Update': {
-                    'IndexName': 'reindexTracker',
-                    'ProvisionedThroughput': {
-                        'WriteCapacityUnits': 50,
-                        'ReadCapacityUnits': 5
-                    }
-                }
-            }
-        ]
-    )
```
mklander: I wonder if we will need this, or something like it again?

alexwlchan: Maybe. But I don’t think it’d be hard to add back.

I removed it because it hits errors on the Sierra reindexes – by the time we reach this line, table capacity has already scaled past 50 write units, and it errors because it thinks you’re trying to decrease capacity.

Taking it out and taking a small hit on Miro reindexes seemed easier than adding error handling logic for now.

PR #2321, opened 2018-06-29
Delete the reindex job creator, shard tracker table, push it inside the trigger_reindex script

alexwlchan:
Continues #2302.

This modifies the trigger reindex script to:

* Query the shard generator for the number of shards it creates, not a table that we have no mechanism for updating
* Send reindex jobs directly to SNS, not via DynamoDB and the reindex_job_creator

Knock-on effect is that we can delete the reindex_job_creator (another slice off the builds!) and the ReindexShardTracker table.


reindexer/reindex_shard_generator/src/test_shard_manager.py
```diff
@@ -1,23 +0,0 @@
...
-
-from shard_manager import create_reindex_shard
-
-
-@pytest.mark.parametrize(
-    'source_name, id, source_size, shard_size, expected_reindex_shard', [
-    ('sierra', 'b0000001', 5e6, 1500, 'sierra/2441'),
-    ('sierra', 'i0000001', 5e6, 2000, 'sierra/2349'),
-    ('miro', 'V0000123', 2.5e5, 1000, 'miro/51'),
-])
```
mklander: the miro example is dropped from the new test, is it worth keeping?

reindexer/reindex_shard_generator/src/reindex_shard_config.py
```diff
@@ -0,0 +1,37 @@
...
+
+def get_number_of_shards(source_name):
+    """How many shards should we create for a given source?"""
+    return math.ceil(SOURCE_SIZES[source_name] / SHARD_SIZE)
+
+
+def create_reindex_shard(source_name, source_id):
+    """What reindex shard should a given source name/ID pair be placed in?"""
+    ascii_id = int(''.join(str(ord(c)) for c in source_id))
+    number_of_shards = get_number_of_shards(source_name)
```
mklander: I can't quite follow the ascii_id here, I see what it's doing but not quite how that maps to the shard boundaries.

reindexer/reindex_shard_generator/src/reindex_shard_generator.py
```diff
@@ -68,6 +60,11 @@ def main(event, _ctxt=None, dynamodb_client=None):
                 }
             )
         except botocore.exceptions.ClientError as e:
+
```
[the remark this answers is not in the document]
alexwlchan: Unfortunately boto3 doesn’t expose specific exceptions like that – you get a generic ClientError, but the specific exception type isn’t created until it’s thrown, so you can’t catch it.

(I think it’s possible if you really want to, but this is less fiddly.)

reindexer/reindex_shard_generator/src/reindex_shard_config.py
alexwlchan: I’ll add a comment.

PR #2323, opened 2018-06-29
Modify the reindex_worker to send individual SNS messages for each record

alexwlchan:
Part of #2202.

This PR changes the functionality and internals of the reindex_worker quite drastically:

* Previously the `ReindexService` read records from DynamoDB, then tried to update them. It’s been deleted, and replaced by two new classes.
* The `RecordReader` class queries DynamoDB to work out which records need updating, and returns their IDs.
* The `NotificationSender` class gets the list of IDs, wraps them in a `ReindexRequest`, and sends them to SNS for somebody else to process.

Models have been tweaked to clarify what we're getting from SQS (`ReindexJob`), reading from DynamoDB (`ReindexableRecord`) and sending for somebody else to deal with (`ReindexRequest`).

Plus the tests have been broken up to reflect the new classes. I’ve added a few new ones, but I haven’t deleted any of the old tests, just moved them around.


alexwlchan:
This is passing tests (locally), ready for review, and ready to merge – I’ll do the Terraform later.

reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex_worker/services/RecordReader.scala
```diff
@@ -0,0 +1,70 @@
...
+        Scanamo.exec(dynamoDbClient)(
+          index.query(
+            'reindexShard -> reindexJob.shardId and
+              KeyIs('reindexVersion, LT, reindexJob.desiredVersion)
+          )
+        )
+      }
+
+      outdatedRecordIds: List[String] = results.map(extractRecordID)
+    } yield outdatedRecordIds
```
mklander: tiny thing, `outdatedRecordIds` doesn't seem quite right to me, perhaps `reindexRecordIds`?

kenoir:
Could we not vary the number of records that get published downstream? I've got a nagging feeling that publishing a request for every record is going to cause other issues - it seems like unnecessarily high overhead to update a single row.

alexwlchan:
> I've got a nagging feeling that publishing a request for every record is going to cause other issues - it seems like unnecessarily high overhead to update a single row.

We publish an SNS notification per single record at every other point in the pipeline – what makes you think this would be different?

kenoir:
@alexwlchan it's not, i think we're wrong elsewhere. Maybe shouldn't be blocking for merge here though.

Issue #2325, opened 2018-06-29
Fix prevPage/nextPage for out-of-bounds API responses


alexwlchan:
For example:

```console
$ curl 'https://api.wellcomecollection.org/catalogue/v1/works?query=ROYAL%20FREE%20HOSPITAL&page=100&pageSize=100' 2>/dev/null | python -m json.tool
{
    "@context": "https://api.wellcomecollection.org/catalogue/v1/context.json",
    "type": "ResultList",
    "pageSize": 100,
    "totalPages": 78,
    "totalResults": 7722,
    "results": [],
    "prevPage": "https://api.wellcomecollection.org/catalogue/v1/works?page=99&pageSize=100&query=ROYAL%20FREE%20HOSPITAL",
    "nextPage": "https://api.wellcomecollection.org/catalogue/v1/works?page=101&pageSize=100&query=ROYAL%20FREE%20HOSPITAL"
}
```

We should omit nextPage, and (maybe?) point prevPage to the last populated page.

There's a separate question as to how crawlers are discovering this page, given we don’t advertise it anywhere. Maybe based on our Swagger definitions?

PR #2326, opened 2018-06-29
Don't show a "nextPage" parameter in API responses if we're beyond the last page

alexwlchan:
First part of #2325 – this makes our API responses a bit more sensible.

Additionally add a bunch of testing around this code, and simplifies a bit of the existing logic.

I’m not sure what to do with prevPage, and @jtweed is away from their desk, so let’s do that in a separate patch.

### DIFF
--- a/catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/responses/ResultListResponseTest.scala
+++ b/catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/responses/ResultListResponseTest.scala
@@ -30,6 +30,16 @@ class ResultListResponseTest extends FunSpec with Matchers {
     request = Request(method = Method.Get, uri = requestUri)
   )
 
+  it("inclues a nextPage and prevPage parameter where appropriate") {
+    val resp = getResponse(
+      displayResultList = displayResultList.copy(totalPages = 5),
+      multipleResultsRequest = multipleResultsRequest.copy(page = 3)
+    )
+
+    resp.prevPage shouldBe Some(s"$requestBaseUri$requestUri?page=2")
+    resp.nextPage shouldBe Some(s"$requestBaseUri$requestUri?page=4")
+  }
+
   describe("nextPage") {
     it("omits the parameter if this is the only page") {
       val resp = getResponse(
--- a/catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/responses/ResultListResponseTest.scala
+++ b/catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/responses/ResultListResponseTest.scala
@@ -68,6 +68,34 @@ class ResultListResponseTest extends FunSpec with Matchers {
     }
   }
 
+  describe("prevPage") {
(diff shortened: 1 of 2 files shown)


catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/responses/ResultListResponseTest.scala
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    it("omits the parameter if the request is beyond the last page") {
+      val resp = getResponse(
+        displayResultList = displayResultList.copy(totalPages = 5),
+        multipleResultsRequest = multipleResultsRequest.copy(page = 10)
+      )
+
+      resp.nextPage shouldBe None
+    }
```
mklander: 👍

mklander: While we are dealing with odd cases is it worth thinking about requests that set pages before the first (0, -1)?

catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/responses/ResultListResponseTest.scala
```diff
@@ -0,0 +1,110 @@
...
+
+    it("omits the parameter if this is the first page") {
+      val resp = getResponse(
+        displayResultList = displayResultList.copy(totalPages = 5),
+        multipleResultsRequest = multipleResultsRequest.copy(page = 1)
+      )
+
+      resp.prevPage shouldBe None
+    }
+
```
mklander: Maybe assert the next page in these cases as well? It's captured elsewhere in the tests, but might be good to cover.

catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/responses/ResultListResponseTest.scala
alexwlchan: Yep, I’ll add that.

catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/responses/ResultListResponseTest.scala
alexwlchan: I’d rather keep them as-is – each test can only fail one way. The next/prev page logic is independent, so independent tests.

I’ll add a new test which tests than both at once, and keep this describe block tightly scoped.

mklander: Yes, completely agree about tests having only a single reason to fail.  Single assertion is a good rule, but sometimes there's a balance with test duplication.

kenoir:
Why don't we just 404?

kenoir:
I reckon @jtweed will have an opinion on the correct behaviour here.

alexwlchan:
I don’t think 404 is right for an empty collection; the collection was found, just not present.

(Let’s discuss this on Monday.)

PR #2329, opened 2018-06-29
Dynamo put heartbeat

mklander:
Zero throughput means dynamoDb does not scale down its provisioned capacity.
This PR creates a periodic lambda (every 10 minutes) that writes to the specified dynamo tables (matcher-lock-table here) to create some traffic so that dynamoDb scales back down to meet demand.


shared_infra/dynamo_write_heartbeat/iam_role_policy.tf
```diff
@@ -0,0 +1,11 @@
+data "aws_iam_policy_document" "allow_dynamodb_delete_item" {
+  statement {
+    actions = [
+      "dynamodb:DeleteItem",
+    ]
+
+    resources = [
+      "${join(",",formatlist("arn:aws:dynamodb:*:*:table/%s", split(",", replace(var.dynamo_table_names, ", ", ","))))}",
```
alexwlchan: I’ll need a comment to explain this please, because ugh terraform syntax

shared_infra/dynamo_write_heartbeat/src/dynamo_write_heartbeat.py
```diff
@@ -0,0 +1,26 @@
...
+import boto3
+import os
+
+from wellcome_aws_utils.lambda_utils import log_on_error
+
+
+@log_on_error
+def main(event, context, dynamodb_client=None):
+    try:
+        table_names = [t.strip(' ') for t in os.environ['TABLE_NAMES'].split(',')]
```
alexwlchan: minor: calling `t.strip()` without any arguments should already strip whitespace for you

shared_infra/dynamo_write_heartbeat/src/dynamo_write_heartbeat.py
```diff
@@ -0,0 +1,26 @@
...
+    except KeyError:
+        raise RuntimeError('TABLE_NAMES not found')
+
+    dynamodb_client = dynamodb_client or boto3.client('dynamodb')
+    for table_name in table_names:
+        response = dynamodb_client.delete_item(TableName=table_name.strip(),
+                                               Key={'id': {'S': 'dummy-id-not-there'}})
+        print("Heartbeat, delete_item on dynamoDb table:{}, returned-status:{}".format(table_name, response['ResponseMetadata']['HTTPStatusCode']))
```
alexwlchan: minor: it would be nice to make this a DEBUG log (I’ll have to check our other Lambdas for an example)

mklander: Thanks, yes if you know how.  The log on error seems to use print.

PR #2334, opened 2018-07-02
Check that the id minter can mint identifiers for a redirected work

alicefuzier:
⌨️ 💫

### DIFF
--- a/catalogue_pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/IdMinterFeatureTest.scala
+++ b/catalogue_pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/IdMinterFeatureTest.scala
@@ -111,7 +111,7 @@ class IdMinterFeatureTest
                 obj = work,
                 location = ObjectLocation(
                   namespace = bucket.name,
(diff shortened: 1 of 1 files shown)


catalogue_pipeline/id_minter/src/test/scala/uk/ac/wellcome/platform/idminter/IdMinterFeatureTest.scala
```diff
@@ -131,6 +131,56 @@ class IdMinterFeatureTest
...
+                sourceIdentifier = identifier,
+                version = 1,
+                redirect = IdentifiableRedirect(sourceIdentifier = SourceIdentifier(IdentifierType("sierra-system-number"), "Work", "b1234567"))
+              )
+
+              val messageBody = put[UnidentifiedRedirectedWork](
+                obj = work,
+                location = ObjectLocation(
+                  namespace = bucket.name,
+                  key = s"redirected-work.json"
```
alexwlchan: nit: why is this an s-string?

alicefuzier: Very good question

Issue #2333, opened 2018-07-02
Post-mortem issues from works page outages on Friday


alexwlchan:
1. [ ] Add a flag to https://github.com/wellcometrust/wellcomecollection.org that lets us use the staging API with /works pages /cc @jamesgorrie 
2. [x] Fix the API pins script so it doesn't unpin the current prod API (#2414)
3. [x] Improve Circe's behaviour when encountering an outdated model in the API – remove defaults, maybe more
4. [x] Check a Miro result remains unchained in the API switching script (#2414)

This is the script in question: https://github.com/wellcometrust/platform/blob/master/catalogue_api/get_api_versions.py (it's somewhat poorly named).

alexwlchan:
(5. Maybe: have our own monitor for /works pages on wellcomecollection.org)

PR #2335, opened 2018-07-02
[WIP] Don't use defaults in our Work model

alexwlchan:
Part of #2333.

When you create an instance of `Work`, you need to specify _all_ the fields – previously we had a variety of defaults. This should make Circe stricter, and it'll be easier to spot places where we have mismatched models in play.

This patch includes a major reworking of the `WorksUtil` trait so the tests can be kept sensible (and in many cases, I’ve trimmed the setting of unnecessary attributes in tests). Plus a new test that trying to decode JSON from an old model fails (the decoding succeeded before I’d made my changes).


sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
```diff
@@ -40,48 +40,48 @@ trait Work extends BaseWork with MultipleSourceIdentifiers {
...
+  lettering: Option[String],
+  createdDate: Option[Period],
+  subjects: List[Subject[Displayable[AbstractConcept]]],
+  genres: List[Genre[Displayable[AbstractConcept]]],
+  contributors: List[Contributor[Displayable[AbstractAgent]]],
+  thumbnail: Option[Location],
+  production: List[ProductionEvent[Displayable[AbstractAgent]]],
+  language: Option[Language],
+  dimensions: Option[String],
+  items: List[Identified[Item]],
```
alexwlchan: The change here is to reorder the fields on UnidentifiedWork and IdentifiedWork so they match, which should make for easier comparisons.

alexwlchan:
Replaced by #2403.

PR #2336, opened 2018-07-02
Loris new infra

kenoir:

Use the new infra modules for Loris

### Who is this change for?

Devs who want good infrastructure.


loris/terraform/loris.tf
```diff
@@ -0,0 +1,43 @@
+module "loris_service" {
+  source = "service"
+
+  namespace = "loris-delta"
+
+  certificate_domain = "api.wellcomecollection.org"
```
alexwlchan: ???

loris/terraform/loris.tf
```diff
@@ -0,0 +1,43 @@
...
+  sidecar_cpu    = "128"
+  sidecar_memory = "128"
+
+  sidecar_env_vars = {}
+
+  app_container_image = "${module.ecr_loris.repository_url}:${var.release_ids["loris"]}"
+  app_container_port  = "8888"
+
+  app_cpu    = "3076"
+  app_memory = "7222"
```
alexwlchan: What’s the difference between this and the CPU/memory specified above?

loris/terraform/loris.tf
kenoir: This is a multi domain certificate (also covers iiif.wc & iiif-origin.wc) not actually a change from existing tf.

PR #2341, opened 2018-07-02
Hide the internal workings of WorksUtil

alexwlchan:
In #2335, `WorksUtil` takes on (even more) responsibility for setting defaults on instances of `Work` created for tests – but it leaks the defaults in a public namespace. This means a test can “know” a particular default will be set there, and use it to assert on.

It’s better for tests to be explicit about where they’re reading values from.

This patch either hides or entirely removes defaults from `WorksUtil`, in favour of looking up the correct value from the generated `Work` instance.


alexwlchan:
Replaced by #2352.

PR #2345, opened 2018-07-03
Log exceptions when using the runStream function

alicefuzier:
Because the ingestor wasn't logging any exception. This will blat the change that we made recently in #2320 , but we could probably change the apps to fill in relevant information in the exceptions instead of relying in SQSStream


sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/sqs/SQSStream.scala
```diff
@@ -99,23 +86,32 @@ class SQSStream[T] @Inject()(actorSystem: ActorSystem,
       .run()
   }
 
-  private def readAndProcess(streamName: String,
-                             message: T,
-                             process: T => Future[Unit]) = {
-    val processMessageFuture = process(message)
+  // Defines a "supervision strategy" -- this tells Akka how to react
```
kenoir: 👍

PR #2348, opened 2018-07-03
Fix id minter handling of different types of work

alicefuzier:
🦄 👂

### DIFF
--- a/catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraGenresTest.scala
+++ b/catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraGenresTest.scala
@@ -233,8 +233,9 @@ class SierraGenresTest extends FunSpec with Matchers {
       .getGenres(bibData)
       .map { _.concepts.head }
       .map {
-        case Identifiable(_: Concept, sourceIdentifier, _, _) => sourceIdentifier
-        case other                                         => assert(false, other)
+        case Identifiable(_: Concept, sourceIdentifier, _, _) =>
+          sourceIdentifier
+        case other => assert(false, other)
       }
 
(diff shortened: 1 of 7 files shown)


sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
```diff
@@ -42,6 +42,7 @@ case class UnidentifiedWork(
   sourceIdentifier: SourceIdentifier,
   version: Int,
   title: String,
+  identifiedType: String = classOf[IdentifiedWork].getSimpleName,
```
alexwlchan: Does this need to be passed around as a case class parameter? Could we put it as a static value on the body of the class instead?

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraGenresTest.scala
```diff
@@ -233,8 +233,9 @@ class SierraGenresTest extends FunSpec with Matchers {
       .getGenres(bibData)
       .map { _.concepts.head }
       .map {
-        case Identifiable(_: Concept, sourceIdentifier, _) => sourceIdentifier
-        case other                                         => assert(false, other)
+        case Identifiable(_: Concept, sourceIdentifier, _, _) =>
+          sourceIdentifier
+        case other => assert(false, other)
```
alexwlchan: Why is this suddenly being flagged by scalafmt? Le sigh…

sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/Work.scala
alexwlchan: Ah, because we need it on the JSON for the ID minter? Might be worth a comment.

PR #2349, opened 2018-07-03
Consolidate the helpers in WorksUtil

alexwlchan:
Another piece spun out from #2335.

We had a variety of methods `workWith` (with different signatures), `identifiedWorkWith`, `createWork`, and so on for creating instances of `Work` in tests.

This patch replaces all of them with a single method `createIdentifiedWorkWith`, and then adds a few helpers around it. Additionally, we default a bunch of fields we didn't before, like canonicalId and sourceIdentifier, and use genuinely unique strings on each call.


catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchServiceTest.scala
```diff
@@ -191,14 +168,14 @@ class ElasticsearchServiceTest
           indexName = indexName,
           limit = 10,
           from = 0,
-          expectedWorks = works.toList
+          expectedWorks = visibleWorks.toList
```
alexwlchan: I’ll be honest; I have no idea how this test was passing before.

PR #2350, opened 2018-07-03
Don't run API tests for storage/messaging changes

alexwlchan:
The API tests are usually 13–14 minutes long, and given the frequency with which we fiddle with either VHS or the SQS streams, this could save us quite a bit of build time.


kenoir:
Wait a minute ... I thought we were using the external storage library!

Looks like I lost the plot a bit here. Why aren't we using https://github.com/wellcometrust/scala-storage?

alicefuzier:
I believe, because the release process is broken, so the change was rolled back until we fix the releases in scala-storage

PR #2351, opened 2018-07-03
Never create instances of Work directly in tests

alexwlchan:
Another pieces sound out from #2335, and follows #2349.

If we’re going to remove all the defaults from the Work/UnidentifiedWork/… constructors, we need to use them in as few places as possible – hence the helpers on `WorksUtil`.

This patch gets all (?) the places where we construct an instance of a Work subclass in the tests, and replaces them with helpers in WorksUtil.

### DIFF
--- a/catalogue_pipeline/merger/src/test/scala/uk/ac/wellcome/platform/merger/MergerTestUtils.scala
+++ b/catalogue_pipeline/merger/src/test/scala/uk/ac/wellcome/platform/merger/MergerTestUtils.scala
@@ -10,6 +10,7 @@ import uk.ac.wellcome.models.matcher.{
 }
 import uk.ac.wellcome.models.recorder.internal.RecorderWorkEntry
 import uk.ac.wellcome.models.work.internal._
+import uk.ac.wellcome.models.work.test.util.WorksUtil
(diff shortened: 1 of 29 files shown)


catalogue_pipeline/merger/src/test/scala/uk/ac/wellcome/platform/merger/MergerTestUtils.scala
```diff
@@ -51,25 +52,8 @@ trait MergerTestUtils { this: SQS with SNS with Messaging =>
...
-  def recorderInvisibleWorkEntryWith(identifierType: String,
-                                     sourceId: String,
-                                     version: Int) =
-    RecorderWorkEntry(
-      UnidentifiedInvisibleWork(
-        sourceIdentifier =
-          SourceIdentifier(IdentifierType(identifierType), "Work", sourceId),
-        version = version))
+  def recorderWorkEntryWith(version: Int) =
+    RecorderWorkEntry(createUnidentifiedWorkWith(version = version))
```
mklander: createRecorderWorkEntryWith?

alexwlchan:
Finally green!

PR #2352, opened 2018-07-04
Remove a bunch of now-unused values from WorksUtil

alexwlchan:
Follow-up to #2351. Replaces #2341.


alexwlchan:
Since this was reviewed in #2341 and only CodeBuild is upset, I’m merging by fiat to unlock testing on two other PRs.

PR #2356, opened 2018-07-05
Improve transformer logging and error handling

alicefuzier:
🖖 🚲

### DIFF
--- a/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/miro/MiroLicenses.scala
+++ b/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/miro/MiroLicenses.scala
@@ -18,12 +18,12 @@ trait MiroLicenses {
     *  TODO: Update these mappings based on the final version of Christy's
     *        document.
     */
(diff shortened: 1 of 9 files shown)


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/miro/MiroLicenses.scala
```diff
@@ -43,10 +45,10 @@ trait MiroLicenses {
           // so they don't end up on the DLQ.
           case "Do not use" =>
             throw new ShouldNotTransformException(
-              "Usage restriction is 'Do not use'")
+              s"Image $miroId has usage restriction 'Do not use'")
           case "Image withdrawn, see notes" =>
             throw new ShouldNotTransformException(
-              "Usage restriction is 'Image withdrawn'")
+              s"Image $miroId has usage restriction 'Image withdrawn'")
         }
```
mklander: should we handle and test for the default case explicitly ('_'), rather than allow a MatchError?  I guess that goes to the DLQ so may be what is required.

alexwlchan:
Only Shippable is whinging, and I think I’ve disabled that integration now.

PR #2358, opened 2018-07-05
[WIP] DynamoDB heartbeat to support more configurable requests

mklander:
Allow configurable dynamoDB requests so that secondary indices can be included.


kenoir:
@mklander what needs to happen to merge this?

PR #2403, opened 2018-07-08
No more defaults in our Work models

alexwlchan:
Replaces #2335. Builds on #2352. Part of #2333.

Plus I rearranged the fields on UnidentifiedWork to have a consistent order with the base trait and IdentifiedWork.


sbt_common/internal_model/src/test/scala/uk/ac/wellcome/models/work/internal/WorkTest.scala
```diff
@@ -0,0 +1,130 @@
...
+        |  ],
+        |  "version": 27,
+        |  "visible": true,
+        |  "ontologyType": "Work"
+        |}
+      """.stripMargin
+
+    intercept[GracefulFailureException] {
+      fromJson[IdentifiedWork](jsonString).get
+    }
```
mklander: Is it possible to be clearer what the failure is?

alexwlchan: I need to rebase, I’ll include a check on the error message

PR #2415, opened 2018-07-10
Return a 302 Found response for redirected works

### LINKED ISSUE #1010: Redirect merged works in the Catalogue API

alexwlchan:
If we index a redirected work in Elasticsearch, retrieving it from the API will return a 302 Found, not a 410 Gone. Note that this preserves query parameters from the original request, because that feels like a sensible thing to do.

Closes #1010.

### DIFF
--- a/catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
+++ b/catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
@@ -6,7 +6,7 @@ import io.swagger.models.parameters.QueryParameter
 import io.swagger.models.properties.StringProperty
 import io.swagger.models.{Operation, Swagger}
 import uk.ac.wellcome.display.models.{ApiVersions, DisplayWork, WorksIncludes}
-import uk.ac.wellcome.models.work.internal.{IdentifiedBaseWork, IdentifiedWork}
+import uk.ac.wellcome.models.work.internal._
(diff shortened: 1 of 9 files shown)


catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
```diff
@@ -134,6 +139,24 @@ abstract class WorksController(apiConfig: ApiConfig,
...
+  private def respondWithRedirect(originalUri: String,
+                                  work: IdentifiedRedirectedWork,
+                                  contextUri: String) =
+    response.found
+      .body("")
+      .location(
+        uri = originalUri.replaceAll(
+          s"/${work.canonicalId}",
+          s"/${work.redirect.canonicalId}"
+        )
```
mklander: I wonder if it's worth anchoring this to ensure a single replacement of the last id?

alexwlchan: Seems sensible.

alexwlchan: Ah, but what do you anchor on? We don‘t want to match on the end of the string – then it doesn’t work for query parameters. I can’t think of a better way to do this that isn’t substantially more complicated. 😕

alexwlchan:
I’m going to merge this as-is, and if we can think of a better way to do the regex, we can do it as a separate patch.

PR #2418, opened 2018-07-11
[WIP] Consolidate helpers for constructing Sierra[Bib|Item]Record in tests

alexwlchan:
Several of the fields on these we don’t really care about exact values – e.g. id, modifiedDate.

By creating some new `createRecordWith` helpers, we can focus on the values we actually care about, and cut down the noise. We already had lots of helpers – but repeated across the codebase, not consolidated in one place! So we can save ourselves some trouble there.

Additionally, the classes had several apply methods that only existed for tests. We can remove those now, and simplify those models. Hooray!


sbt_common/internal_model/src/test/scala/uk/ac/wellcome/models/transformable/sierra/test/utils/SierraUtil.scala
```diff
@@ -0,0 +1,80 @@
...
+  val newerDate: Instant = Instant.parse("2001-01-01T01:01:01Z")
+
+  /** Returns a random digit as a string.  This is copied from the
+    * definition of Random.alphanumeric.
+    *
+    * @return
+    */
+  private def randomNumeric: Stream[Char] = {
+    def nextAlphaNum: Char = {
+      val chars = "0123456789"
```
mklander: Is this nextAlphaNum or nextDigit?

perhaps `val chars = ('0' to '9')

alexwlchan: Copy-pasta error from `Random.alphaNumeric`, will fix

sierra_adapter/sierra_reader/src/main/scala/uk/ac/wellcome/platform/sierra_reader/modules/WindowManager.scala
```diff
@@ -61,11 +61,7 @@ class WindowManager @Inject()(
         val triedStatus = triedMaybeLastId
           .map {
             case Some(id) =>
-              // The Sierra IDs we store in S3 are prefixed with "b" or "i".
-              // Remove the first character
-              val unprefixedId = id.substring(1)
-
-              val newId = (unprefixedId.toInt + 1).toString
```
alexwlchan: Flagging this for extra review – this is the correct behaviour, but does this mean we’re missing Sierra data? In particular, I can’t see how this logic would ever cause us to fetch more than the first page of seven-digit Sierra IDs

(Why didn’t we put an `assert id.substring(0) in ('b', 'i')`, gah)

kenoir: You might have to explain this in person.

alexwlchan: When we first wrote the Sierra reader, the IDs we stored were of the form “b1234567”.

The Sierra API wants to see the “1234567” part, no “b” – so we discard it here.

Now IDs are stored in the form “1234567”, so this transformation would strip it to “234567”, and we’d request that from the Sierra API – which is wrong.

So I think we might be dropping all but the first page of results. 😕

alexwlchan:
Replaced by #2442 and #2443.

PR #2421, opened 2018-07-12
Reindex processor terraform

mklander:
Deploy the reindex processor changes

### DIFF
--- a/reindexer/terraform/service_reindex_request_creator.tf
+++ b/reindexer/terraform/service_reindex_request_creator.tf
@@ -17,6 +17,7 @@ module "reindex_request_creator" {
     reindex_requests_topic_arn = "${module.reindex_requests_topic.arn}"
     metrics_namespace          = "reindex_request_creator"
   }
+
   env_vars_length = 4
 
   cluster_name               = "${local.catalogue_pipeline_cluster_name}"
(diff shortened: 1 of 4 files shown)


reindexer/terraform/service_reindex_request_creator.tf
```diff
@@ -1,5 +1,5 @@
 module "reindex_request_creator" {
-  source = "git::https://github.com/wellcometrust/terraform-modules.git//sqs_autoscaling_service?ref=v10.3.0"
+  source = "git::https://github.com/wellcometrust/terraform-modules.git//sqs_autoscaling_service?ref=v10.2.2"
```
kenoir: Why the step to 10.2.2?

alicefuzier: I can only guess, but I assume it's because 10.2.2 is the last version of the old modules supporting `env_vars_length` parameter. Probably the environment variables interpolation wasn't quite working

mklander: Yes, it's because the  environment variables interpolation wasn't working.  I think the latest version has this removed, so I'll switch it back.

mklander: After discussing this just now, I'm going to merge as is because we will revisit this when we move to the new infrastructure.

PR #2429, opened 2018-07-17
Set up skeleton for archive service project

kenoir:

Adds the skeleton for the archive service project.

### Who is this change for?

Archivists, developers, etc.


alexwlchan:
We should teach travistooling to ignore changes here, but it doesn't block merge.

PR #2434, opened 2018-07-18
Simple Archiver functionality

kenoir:

Demonstrate:

- BagIt verification from Zip file
- Verifying uploads to S3

### Who is this change for?

📖


kenoir:
For https://github.com/wellcometrust/platform/issues/2368

archive/archiver/src/main/scala/uk/ac/wellcome/platform/archiver/messaging/MessageStream.scala
```diff
@@ -49,13 +49,20 @@ class MessageStream[T, R] @Inject()(actorSystem: ActorSystem,
       ActorMaterializerSettings(system)
         .withSupervisionStrategy(decider(metricName, Supervision.Resume)))
 
-    val typeConversion = Flow[Message].map(m => fromJson[T](m.getBody).get)
+    val typeConversion = Flow[Message].map( m => {
```
kenoir: Can we not use the `log` function from akka streams here?

mklander: The Akka log looks a bit all or nothing to me, I wanted to log the incoming msg at `info` and leave everything else at `debug`.

kenoir:

PR #2445, opened 2018-07-23
Loris up to date

kenoir:

Move Loris to the latest version of the terraform modules for services to avoid permissions and env var issues present in earlier version.

### Who is this change for?

🍭


kenoir:
@gestchild ^ this was the issue causing 502s over the weekend. The cache cleaner service wasn't correctly configured so did not work properly ... cluster host disks filled up (eventually) and some services on those hosts started returning 502s.

PR #2444, opened 2018-07-23
Tidy up creation of SierraTransformable

alexwlchan:
Follows #2444, spun out of #2432.

PR #2446, opened 2018-07-23
Start rolling out the SierraRecordNumber type

alexwlchan:
Follows #2444, spun out of #2432.


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/source/SierraBibData.scala
```diff
@@ -15,4 +13,6 @@ case class SierraBibData(
   materialType: Option[SierraMaterialType] = None,
   fixedFields: Map[String, FixedField] = Map(),
   varFields: List[VarField] = List()
-)
+) {
+  val sierraId = SierraRecordNumber(id)
```
alexwlchan: Custom apply method?

sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/transformable/sierra/SierraRecordNumber.scala
```diff
@@ -1,41 +1,21 @@
...
-/** Utilities for working with Sierra "record numbers".
-  *
-  * These record numbers are the seven- or eight-digit IDs that
-  * are used to identify resources in Sierra.
+/** Represents the seven-digit IDs that are used to identify resources
+  * in Sierra.
   */
-object SierraRecordNumbers {
+case class SierraRecordNumber(s: String) {
+  require(
```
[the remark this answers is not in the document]
alexwlchan: The intention (and I should document this) is that you should never be accessing the `s` field directly; only the helper methods `withoutCheckDigit` and `withCheckDigit`.

Maybe I just rename this field to `withoutCheckDigit`?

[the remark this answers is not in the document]
alexwlchan: Thinking on this overnight, I can kill two birds with one stone – I’ll make this a class, not a case class (so this field can be private), and then:

* The `s` field no longer becomes accessible
* Circe automatic derivation fails, so I can consolidate the decoder/encoder and be sure it's being used

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/platform/transformer/receive/NotificationMessageReceiverTest.scala
```diff
@@ -195,40 +167,32 @@ class NotificationMessageReceiverTest
...
-            bucket,
-            Some(mockSnsClientFailPublishMessage)) { recordReceiver =>
+        withNotificationMessageReceiver(Topic("does-not-exist"), bucket) {
+          recordReceiver =>
             val future = recordReceiver.receiveMessage(message)
 
             whenReady(future.failed) { x =>
-              x.getMessage should be("Failed publishing message")
+              x.getMessage.startsWith("Unknown topic: does-not-exist") shouldBe true
             }
```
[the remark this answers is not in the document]
alexwlchan: Ah, so I tried this when I first wrote the code and the compiler got miserable because `startWith` expected a Word and I couldn't work out how to make it work after three attempts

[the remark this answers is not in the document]
alicefuzier: Ah when it says that it's usually about parenthesis. Try:
```
x.getMessage should startWith ("Unknown topic: does-not-exist")
```

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/platform/transformer/SierraTransformerFeatureTest.scala
```diff
@@ -65,21 +63,21 @@ class SierraTransformerFeatureTest
                 val sourceIdentifier = SourceIdentifier(
                   identifierType = IdentifierType("sierra-system-number"),
                   ontologyType = "Work",
-                  value = "b10010014"
+                  value = sierraTransformable.sierraId.withCheckDigit(
```
alicefuzier: mmmh...isn't this a bit testing the prod code with the same code? Have we kept a test somewhere where we explicitly check the output value against a hardcoded string?

sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/transformable/Transformable.scala
```diff
@@ -14,28 +16,32 @@ case class MiroTransformable(sourceId: String,
...
+    )
+
+  // Because the [[SierraTransformable.itemRecords]] field is keyed by
+  // [[SierraRecordNumber]] in our case class, but JSON only supports string
+  // keys, we need to turn the ID into a string when storing as JSON.
+  //
+  // This is based on the "Custom key types" section of the Circe docs:
+  // https://circe.github.io/circe/codecs/custom-codecs.html#custom-key-types
+  //
+  implicit val keyEncoder: KeyEncoder[SierraRecordNumber] =
```
alicefuzier: Shouldn't  this live in the SierraRecordNumber companion object?

alicefuzier: Ah ok I see

sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/transformable/sierra/SierraRecordNumber.scala
```diff
@@ -0,0 +1,99 @@
...
+  *
+  *  1. It makes the [[s]] field private -- callers have to use the methods
+  *     [[withCheckDigit()]] and [[withoutCheckDigit()]].
+  *  2. It forces callers to consistently load the Circe decoder/encoder,
+  *     rather than relying on automatic case class derivations.  We _need_
+  *     to use the custom decoder when getting bib data from the Sierra API,
+  *     and then it would be good to be consistent internally.
+  *
+  */
+class SierraRecordNumber(s: String) {
```
alicefuzier: This could have a second parameter that is the record type: 
```
class SierraRecordNumber(s: String, recrodType: SierraRecordTypes.Value)
```
This way you would avoid passing the type to the `withCheckDigit` method, which feels a bit weird anyway, cause a bib number is always a bib number and same for an item number. You shouldn't be able to write a bib number prefixed with an i, which you can with this code

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/platform/transformer/SierraTransformerFeatureTest.scala
alexwlchan: See `SierraRecordNumberTest`

sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/transformable/sierra/SierraRecordNumber.scala
alexwlchan: Hmm. But when you're decoding a case class, how do you know if you're a bib record or an item record?

I feel a subclassing API coming on, I'll have a look…

sbt_common/internal_model/src/test/scala/uk/ac/wellcome/models/transformable/sierra/SierraRecordNumberTest.scala
```diff
@@ -0,0 +1,51 @@
+package uk.ac.wellcome.models.transformable.sierra
+
+import org.scalatest.{FunSpec, Matchers}
+import org.scalatest.prop.TableDrivenPropertyChecks._
+
+class SierraRecordNumberTest extends FunSpec with Matchers {
+
+  describe("withCheckDigit") {
+    val testCases = Table(
```
alicefuzier: Are these all different test cases that test different scenarios? If yes, I'd much prefer them to be separate test with individual titles. If not, we should delete the not useful ones. I think in these cases test titles are much better than comments to explain the reason for tests and are much more visible. Code duplication can be reused by using shared functions

sierra_adapter/sierra_items_to_dynamo/src/main/scala/uk/ac/wellcome/platform/sierra_items_to_dynamo/dynamo/package.scala
```diff
@@ -0,0 +1,34 @@
...
+object dynamo {
+  implicit val recordNumberFormat =
+    DynamoFormat
+      .coercedXmap[SierraRecordNumber, String, IllegalArgumentException](
+        SierraRecordNumber
+      )(
+        _.withoutCheckDigit
+      )
+
+  implicit val idGetter: IdGetter[SierraItemRecord] =
```
alicefuzier: I know this is used by the VersionedDao, but it doesn't feel like this pckage is the right place for this. My personal preference would be to put it in the SierraItemRecord companion object

sierra_adapter/sierra_items_to_dynamo/src/main/scala/uk/ac/wellcome/platform/sierra_items_to_dynamo/dynamo/package.scala
```diff
@@ -0,0 +1,34 @@
...
+    (record: SierraItemRecord) => record.id.withoutCheckDigit
+
+  implicit val updateExpressionGenerator
+    : UpdateExpressionGenerator[SierraItemRecord] =
```
alicefuzier: Ditto

sierra_adapter/sierra_items_to_dynamo/src/main/scala/uk/ac/wellcome/platform/sierra_items_to_dynamo/dynamo/package.scala
```diff
@@ -0,0 +1,34 @@
...
+  SierraItemRecord,
+  SierraRecordNumber
+}
+import uk.ac.wellcome.storage.dynamo._
+import uk.ac.wellcome.storage.type_classes.IdGetter
+
+object dynamo {
+  implicit val recordNumberFormat =
+    DynamoFormat
+      .coercedXmap[SierraRecordNumber, String, IllegalArgumentException](
```
alicefuzier: Actually same goes for this..Id' rather have those in their companion object than in a package that then needs to be imported every time they're needed

sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/transformable/sierra/SierraRecordNumber.scala
alicefuzier: Ah I see, Yeah they become two different record numbers a that point

sierra_adapter/sierra_items_to_dynamo/src/main/scala/uk/ac/wellcome/platform/sierra_items_to_dynamo/dynamo/package.scala
alicefuzier: mmh, but you would need the dependecy to scanamo wherever SierreItemRecord and SierraRecordNumber are defined...
Not too fussed about it but I'm not a big fan of these package objects. I think we should keep them to a minimum. They make things hard to find unless you know already where to look for them

alicefuzier:
Look good to me and much better than before. Only thing for me is: can we split the single test in SierraRecordNumberTests into multiple tests with individual titles, please? 
Otherwise good to merge for me

sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/transformable/sierra/SierraRecordNumber.scala
alicefuzier: Actually, I'm not sure that you would need two record number types. You could just a `recordType` field when encoding and then use that when decoding. You would need an encoder and decoder for `SierraRecordTypes`, probably similar to the one we have for `Licence`. Anyway, it should probably be in a separate PR

alexwlchan: It’s so disruptive, I’d rather get it right once than push a broken version that immediately needs patching.

PR #2450, opened 2018-07-23
Merger to merge physical and digital work pairs

mklander:

The merger should merge a pair of digital and physical works to form a work with a physical and digital location and a redirect.


catalogue_pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/Merger.scala
```diff
@@ -0,0 +1,59 @@
...
+      mergePhysicalAndDigitalWorks(physicalWorks, digitalWorks)
+    } else {
+      None
+    }
+  }
+
+  private def mergePhysicalAndDigitalWorks(
+    physicalWorks: Seq[UnidentifiedWork],
+    digitalWorks: Seq[UnidentifiedWork]) = {
+    if (physicalWorks.size == 1 && digitalWorks.size == 1) {
```
alicefuzier: You could use pattern matching here possibly:
```
(physicalWorks, digitalWorks) match {
  case (List(physicalWork), List(digitalWork)) => 
    Some(
        List(
          mergePhysicalWithDigitalWork(physicalWork, digitalWork),
          redirectWork(digitalWork, physicalWork.sourceIdentifier)))
  case _ => None
}
```

catalogue_pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/Merger.scala
```diff
@@ -0,0 +1,59 @@
...
+      sourceIdentifier = workToRedirect.sourceIdentifier,
+      version = workToRedirect.version,
+      redirect = IdentifiableRedirect(redirectTargetSourceIdentifier)
+    )
+  }
+
+  private def isDigitalWork(work: UnidentifiedWork): Boolean = {
+    work.workType match {
+      case None    => false
+      case Some(t) => t.id == "v" && t.label == "E-books"
```
alicefuzier: Why do we have to check for id being "v"? Why is the check on label not enough?

mklander: I have no idea, I think it's just the example I got.  Will check.

mklander: Have confirmed this is correct at present given what Sierra holds, but will change in the future.

PR #2451, opened 2018-07-23
Make matcher process updates for an already stored version of a work

alicefuzier:
The matcher was eating messages without sending them forward or to the dlq. Inspecting the logs I found a bunch of lines like this:
```
14:38:30.420 [main-actor-system-akka.actor.default-dispatcher-5] DEBUG u.a.w.p.m.m.MatcherMessageReceiver - Not matching work due to version: miro-image-number/L0052589 v51 is not newer than existing work v51
```
The current logic doesn't process updates for a version already store on the assumption that, if it's already stored, it's already been sent on. That assumption, though is not correct in the case where something fails in between storing the graph and sending the SNS message, for example unlocking the rows in the lock table. We know that unlocking errors do happen due to ProvisionedThroughputExceeded exception when the table is still warming up, so this would cause messages to be eaten by the matcher

### DIFF
--- a/catalogue_pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/workgraph/WorkGraphUpdater.scala
+++ b/catalogue_pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/workgraph/WorkGraphUpdater.scala
@@ -23,7 +23,7 @@ object WorkGraphUpdater extends Logging {
         throw VersionExpectedConflictException(versionConflictMessage)
       case Some(WorkNode(_, existingVersion, linkedIds, _)) if existingVersion == workUpdate.version && workUpdate.referencedWorkIds != linkedIds.toSet =>
         val versionConflictMessage =
-          s"${workUpdate.workId} v${workUpdate.version} is exists with different content!"
+          s"${workUpdate.workId} v${workUpdate.version} exists with different content!"
         error(versionConflictMessage)
         throw VersionUnexpectedConflictException(versionConflictMessage)
       case _ => doUpdate(workUpdate, existingGraph)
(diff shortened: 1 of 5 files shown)


catalogue_pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/workgraph/WorkGraphUpdater.scala
```diff
@@ -16,18 +17,23 @@ import scala.util.hashing.MurmurHash3
...
+          if existingVersion > workUpdate.version =>
+        val versionConflictMessage =
+          s"${workUpdate.workId} v${workUpdate.version} is not newer than existing work v$existingVersion"
+        debug(versionConflictMessage)
+        throw VersionExpectedConflictException(versionConflictMessage)
+      case Some(WorkNode(_, existingVersion, linkedIds, _))
+          if existingVersion == workUpdate.version && workUpdate.referencedWorkIds != linkedIds.toSet =>
+        val versionConflictMessage =
+          s"${workUpdate.workId} v${workUpdate.version} is exists with different content!"
+        error(versionConflictMessage)
```
mklander: typo, 'is exists' -> 'exists'

catalogue_pipeline/matcher/src/test/scala/uk/ac/wellcome/platform/matcher/workgraph/WorkGraphUpdaterTest.scala
```diff
@@ -158,6 +158,73 @@ class WorkGraphUpdaterTest extends FunSpec with Matchers with MatcherFixtures {
...
+              WorkNode("B", 0, List(), hashed_AB)))
+        )
+        .nodes should contain theSameElementsAs
+        List(
+          WorkNode("A", 2, List("B"), hashed_AB),
+          WorkNode("B", 0, List(), hashed_AB))
+    }
+
+    it(
+      "doesn't an update for the same version if the work is different from the one stored") {
```
mklander: typo, 'doesn't process an update ...'?

catalogue_pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/workgraph/WorkGraphUpdater.scala
alicefuzier: oops

catalogue_pipeline/matcher/src/test/scala/uk/ac/wellcome/platform/matcher/workgraph/WorkGraphUpdaterTest.scala
alicefuzier: and oops

Issue #2361, opened 2018-07-06
Perfect reindexes, with no failures


jtweed:

alexwlchan:
### 10th July

The reindexer ran slowly because DynamoDB didn't scale up correctly.

20 records went to the reindexer DLQ. 347 records weren't indexed into Elasticsearch.

alexwlchan:
### 11th July

We preset the DynamoDB capacity on the SourceData and recorder tables to the max, and disabled autoscaling. The reindexer ran much faster, and no new records ended up on the DLQ.

876 records went to the ingestor DLQ. 29 records went to the merger DLQ. 79 records weren't indexed into Elasticsearch.

alexwlchan:
### 11th July (take 2)

DynamoDB was similarly preset to max capacity.

I’m reducing the max instances on the ingestor to 10 (see 88b9f2b).

PR #2454, opened 2018-07-24
Reduce likelihood of clashes in messages sent by MessageWriter (and the transformer)

alicefuzier:
For #2361

### DIFF
--- a/catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/platform/transformer/MiroTransformerFeatureTest.scala
+++ b/catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/platform/transformer/MiroTransformerFeatureTest.scala
@@ -12,7 +12,7 @@ import uk.ac.wellcome.utils.JsonUtil
 import uk.ac.wellcome.utils.JsonUtil._
 
 class MiroTransformerFeatureTest
-  extends FunSpec
+    extends FunSpec
     with Matchers
     with SQS
     with SNS
@@ -88,14 +88,12 @@ class MiroTransformerFeatureTest
               topic)
 
             withServer(flags) { _ =>
-
               val miroHybridRecordMessage1 =
                 hybridRecordNotificationMessage(
                   message = createValidMiroTransformableJson(
                     MiroID = "L0011975",
                     MiroCollection = "images-L",
-                    data =
-                      """
+                    data = """
                       {
                           "image_cleared": "Y",
                           "image_copyright_cleared": "Y",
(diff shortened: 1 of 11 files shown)


catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/platform/transformer/MiroTransformerFeatureTest.scala
```diff
@@ -72,4 +72,96 @@ class MiroTransformerFeatureTest
...
+              )
+
+              sqsClient.sendMessage(
+                queue.url,
+                JsonUtil.toJson(miroHybridRecordMessage2).get
+              )
+              eventually {
+                val works = getMessages[UnidentifiedWork](topic)
+                works.distinct.length shouldBe >=(2)
+
```
mklander: I don't think this is wrong, but it's not obvious to me why there can be more than, rather than exactly, 2 works after processing.

sbt_common/messaging/src/test/scala/uk/ac/wellcome/messaging/message/MessageWriterTest.scala
```diff
@@ -93,4 +95,27 @@ class MessageWriterTest
     }
   }
 
+  it("gives distinct s3 keys when sending the same message twice") {
+    withLocalSnsTopic { topic =>
+      withLocalS3Bucket { bucket =>
+        withExampleObjectMessageWriter(bucket, topic) { messageWriter =>
+          val eventualAttempt1 = messageWriter.write(message, subject)
+          Thread.sleep(2)
```
mklander: I guess the `sleep` here is to make sure these occur in different epochs and hence different keys?  Perhaps make this a bit clearer?

catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/platform/transformer/MiroTransformerFeatureTest.scala
alexwlchan: Hmm. Possibly for more-than-once delivery, but we'd get duplicate SNS messages – not duplicate works.

I think we should change this to `shouldBe 2`.

alicefuzier: yeah you're right. It's a copy pasta error from test where we have to account for at least once delivery

sbt_common/messaging/src/test/scala/uk/ac/wellcome/messaging/message/MessageWriterTest.scala
alicefuzier: 👍

PR #2459, opened 2018-07-26
Move Sourced out of the common lib

alexwlchan:
internal_model makes more sense as a home for it, and gets one more piece out of the common-common lib.


alexwlchan:
Merging by fiat, this is a tiny change.

PR #2460, opened 2018-07-26
[WIP] Use the monitoring library from the external repo

alexwlchan:


catalogue_pipeline/merger/src/test/scala/uk/ac/wellcome/platform/merger/services/MergerWorkerServiceTest.scala
```diff
@@ -98,7 +104,8 @@ class MergerWorkerServiceTest
             val worksSent = getMessages[BaseWork](topic)
             worksSent should contain only recorderWorkEntry.work
 
-            assertSuccessMetricIncremented(metricsSender)
+            verify(metricsSender, times(1))
+              .countSuccess(any[String])
```
alicefuzier: Does this mean we have to change all usages of gracefulFailureException to RecognisedFailureException? I yes, then I think it should be done in this PR otherwise monitoring will stop working as expected

alexwlchan: No – I created a new `RecognisedFailureException` in the monitoring stack that we could catch on if you pass a Future into count, and added helper methods for it, but realised we never actually need to throw it in the platform.

If you look at SQSStream, we now call one of three helpers – `countSuccess`, `countFailure` and `countRecognisedFailure` – based on the type of exception we got. But we don't need to pass it into monitoring.

I did think about swapping all the GracefulFailureException's around, but that gets very choppy. This is neater.

alicefuzier: 👍

PR #2462, opened 2018-07-26
Undo renaming of maybeBibData 

alicefuzier:
So that data in the SourceData table is still transformable

### DIFF
--- a/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/SierraTransformableTransformer.scala
+++ b/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/SierraTransformableTransformer.scala
@@ -43,7 +43,7 @@ class SierraTransformableTransformer
         )
       )
 
-      sierraTransformable.maybeBibRecord
+      sierraTransformable.maybeBibData
         .map { bibData =>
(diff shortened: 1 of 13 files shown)


alexwlchan:
Mhmm

What about itemData?

I am going to want to change this back when I reharvest all the Sierra stuff.

alicefuzier:
Yep, changed itemData as well

alicefuzier:
We're going to have to be careful about this even when we do the reharvest though. It will make the current data in the source data table not transformable, so we will not be able to do reindexes for the whole time that the rehavest take

alexwlchan:
We can deploy the Sierra adapter separately from the transformer, so the existing data can still be used

On 26 Jul 2018, at 17:01, Alice Fuzier Cayla <notifications@github.com<mailto:notifications@github.com>> wrote:


We're going to have to be careful about this even when we do the reharvest though. It will make the current data in the source data table not transformable, so we will not be able to do reindexes for the whole time that the rehavest take

alicefuzier:
Yeah but that model is shared between the transformer and the sierra adapter. If you change it we have to pin the transformer so that we don't deploy a version with the new model until the new table is ready to be used

alicefuzier:
I'm going to merge this, so transformer images published by travis are able to transform the data we currently have

PR #2464, opened 2018-07-26
Matcher - use sha256 for componentId hashing

mklander:
Switch to SHA256 when hashing the matcher componentId field, to avoid collisions


catalogue_pipeline/matcher/src/test/scala/uk/ac/wellcome/platform/matcher/workgraph/WorkGraphUpdaterTest.scala
```diff
@@ -13,12 +13,18 @@ class WorkGraphUpdaterTest extends FunSpec with Matchers with MatcherFixtures {
   // works comprised of linked works are identified by compound id together with their linked works.
   // A+B:(A->B, B) means compound work with id "A+B" is made of work A linked to B and work B.
 
-  private val hashed_A = "217b21f3" // ciHash("A")
-  private val hashed_B = "4d7e95ff" // ciHash("B")
-  private val hashed_C = "8171cb29" // ciHash("C")
-  private val hashed_AB = "2f2f1575" // ciHash("A+B")
-  private val hashed_ABC = "45e72cb2" // ciHash("A+B+C")
-  private val hashed_ABCD = "73200083" // ciHash("A+B+C+D")
+  private val hashed_A =
```
alexwlchan: Why not compute these values in the test, rather than hard-coding strings and relying on the comments?

mklander: Yes, I have had it both ways.  If it is computed from the `ciHash` method this just repeats the code under test, which is why I swapped it for the hard coded outputs.  Also, having the strings in the test makes it clearer what the keys actually look like if you want to understand the code.

I see merits in both.

alexwlchan: It repeats the code under test if `ciHash` has the same implementation as `componentIdentifier`.

Perhaps have some separate tests for `componentIdentifier` that use hard-coded strings?

```scala
componentIdentifier(List("A")) shouldBe "559aead08264d5795d3909718cdd05abd49572e84fe55590eef31a88a08fdffd"
```

and so on. Best of both worlds.

catalogue_pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/workgraph/WorkGraphUpdater.scala
```diff
@@ -68,18 +68,16 @@ object WorkGraphUpdater extends Logging {
...
-              componentIdentifierHash(nodeIds))
+              componentIdentifier(nodeIds))
           })
         })
         .toSet
     )
   }
 
-  def componentIdentifierHash(nodeIds: List[String]) = {
+  def componentIdentifier(nodeIds: List[String]) = {
```
alicefuzier: should this be private?

catalogue_pipeline/matcher/src/test/scala/uk/ac/wellcome/platform/matcher/workgraph/WorkGraphUpdaterTest.scala
alexwlchan: Does it matter if those component IDs change? If so, we should stay that they’re deterministic.

alexwlchan: /s/stay/test

catalogue_pipeline/matcher/src/main/scala/uk/ac/wellcome/platform/matcher/workgraph/WorkGraphUpdater.scala
mklander: Yup, definitely should be private 😃

alicefuzier:
I'm going to merge as is 'cause it's fixing a potential issue and it's already good enough. We can improve the tests in another PR but don't want to leave this open forever

PR #2465, opened 2018-07-26
*slaps roof of Scala* This type system can hold so many Sierra records

alexwlchan:
Ahem.

Comments from @alicefuzier and @mklander on #2446 got me thinking – why don’t we bake the Sierra record type _into_ the Sierra record number? It would be an error to use a bib record number in place of an item, and vice versa. Plus it makes the proposed API a lot cleaner.

Right now you can do `withCheckDigit(bibs)` on an item number, or vice versa. That doesn’t seem right!

In most cases, we can definitely say if something should be a bib or an item ID, and assign it a record type accordingly. But in the Sierra reader, we have a generic `SierraRecord` type, which contains an “untyped” Sierra API response. What record type do we use there?

Little secret: we actually do know what it is, we just don’t type it. Well, let’s lean more heavily on the type system. This patch modifies the Sierra reader to emit *typed* records in all cases, which unlocks the door for typed record numbers elsewhere.

---

Yes, this will almost certainly cause issues with the current Sierra mergers. Since we're reharvesting all this data soon, I’m taking the opportunity to clean up cruft in the Sierra adapter stack while I’ve got the chance.


sbt_common/internal_model/src/test/scala/uk/ac/wellcome/models/transformable/sierra/SierraItemRecordTest.scala
```diff
@@ -0,0 +1,83 @@
...
+  it("throws an exception when bibIds is not a list") {
+    assertCreatingFromDataFails(
+      data = """{"bibIds":"blah"}""",
+      expectedMessage = "CanBuildFrom for A: DownField(bibIds)"
+    )
+  }
+
+  private def assertCreatingFromDataFails(
+    data: String,
+    expectedMessage: String
```
alicefuzier: is this used?

sierra_adapter/sierra_reader/src/main/scala/uk/ac/wellcome/platform/sierra_reader/modules/WindowManager.scala
```diff
@@ -52,28 +51,23 @@ class WindowManager @Inject()(
...
-
-              val newId = (unprefixedId.toInt + 1).toString
-              WindowStatus(id = Some(newId), offset = offset + 1)
-            case None =>
-              throw GracefulFailureException(
-                new RuntimeException("Json did not contain an id"))
-          }
-
-        // Let it throw the exception if it's a failure
-        triedStatus.get
```
alicefuzier: oh god the horribleness we had here....Looks much nicer now!

sierra_adapter/sierra_reader/src/main/scala/uk/ac/wellcome/platform/sierra_reader/services/SierraReaderWorkerService.scala
```diff
@@ -43,9 +55,9 @@ class SierraReaderWorkerService @Inject()(
     process = processMessage
   )
 
-  def processMessage(norificationMessage: NotificationMessage): Future[Unit] =
+  def processMessage(notificationMessage: NotificationMessage): Future[Unit] =
     for {
-      messageString <- Future(norificationMessage.Message)
+      messageString <- Future(notificationMessage.Message)
```
alicefuzier: We should use `Future.successful` if there isn't an actually blocking operation (i know this is how it was before but it would be good to fix it anyway)

PR #2466, opened 2018-07-26
Correct merged digital and physical works

mklander:
* Merged digital and physical works add the location from the digital work to the locations present in the single physical item.  This replaces current incorrect behaviour that creates two item records, one for each location.

* Add a new `iiif-presentation` type for use in digital locations

### DIFF
--- a/catalogue_pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/Merger.scala
+++ b/catalogue_pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/Merger.scala
@@ -54,7 +54,11 @@ class Merger extends Logging {
 
   private def mergePhysicalWithDigitalWork(physicalWork: UnidentifiedWork,
                                            digitalWork: UnidentifiedWork) = {
-    physicalWork.copy(items = physicalWork.items ++ digitalWork.items)
+    physicalWork.copy(items = List(
+      physicalWork.items.head.copy(
+        agent = physicalWork.items.head.agent.copy(
+          locations =
+            physicalWork.items.head.agent.locations ++ digitalWork.items.head.agent.locations))))
   }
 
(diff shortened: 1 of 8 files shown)


catalogue_pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/Merger.scala
```diff
@@ -54,7 +54,9 @@ class Merger extends Logging {
 
   private def mergePhysicalWithDigitalWork(physicalWork: UnidentifiedWork,
                                            digitalWork: UnidentifiedWork) = {
-    physicalWork.copy(items = physicalWork.items ++ digitalWork.items)
+    physicalWork.copy(items = List(physicalWork.items.head.copy(
```
alexwlchan: This would be easier to read with a few intermediate variables. How about:

```scala
val physicalItem = physicalWork.items.head
val digitalItem = digitalWork.items.head

val newItem = physicalItem.copy(
  locations = physicalItem.agent.locations ++ digitalItem.agent.locations
)

physicalWork.copy(items = List(newItem))
```

sbt_common/internal_model/src/main/resources/location-types.csv
```diff
@@ -1,5 +1,6 @@
 ThumbnailImage,thumbnail-image,Thumbnail Image
-IIIFImage,iiif-image,IIIF image
+IIIFImage,iiif-image,IIIF Image API
+IIIFPresentation,iiif-presentation,IIIF Presentation API
```
alexwlchan: This file also lives in the ontologies folder, under Reference Data. It needs updating there as well.

(Yes, the duplication is daft, but not something to fix today.)

alicefuzier:
Ah ignore me, just noticed that they're here

PR #2468, opened 2018-07-27
Please have your ID ready for inspection

alexwlchan:
One of the complications in #2446 are the `SierraBibData` and `SierraItemData` types. They have to parse the ID from the Sierra JSON – even though we already parsed the ID when we did the original read from Sierra, and we're keeping it as a field on `SierraBibRecord` and `SierraItemRecord`.

If you look at the Miro transformer, we don't parse the ID twice – we pass it around as a string into the transformer methods that need it.

The benefit for #2446 is that we can read the ID once with circe-optics (in the sierra_reader), then stash it as typed object and don't need special decoders to get the ID from the JSON when we re-parse it.

### DIFF
--- a/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraItems.scala
+++ b/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraItems.scala
@@ -34,14 +34,14 @@ trait SierraItems extends Logging with SierraLocation {
       .flatten
       .toMap
 
-  def transformItemData(sierraItemData: SierraItemData): Identifiable[Item] = {
-    debug(s"Attempting to transform ${sierraItemData.id}")
+  def transformItemData(sierraItemId: String, sierraItemData: SierraItemData): Identifiable[Item] = {
+    debug(s"Attempting to transform $sierraItemId")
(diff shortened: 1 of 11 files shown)


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraItems.scala
```diff
@@ -18,52 +18,60 @@ import scala.util.{Failure, Success}
...
+    sierraTransformable: SierraTransformable): Map[String, SierraItemData] =
+    sierraTransformable.itemData
+      .map { case (id, itemRecord) => (id, itemRecord.data) }
+      .map {
+        case (id, jsonString) =>
+          fromJson[SierraItemData](jsonString) match {
+            case Success(data) => Some(id -> data)
+            case Failure(e) => {
+              error(s"Failed to parse item!", e)
+              None
```
alexwlchan: It’s a discussion for another time, but I don't think we should silently drop malformed records here. We should made a dashboard go red, because this should really never happen.

mklander: eek, straight to DLQ 🗑

PR #2470, opened 2018-07-27
Bucket-to-bucket archive storage bagit flow

tomcrane:
This is an experiment with the flow that will eventually, for a given identifier (for now, only b numbers of digitised objects) collect METS, METS-ALTO, and digitised assets from various locations, create a bagit ingest package, drop it in an ingest bucket, and initiate the archival storage process for that identifier.

This serves three purposes:

1. A tool to create a vast supply of test content for the archival storage service by bagging existing things
2. A means of evolving the specification for what Intranda need to change in Goobi, by transforming METS into what we think we want
3. The code to migrate every digitised thing in Preservica into the new storage


tomcrane:
Currently only works for single-manifestation METS files (not multi volume works) and doesn't actually collect the assets.

PR #2473, opened 2018-07-30
JSON decoding errors in the Sierra transformer should bubble up, not be silently dropped

### LINKED ISSUE #2469: Don't silent drop item records which don't validate as Sierra item JSON
As mentioned on #2468.

alexwlchan:
You know how I said we should fix all the errors in the Sierra transformer? Yeah, about that…

Previously we’d quietly discard any items which can't be decoded as JSON – even though we know we're getting JSON from the Sierra API, and a problem decoding it as our ItemData model is the sort of thing we'd like to know about.

In theory, we should never hit this codepath – but until now, we've had no way of knowing that, so let's find out.

Resolves #2469.


alexwlchan:
> Why the hell did we do that for?

Given how much of the transformer code I wrote, I took the ~cowardly~ diplomatic decision not to inspect the commit history. 😜

PR #2475, opened 2018-07-30
Working on archive asset lookup lambda 

fractos:
This PR will create a Lambda that will serve requests from the DDS and potentially other systems. The requests will reference identifiers (e.g. b-numbers) and will yield a storage manifest location in S3 from a backing DynamoDB table. The S3 location will be dereferenced and the contents of the S3 object will be returned as the response body to the caller.

Assumptions:
- the body of a storage manifest must be dereferenced from a URI in the DynamoDB table
- the DynamoDB table has a key called `id` and a field called `s3key`


archive/archive_asset_lookup/src/endpoint.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def lambda_handler(event, context):
+
+    # setup logging and AWS clients
+    logger = logging.getLogger()
+    dynamodb = boto3.resource('dynamodb', region_name=settings.AWS_REGION)
+    s3 = boto3.resource('s3', region_name=settings.AWS_REGION)
+
+    logger.info('got event{}'.format(event))
```
kenoir: We usually try and break the entry point from the main application code and hand dependencies to that for easier testing.

e.g.

```python
def lambda_handler(event, context):
  logger = logging.getLogger()
  dynamodb = boto3.resource('dynamodb', region_name=settings.AWS_REGION)
  s3 = boto3.resource('s3', region_name=settings.AWS_REGION)

def asset_lookup(dynamodb, table_name, identifier):
  table = dynamodb.Table(table_name)
  # ... etc etc 
```

fractos: Yep good plan, I'll work that in for next version.

archive/archive_asset_lookup/src/settings.py
```diff
@@ -0,0 +1,6 @@
+import os
```
kenoir: I like this pattern.

archive/archive_asset_lookup/src/endpoint.py
```diff
@@ -0,0 +1,103 @@
+import os
+import tempfile
+import logging
```
kenoir: We tend to just print ... stdout gets logged in lambdas. Worth asking @alexwlchan if there is another reason.

archive/archive_asset_lookup/src/endpoint.py
```diff
@@ -0,0 +1,103 @@
...
+import boto3
+import botocore
+import settings
+from boto3.dynamodb.conditions import Key
+
+
+def lambda_handler(event, context):
```
kenoir: The aws-utils package has some helpers for lambdas that might be worth a look (see the snapshot_scheduler lambda).

archive/archive_asset_lookup/src/endpoint.py
fractos: mostly there for isomorphism really, I think it'll pick up whatever ambient settings for the logging have been set up (if this were called as a module)

archive/archive_asset_lookup/src/endpoint.py
fractos: we're just changing this to use a `main` function so it works with the lambda terraform module properly. will check it out!

kenoir:

archive/archive_asset_lookup/src/endpoint.py
fractos: Turned out there were various hoops to jump through to get standard logging working in Lambda which are now in the `core-archive-function` branch.

Issue #2422, opened 2018-07-16
The Sierra reader isn't resuming incomplete windows correctly


alexwlchan:
See discussion on https://github.com/wellcometrust/platform/pull/2418/files#r202581501

Blocks #2418.

alexwlchan:
This is fixed and we've reharvested. Closing.

PR #2478, opened 2018-07-30
Make sure the Sierra reader resumes windows correctly

alexwlchan:
Code fix for #2422.

The test for resuming windows is seeded with real reader output, so if the format changes again the test will still catch it. And now we validate Sierra record numbers passed around internally as valid, where previously we just passed around unchecked strings.

### DIFF
--- a/sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/transformable/sierra/SierraRecordNumber.scala
+++ b/sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/transformable/sierra/SierraRecordNumber.scala
@@ -62,9 +62,7 @@ sealed trait SierraTypedRecordNumber extends SierraRecordNumber {
   }
 }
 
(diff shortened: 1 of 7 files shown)


sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/transformable/sierra/SierraRecordNumber.scala
```diff
@@ -6,7 +6,7 @@ object SierraRecordTypes extends Enumeration {
 
 sealed trait SierraRecordNumber {
   val recordNumber: String
-  val recordType: SierraRecordTypes.Value
+  val recordType: Option[SierraRecordTypes.Value]
```
alicefuzier: I'm not sure I like this being optional and then `withCheckDigit` potentially throwing an exception. It feels like `recordType` should not exist at all in `UntypedSierraRecordNumber`

Issue #2280, opened 2018-06-25
Be better about handling Sierra records with a 260 and a 264


alexwlchan:
Spun out from #2252.

There are 92 Sierra records with both a 260 and a 264, which the transformer currently flags as a cataloguing error.

It looks like, at least in some cases, the 264 field only contains a copyright statement, which we discard. For example:

```
260    Würzburg :|bKönigshausen & Neumann,|c[2015] 
264  4 |c©2015 
```

<details>
<summary>Afflicted records</summary>

1528064
2064911
2119618
2119830
2124305
2124759
2124895
2124946
2126925
2126978
2132114
2133427
2136956
2148880
2150151
2153999
2159646
2163242
2163244
2163272
2173146
2177570
2199854
2199855
2200264
2200398
2200399
2200481
2256170
2264988
2267509
2277618
2347667
2397796
2398615
2438982
2438986
2439205
2472339
2472341
2472344
2472349
2474267
2474409
2474435
2475494
2475517
2476701
2476718
2477059
2477190
2477191
2477299
2490813
2491108
2491132
2491381
2496696
2497815
2498764
2500828
2500878
2535394
2707300
2823118
2826732
2846303
2846354
2846850
2846902
2846904
2846920
2847024
2849024
2851379
2851422
2851473
2852703
2853511
2855098
2855232
2855775
2855867
2871407
2881998
2928394
2992788
3006352
3015891
3016012
3034408
3043187

</details>

alexwlchan:
We can apply two fairly blunt heuristics to reduce the size of this problem:

* If the only thing in 264 is subfield c with something matching `©\d{4}`, we throw away 264 and just use 260
* If 260 and 264 have matching subfields, ditto

That leaves us with four records which will hit this error, which have moderately non-trivial resolutions. I've flagged them to Branwen, because I suspect the best fix is updating Sierra:

```
('2173146.json',
 {'260': [('a', '[Washington, D.C.] :'),
          ('b', 'Congressional Research Service,'),
          ('c', '[2014]')],
  '264': [('a', 'Marston Gate, Great Britain :'),
          ('b', 'amazon.co.uk, Ltd.,'),
          ('c', '[2014?]')]})
('2398615.json',
 {'260': [('b', 'Little Brown & Co'), ('c', '2015.')],
  '264': [('a', 'New York :'),
          ('b', 'Little, Brown and Company,'),
          ('c', '[2015]')]})
('2823118.json',
[...]

PR #2480, opened 2018-07-30
Improve our handling of Sierra records that have both a 260 and a 264 field

alexwlchan:
We use 260 and 264 for production events, and Silver's original rules say “if both are present, that’s a cataloguing error” – and the original transformer raised an error, sending 95 records to a DLQ.

This patch applies two blunt heuristics to reduce that number to 4:

* If 264 only contains a copyright statement, discard it
* If 260 and 264 have matching subfields, discard 264

Part of #2280.

### DIFF
--- a/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraProduction.scala
+++ b/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraProduction.scala
@@ -146,10 +146,12 @@ trait SierraProduction {
         )
       }
 
-  // In general, it's a cataloguing error for a bib record to have both
-  // 260 and 264 fields.  Sometimes we can detect the duplicate data, and
-  // still get a useful result.
-  //
+  /** Populate the production data if both 260 and 264 are present.
+    *
+    * In general, this is a cataloguing error, but sometimes we can do
+    * something more sensible depending on if/how they're duplicated.
+    *
+    */
   private def getProductionFromBothFields(
     marc260fields: List[VarField],
     marc264fields: List[VarField]) = {
--- a/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraProduction.scala
+++ b/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraProduction.scala
@@ -33,7 +33,8 @@ trait SierraProduction {
       case (Nil, Nil)           => List()
       case (marc260fields, Nil) => getProductionFrom260Fields(marc260fields)
       case (Nil, marc264fields) => getProductionFrom264Fields(marc264fields)
-      case (marc260fields, marc264fields) => getProductionFromBothFields(marc260fields, marc264fields)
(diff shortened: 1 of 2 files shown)


catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraProduction.scala
```diff
@@ -150,6 +147,42 @@ trait SierraProduction {
...
+        marc264fields.head.subfields.head.content.matches("^©\\d+$")) {
+      getProductionFrom260Fields(marc260fields)
+    }
+
+    // We've also seen cases where the 260 and 264 field are both present,
+    // and they have matching subfields!  We use the 260 field as it's not
+    // going to throw an exception about unrecognised second indicator.
+    else if (marc260fields.map { _.subfields } == marc264fields.map {
+               _.subfields
+             }) {
```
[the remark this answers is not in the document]
alexwlchan: (This only affects a dozen records, all of which have matching order. If it turns out to be insignificant, those records will land on the DLQ – and then we can deal with them.)

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraProduction.scala
```diff
@@ -150,6 +147,42 @@ trait SierraProduction {
...
+    // We've seen cases where the 264 field only has the following subfields:
+    //
+    //      [('tag', 'c'), ('content', '©2012')]
+    //
+    // or similar, and the 260 field is populated.  In that case, we can
+    // discard the 264 and just use the 260 fields.
+    if (marc264fields.length == 1 &&
+        marc264fields.head.subfields.length == 1 &&
+        marc264fields.head.subfields.head.tag == "c" &&
+        marc264fields.head.subfields.head.content.matches("^©\\d+$")) {
```
alexwlchan: Tighten this regex – it should be 4 digits.

mklander: 👍

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraProduction.scala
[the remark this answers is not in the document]
mklander: Thanks, yes better to be tighter in the match

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraProduction.scala
```diff
@@ -147,10 +147,17 @@ trait SierraProduction {
...
       case List(
-        VarField(_, _, Some("260"), _, _, , List(MarcSubfield("c", content)))) =>
+          VarField(
+            _,
+            _,
+            Some("264"),
+            _,
+            _,
+            List(MarcSubfield("c", content)))) =>
         content.matches("^©\\d{4}$")
```
mklander: !!

Issue #2430, opened 2018-07-18
Separate the messaging lib from the monorepo


alexwlchan:
The natural follow-up to #2360.

- [x] Move JsonUtil into a standalone library
- [ ] Modify monitoring/storage to use the external JsonUtil
- [ ] Have another go at moving the messaging library out

alexwlchan:
Waiting for Travis to release scala-json 1.0.

PR #2481, opened 2018-07-30
Changes to messaging to pull it out of the monorepo

alexwlchan:
For #2430. I’m not actually going to merge this, but it shows the changes I‘m planning to make to split out the messaging library from the rest of the repository. Comments welcome; I’ll do the first release of scala-messaging tomorrow.

No changes to the external API, just adding the internal helpers we no longer get from the main repo.


kenoir:
Looks like a good approach to me 👍

alicefuzier:
Uhm the only thing that I'm a bit wary about is not using JsonUtil and using circe parsers directly. JsonUtil has a bunch of customisations that we rely on, so circe parsers would probably work 80% of the time but not a 100. For example in JsonUtil we configure circe to use the type discrimitaro field instead of the default wrapper and we also extend TimeInstances which gives us encoder and decoder for instant and other time classes, which are not available by default.
Would there be a way of making JsonUtil a part of messaging?

sbt_common/messaging/src/test/scala/uk/ac/wellcome/messaging/utils/JsonUtil.scala
```diff
@@ -0,0 +1,44 @@
...
+import io.circe.parser._
+import io.circe.syntax._
+import io.circe.{Decoder, Encoder, Json}
+import org.scalatest.Matchers
+
+import scala.util.Try
+
+object JsonUtil extends AutoDerivation with TimeInstances with Matchers {
+
+  private[messaging] def toJson[T](value: T)(implicit encoder: Encoder[T]): Try[String] = {
```
alicefuzier: Why private?

alexwlchan:
Doesn't passing around the implicit decoders protect us from that?

alexwlchan:
Closing due to inactivity.

Issue #2163, opened 2018-05-31
The transformer is rejecting some Sierra records that have concept identifiers that we don't expect


alexwlchan:
Specifically, ~400–500 records are hitting an error like this:

```
java.lang.RuntimeException: Too many identifiers fields: List(MarcSubfield(0,D016615Q000502), MarcSubfield(0,(DNLM)D016615Q000502))
```

When we look in the MARC tag for a subject/genre, we inspect the contents of subfield(s) $0 for the identifier value. If there are multiple subfields $0, we error out. This is a list of all the records causing us an issue, and their associated subfields:

<details>

```
1000255
['D000056', 'D000056']
['D000062', 'D000062']
['D012449', 'D012449']
['D004496', 'D004496']
1014089
['(FrPBN)11975999', '(FrPBN)11933219', '(FrPBN)11931476']
['(FrPBN)11976032', '(FrPBN)11933219', '(FrPBN)11931476']
1019549
['(DE-601)09134591X', '(DE-STW)13444-6']
['(DE-601)091369827', '(DE-STW)17707-1']
['(DE-601)091374189', '(DE-STW)12945-3']
['(DE-601)091400961', '(DE-STW)15706-6']
1023833
['(DE-601)091353637', '(DE-STW)18430-6']
['(DE-601)091353645', '(DE-STW)12996-0']
['(DE-601)09136079X', '(DE-STW)18775-0']
['(DE-601)091383234', '(DE-STW)15669-0']
1029438
['(DE-601)091364442', '(DE-STW)17374-2']
1032584
['(DE-601)091353645', '(DE-STW)12996-0']
['(DE-601)091354749', '(DE-STW)18012-3']
1089724
['(DE-601)091352223', '(DE-STW)19203-5']
1099473
['(DE-601)091350441', '(DE-STW)11290-3']
['(DE-601)091396867', '(DE-STW)17829-1']
1100664
['(DE-601)091359244', '(DE-STW)15938-6']
['(DE-601)091361087', '(DE-STW)15920-4']
['(DE-601)091399882', '(DE-STW)11284-5']
1151463
1164290
['sh 85133698 ', 'sh 85133700']
1166054
['sh2017003446', 'sh 85040190']
1184362
['sh 85133698 ', 'sh 85133700']
1189007
1190704
1204552
['(DE-601)091354595', '(DE-STW)15916-2.']
['(DE-601)091420334', '(DE-STW)16834-6.']
1204808
[...]

PR #2490, opened 2018-08-01
Ignore extra identifiers in subfield 0 when identifying concepts

alexwlchan:
A partial fix for #2163. Long term it would be nice to do something other than “throw away these identifiers”, but until we get a better solution this will at least clean up the DLQs.

### DIFF
--- a/catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraConceptsTest.scala
+++ b/catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraConceptsTest.scala
@@ -1,12 +1,7 @@
 package uk.ac.wellcome.platform.transformer.transformers.sierra
 
 import org.scalatest.{FunSpec, Matchers}
-import uk.ac.wellcome.models.work.internal.{
-  Concept,
-  Identifiable,
-  IdentifierType,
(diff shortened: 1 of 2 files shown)


catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraConceptsTest.scala
```diff
@@ -81,25 +76,27 @@ class SierraConceptsTest extends FunSpec with Matchers {
...
-          marcTag = "655",
-          indicator1 = "",
-          indicator2 = "0",
-          subfields = List(
-            MarcSubfield(tag = "a", content = "hitchhiking"),
-            MarcSubfield(tag = "0", content = "u/xxx"),
-            MarcSubfield(tag = "0", content = "u/yyy")
-          )
+  it("ignores multiple instances of subfield 0 im the otherIdentifiers") {
+    val concept = Concept(label = "Hitchhiking horses hurry home")
```
mklander: typo: s/im/in

mklander: I agree it's better to process these records than to throw errors, but I'm starting to wonder about the amount of data cleaning we are pushing into the transformers.

PR #2491, opened 2018-08-01
Move JsonUtil out of the repo so it can be used by external libraries

alexwlchan:


sbt_common/display/src/test/scala/uk/ac/wellcome/display/test/util/JsonMapperTestUtil.scala
```diff
@@ -3,9 +3,9 @@ package uk.ac.wellcome.display.test.util
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.google.inject.{Guice, Injector}
 import uk.ac.wellcome.display.modules.DisplayJacksonModule
-import uk.ac.wellcome.test.utils.JsonTestUtil
+import uk.ac.wellcome.json.utils.JsonAssertions
 
-trait JsonMapperTestUtil extends JsonTestUtil {
+trait JsonMapperTestUtil extends JsonAssertions {
```
mklander: `JsonAssertions` is nicer

mklander: Looks fine to me, but I think @alicefuzier should take a look because Json parsing is so critical. 

I'm not sure why we have introduced `JsonDecodingError`.

alexwlchan:
scala-json repo is over here: https://github.com/wellcometrust/scala-json

Identical to the code I removed from master, modulo naming.

Previously `fromJson` would throw a GracefulFailureException from the common lib; it now throws the more specific JsonDecodingError.

I think we should get away from the model where _everything_ throws a GFE (which is also quite badly named) – in this case, there’s a good name for the single error that library throws, so I used it.

mklander:
Agree regarding throwing the more specific `JsonDecodingError` exception, however wrapping exceptions is used to indicate the severity and recoverability of errors. :+1 regarding renaming `GracefulFailureException` it's not very intent revealing.

PR #2492, opened 2018-08-01
My patience has expired [remove ExtendedPatience]

alexwlchan:
Another piece of the common-common lib we don't need. I've been replacing ExtendedPatience with IntegrationPatience as I move libraries which used it, and figured I'd try doing it across the entire repo.


kenoir:
I'd say this is only worth it if the number of projects in which this change makes no difference is only a few (one or two). That way we could justify copy pasting the `ExtendedPatience` code where required.

alexwlchan:
I may come back to this, but not right now.

alexwlchan:
Closing due to inactivity.

PR #2494, opened 2018-08-01
[WIP] Modify sierra_items_to_dynamo to save items in a VHS, not DynamoDB

### LINKED ISSUE #2489: [RFC] Modify the sierra_items_to_dynamo app to use a VHS instead of DynamoDB
With strongly typed Sierra numbers, the items_to_dynamo app can't read anything off its queue. Why? Because we serialise the item records using `DynamoFormat[SierraItemRecord]` when they're saved to DynamoDB, but we have no way to deserialise them when getting them from a queue. We could try to find a way to unpack a JSON string using `DynamoFormat`, but it's going to be a pain to test.

I think we should modify the items_to_dynamo app to use a VHS.

Then serialisation is all in JSON, we can ditch a bunch of custom DynamoDB logic, and we get neater DynamoDB tables.

alexwlchan:
This lets us delete a bunch of custom DynamoDB serialisation logic from items_to_dynamo, and just use regular JSON deserialisation on the other side.

Closes #2489.


sierra_adapter/sierra_items_to_dynamo/src/test/scala/uk/ac/wellcome/platform/sierra_items_to_dynamo/services/DynamoInserterTest.scala
```diff
@@ -1,279 +1,294 @@
...
-
-        val futureUnit = dynamoInserter.insertIntoDynamo(record)
-
-        whenReady(futureUnit) { _ =>
-          Scanamo.get[SierraItemRecord](dynamoDbClient)(table.name)(
-            'id -> record.id.withoutCheckDigit) shouldBe Some(
-            Right(record.copy(version = 1)))
+      withLocalS3Bucket { bucket =>
+        withTypeVHS[SierraItemRecord, EmptyMetadata, Assertion](bucket, table) {
+          versionedHybridStore =>
```
alicefuzier: You could use the `DynamoInserterFixture` here couldn't you?

sierra_adapter/sierra_items_to_dynamo/src/test/scala/uk/ac/wellcome/platform/sierra_items_to_dynamo/services/SierraItemsToDynamoWorkerServiceTest.scala
```diff
@@ -33,116 +36,153 @@ class SierraItemsToDynamoWorkerServiceTest
...
+              }
+          }
+        }
+      }
+    }
+  }
 
-        val expectedBibIds = List(bibIds(2), bibIds(3), bibIds(4))
-        val expectedUnlinkedBibIds = List(bibIds(0), bibIds(1))
+  def storeSingleRecord(
```
alicefuzier: Can we put this somewhere where it can be shared? Possibly as a type generic function in `LocalVersionedHybridStore`?

alexwlchan: I tried that, but it got upset about the implicit HybridRecordEnricher when I passed around a generic VHS instance and I couldn't be bothered to make it work properly!

alexwlchan:
It needs some Terraform, but I’ll do that separately.

sierra_adapter/sierra_items_to_dynamo/src/test/scala/uk/ac/wellcome/platform/sierra_items_to_dynamo/services/SierraItemsToDynamoWorkerServiceTest.scala
alicefuzier: ok fair enough

sierra_adapter/sierra_items_to_dynamo/src/test/scala/uk/ac/wellcome/platform/sierra_items_to_dynamo/services/DynamoInserterTest.scala
```diff
@@ -1,279 +1,293 @@
...
-
-        val futureUnit = dynamoInserter.insertIntoDynamo(oldRecord)
-        whenReady(futureUnit) { _ =>
-          Scanamo.get[SierraItemRecord](dynamoDbClient)(table.name)(
-            'id -> newRecord.id.withoutCheckDigit) shouldBe Some(
-            Right(newRecord))
+      withLocalS3Bucket { bucket =>
+        withTypeVHS[SierraItemRecord, EmptyMetadata, Assertion](bucket, table) {
+          versionedHybridStore =>
+            val dynamoInserter = new DynamoInserter(versionedHybridStore)
```
alicefuzier: It seems like you could use the DynamoInserter fixture a bit everywhere in this test suite

alexwlchan: I was on the fence about it – we need the VHS in the rest of these tests, and you either create it and pass it into a fixture (which is barely any different from the current code), or pass it out in a 2-tuple, which I’m not a fan of.

Issue #1408, opened 2018-01-12
Implement rules for merging single-page Miro and Sierra works


jtweed:
When matching Miro and Sierra records for merging, we should base the decision only on data from Sierra. This is because the Sierra data is canonical and can be updated, whereas the Miro data cannot.

The rules are as follows:

* Sierra work of type `Picture` or `Digital images`
* Sierra has a Miro ID that can be extracted from 962
* There is only one Miro ID in 962 (there can be multiple, if so do not merge any of them)

Where there a decision to merge, the Sierra work should be used as the base work, with the following data copied from Miro:

* the digital location from the Miro item added to the Sierra item
* the source identifiers (excluding `sierra-system-number`) from the Miro work added to the Sierra work
* the source identifiers (excluding `sierra-system-number`) from the Miro item added to the Sierra item

No other data should be taken from the Miro work.

alexwlchan:
There's an additional rule here:

* Only if the Sierra record has a single item (so we know where to put the location)

PR #2497, opened 2018-08-02
Merger doesn't handle multiple output works correctly

alexwlchan:
Consider the following scenario: four works joined together, which get split into two pairs of physical/digital works when a link breaks:

graph

The merger should send two digital works, two physical works; in fact, it sees an attempt to merge four works together and doesn't do any merging. I've added a test to show the error, and included a comment showing where I think the error appears.

Spotted doing a code read ahead of #1408.


kenoir:
Looks fine to me, maybe worth @jtweed eyes on the merging tests?

PR #2500, opened 2018-08-02
Break up MergerWorkerService to add a PlaybackService

alexwlchan:
A possible factor in #2497 is that we’ve got quite a lot of private methods in the worker service, which are hard to test individually. This PR pulls out a “PlaybackService” (for fetching recorded works), which can be tested separately.

It’s not what’s at fault in #2497, but lays the ground for the fix in another PR.

### DIFF
--- a/catalogue_pipeline/merger/src/test/scala/uk/ac/wellcome/platform/merger/MergerFeatureTest.scala
+++ b/catalogue_pipeline/merger/src/test/scala/uk/ac/wellcome/platform/merger/MergerFeatureTest.scala
@@ -11,8 +11,6 @@ import uk.ac.wellcome.storage.vhs.EmptyMetadata
 import uk.ac.wellcome.test.utils.ExtendedPatience
 import uk.ac.wellcome.json.JsonUtil._
 
-import scala.concurrent.ExecutionContext.Implicits.global
-
 class MergerFeatureTest
     extends FunSpec
     with Messaging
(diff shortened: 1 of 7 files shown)


catalogue_pipeline/merger/src/test/scala/uk/ac/wellcome/platform/merger/MergerFeatureTest.scala
```diff
@@ -41,17 +41,17 @@ class MergerFeatureTest
                     val recorderWorkEntry =
                       createRecorderWorkEntryWith(version = 1)
 
-                    whenReady(storeInVHS(vhs, recorderWorkEntry)) { _ =>
-                      val matcherResult =
-                        matcherResultWith(Set(Set(recorderWorkEntry)))
-                      sendNotificationToSQS(queue, matcherResult)
+                    storeInVHS(vhs, recorderWorkEntry)
```
alicefuzier: Why removing whenReady? The old logic seemed to call sendNotificationToSQS only when storeInVHS completed successfully. Does it matter? Could this change cause flakyness?

catalogue_pipeline/merger/src/test/scala/uk/ac/wellcome/platform/merger/MergerTestUtils.scala
```diff
@@ -30,19 +33,25 @@ trait MergerTestUtils extends WorksUtil { this: SQS with SNS with Messaging =>
   def storeInVHS(vhs: VersionedHybridStore[RecorderWorkEntry,
                                            EmptyMetadata,
                                            ObjectStore[RecorderWorkEntry]],
-                 recorderWorkEntry: RecorderWorkEntry): Future[Unit] =
+                 recorderWorkEntry: RecorderWorkEntry): Assertion = {
     vhs.updateRecord(recorderWorkEntry.id)(
       ifNotExisting = (recorderWorkEntry, EmptyMetadata()))((_, _) =>
       throw new RuntimeException("Not possible, VHS is empty!"))
 
+    eventually {
```
alicefuzier: Oh I see, ignore me

catalogue_pipeline/merger/src/test/scala/uk/ac/wellcome/platform/merger/MergerFeatureTest.scala
alexwlchan: The opposite. I didn’t realise I needed the whenReady() and so my test was flaky. I decided to make this function blocking to match our other fixture methods.

Issue #2479, opened 2018-07-30
Consistent rules for API includes


jtweed:
Right now it’s a bit unclear as to when something should be an include in the Catalogue API. I’d like to propose a consistent set of rules:

- Single value fields on a resource are always returned
- Includes are always used for lists
- Includes are always used for related resources

This is on the assumption that lists may be things we are unable to store in the same document or be very large. Related resources may be identified resources of the same type and so will almost certainly be stored separately.

That means that `/works` should currently have the following includes:

- identifiers
- contributors
- subjects
- genres
- items
- production

With everything else returned by default.

I also think we should rename the parameter from `includes` to `include`, as it sounds more like an action.

What do people think?

alexwlchan:
Sounds good to me. Plumbing identifiers through everything will be fun, but not a reason not to!

alexwlchan:
Thinking some more, I lean towards subjects and genres being included by default, and not the rest.

It's not a failure, but it's confusing when the API returns results and you don't see your search terms in the response. I think subjects/genres are going to be especially prone to that (for short words).

No hard evidence, just a gut feeling.

alicefuzier:
> That means that /works should currently have the following includes:
- identifiers
- contributors
- subjects
- genres
- items
- production

So thumbnail is included by deafult?

alicefuzier:
TODO:

- [x] Rename includes to include in v2 api (#2502 )
- [ ] Include thumbnail by default in v2 responses
- [ ] subjects
- [ ] genres
- [ ] contributors 
- [ ] production

jamesgorrie:
Does this mean we will be returning null values for single value fields with no value?

jtweed:
I don't think so, I think they'll remain as just not in the output.

jamesgorrie:
Ah, okay, it was because of the word `always` in 

> Single value fields on a resource are always returned

PR #2502, opened 2018-08-03
Includes -> include

alicefuzier:
Part of #2479 
👔 ⚡️

### DIFF
--- a/catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/V2WorksController.scala
+++ b/catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/V2WorksController.scala
@@ -5,6 +5,7 @@ import uk.ac.wellcome.display.models.ApiVersions
 import uk.ac.wellcome.display.models.v2.DisplayWorkV2
 import uk.ac.wellcome.elasticsearch.ElasticConfig
 import uk.ac.wellcome.platform.api.models.ApiConfig
+import uk.ac.wellcome.platform.api.requests.{V2MultipleResultsRequest, V2SingleWorkRequest}
 import uk.ac.wellcome.platform.api.services.WorksService
 
 import scala.concurrent.ExecutionContext
@@ -14,12 +15,13 @@ class V2WorksController @Inject()(
   apiConfig: ApiConfig,
   elasticConfig: ElasticConfig,
   worksService: WorksService)(implicit ec: ExecutionContext)
-    extends WorksController(
+    extends WorksController[V2MultipleResultsRequest, V2SingleWorkRequest](
       apiConfig = apiConfig,
       indexName = elasticConfig.indexV2name,
       worksService = worksService
     ) {
   implicit protected val swagger = ApiV2Swagger
+  lazy override protected val includeParameterName: String = "include"
 
   prefix(s"${apiConfig.pathPrefix}/${ApiVersions.v2.toString}") {
     setupResultListEndpoint(ApiVersions.v2, "/works", DisplayWorkV2.apply)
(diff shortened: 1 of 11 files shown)


alexwlchan:
I'm reviewing this.

catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
```diff
@@ -25,15 +25,17 @@ import scala.collection.JavaConverters._
 import scala.concurrent.ExecutionContext
 import scala.reflect.runtime.universe.TypeTag
 
-abstract class WorksController(
+abstract class WorksController[M <: MultipleResultsRequest,
+                               S <: SingleWorkRequest](
```
alexwlchan: 👍 

Nice choice of parameter names.

catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/V2WorksController.scala
```diff
@@ -14,12 +18,13 @@ class V2WorksController @Inject()(
...
   elasticConfig: ElasticConfig,
   worksService: WorksService)(implicit ec: ExecutionContext)
-    extends WorksController(
+    extends WorksController[V2MultipleResultsRequest, V2SingleWorkRequest](
       apiConfig = apiConfig,
       indexName = elasticConfig.indexV2name,
       worksService = worksService
     ) {
   implicit protected val swagger = ApiV2Swagger
+  lazy override protected val includeParameterName: String = "include"
```
alexwlchan: minor: Do you need the `override`? Since `WorksController` doesn't define a value, you're not overriding anything.

catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/ApiSwaggerTest.scala
```diff
@@ -47,6 +47,32 @@ class ApiSwaggerTest extends FunSpec with Matchers with fixtures.Server {
...
+
+  it("shows the include request parameter model for v2") {
+    val tree = readTree(s"/catalogue/${ApiVersions.v2.toString}/swagger.json")
+    tree.at("/definitions/Work").isObject shouldBe true
+    val parameters = tree
+      .at("/paths")
+      .findPath(s"/catalogue/${ApiVersions.v2.toString}/works")
+      .at("/get/parameters")
+      .asScala
+      .map(_.findPath("name").asText())
```
alexwlchan: minor: Pulling this into a helper method would let you reduce repetition and make these tests clearer.

catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/V2WorksController.scala
alicefuzier: In this case probably not, because as you say, WorksController doesn't give it a value so it wouldn't compile anyway. I tend as a habit to put overrides a bit everywhere though, cause it makes compilation fail in cases where there is a default value in the base class, typos etc. It makes a bit more robust in case of further modifications and refactorings I think.

catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/ApiSwaggerTest.scala
alicefuzier: Yeah, I did think about that, but it's only copied twice and probably it's not going to be used again so ¯\_(ツ)_/¯

PR #2503, opened 2018-08-03
Continue breaking up the merger: introducing the Merger Manager

alexwlchan:
Another piece of the MergerWorkerService spun out into its own object doing in-memory operations, and with a battery of extra tests just because.


catalogue_pipeline/merger/src/main/scala/uk/ac/wellcome/platform/merger/services/MergerManager.scala
```diff
@@ -0,0 +1,28 @@
+package uk.ac.wellcome.platform.merger.services
+
+import uk.ac.wellcome.models.recorder.internal.RecorderWorkEntry
+import uk.ac.wellcome.models.work.internal.{BaseWork, UnidentifiedWork}
+
+object MergerManager {
```
alicefuzier: Why making it an object instead of a class? If it was a class you could pass the merger in the constructor and not have that process function. It seems to me the usage of this object is a bit weird and forces users of this class to know more than they should know

alexwlchan: Sure, I can make it a class as well.

PR #2507, opened 2018-08-06
should or should not, there is no [[Scala.util.Try]]

alexwlchan:
Clean up our test descriptions – we've been doing it piecemeal, this solves it in one fell swoop.


sbt_common/display/src/test/scala/uk/ac/wellcome/display/models/v2/DisplayLicenseV2Test.scala
```diff
@@ -5,7 +5,7 @@ import uk.ac.wellcome.models.work.internal.License_CCBY
 
 class DisplayLicenseV2Test extends FunSpec with Matchers {
 
-  it("should read a License as a DisplayLicenseV2 correctly") {
+  it("reads a License as a DisplayLicenseV2 correctly") {
     val displayLicense = DisplayLicenseV2(License_CCBY)
```
mklander: `should` and `correctly` 😄 
s/correctly//

PR #2508, opened 2018-08-06
V2 thumbnail included by default

alicefuzier:
Part of #2479 
🗡 🏷


alexwlchan:
Ah, merge conflicts, which would explain what appear to be things that scalafmt missed.

PR #2510, opened 2018-08-06
Don't use the "Sierra" namespace in the Sierra data VHS

alexwlchan:
The new Sierra-only VHS is prefixing every ID with "sierra/", which doesn't add anything except noise. Let's get rid of it.


alexwlchan:
Replaced by #2521.

PR #2514, opened 2018-08-07
And try going in the other direction

alexwlchan:
The counterpart to #2514.


alexwlchan:
Failure in the merger looks like an at-least-once delivery issue.

alexwlchan:
Note: there's a fix for the merger issue in #2503.

PR #2516, opened 2018-08-07
 Add a digital item if we have location 'dlnk'

### LINKED ISSUE #2515: Expose IIIF presentation URLs in the API for all digitised records
Currently we only expose a manifest URL on records with a material type of "E-Book".

As @gestchild has just discovered, there are bib records which have been digitised, and which appear in the online viewer, but we don't expose a location in the catalogue API. For example:

https://wellcomelibrary.org/item/b15701712
https://wellcomelibrary.org/iiif/b15701712/manifest
https://api.wellcomecollection.org/catalogue/v2/works/nzn2vaju

Inspecting the JSON for the corresponding bib record, I found the following:

```json
      "locations": [
        {
          "code": "digi",
          "name": "Digitised Collections"
        },
        {
          "code": "dlnk",
          "name": "Digitised content"
        },
        {
          "code": "elro",
          "name": "Online"
        },
```

Talking to Natalie and Branwen, these codes work as follows:

* "digi" appears very early – it's just a note that a work is a candidate for digitisation. It doesn’t mean that it’s been digitised yet, or that it’s still going to be digitised (sometimes things go out of scope).
* "dlnk" is added when there's something in the online viewer ready for display. Some records without a "dlnk" also have an online viewer page, but that would be unusual.
* "elro" means we're linking to some other online resource on a different site.

For our purposes, the "dlnk" location seems like the best distinction for deciding if something needs a digital location (if it's not an ebook).

alexwlchan:
Resolves #2515.

### DIFF
--- a/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraItems.scala
+++ b/catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraItems.scala
@@ -90,11 +90,19 @@ trait SierraItems extends Logging with SierraLocation {
   def getDigitalItems(
     sourceIdentifier: SourceIdentifier,
     sierraBibData: SierraBibData): List[Unidentifiable[Item]] = {
-    sierraBibData.materialType match {
-      case Some(SierraMaterialType("v", "E-books")) =>
-        List(getDigitalItem(sourceIdentifier))
-      case _ => List.empty
+
+    val hasEbookMaterialType =
+      sierraBibData.materialType.contains(SierraMaterialType("v", "E-books"))
+
+    val hasDlnkLocation = sierraBibData.locations match {
+      case Some(locations) => locations.map { _.code }.contains("dlnk")
+      case None => false
+    }
+
+    if (hasEbookMaterialType || hasDlnkLocation) {
+      List(getDigitalItem(sourceIdentifier))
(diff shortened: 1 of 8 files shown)


alexwlchan:
Part of #2379.

catalogue_pipeline/transformer/src/main/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/SierraItems.scala
```diff
@@ -75,14 +75,34 @@ trait SierraItems extends Logging with SierraLocation {
...
-        List(getDigitalItem(sourceIdentifier))
-      case _ => List.empty
+
+    val hasEbookMaterialType =
+      sierraBibData.materialType.contains(SierraMaterialType("v", "E-books"))
+
+    val hasDlnkLocation = sierraBibData.locations match {
+      case Some(locations) => locations.map { _.code }.contains("dlnk")
+      case None => false
     }
```
mklander: Shorter, but I'm not sure if this is as clear:
`val hasDlnkLocation = sierraBibData.locations.exists(_.exists(location => location.code == "dlnk"))`

alexwlchan: Mmm. I don’t think either is clearly better, so I’ll stick with this version so I can get it into Travis.

PR #2528, opened 2018-08-09
Only add a digital location if the bib record has a location 'dlnk'

alexwlchan:
This is a fix for cases where we're showing a IIIF Presentation manifest URL that doesn't resolve, for example:

https://api.wellcomecollection.org/catalogue/v2/works?query=2436758&includes=items

Talking to Branwen, the only good distinction of whether a manifest exist is whether we've digitised it – which is the 'dlnk' code in locations. This patch updates the transformer to match, but I wonder if this is going to affect the merging? @jtweed @mklander


alexwlchan:
Turns out the merger will do the right thing, hooray! So rebasing and then merging.

PR #2530, opened 2018-08-09
Puts API gateway lambdas into their own module

kenoir:

In preparation for adding another api gw lambda, modularise the infra.

### Who is this change for?

Developers wanting to write less code.


kenoir:
@fractos worth a look as it builds off your work.

kenoir:
@fractos i've addressed review comments. Could you take another look?

archive/terraform/api_gw_lambda/api_gateway.tf
```diff
@@ -1,47 +1,46 @@
...
   resource_id = "${aws_api_gateway_method.proxy_root.resource_id}"
   http_method = "${aws_api_gateway_method.proxy_root.http_method}"
 
   integration_http_method = "POST"
   type                    = "AWS_PROXY"
-  uri                     = "${module.lambda_archive_asset_lookup.invoke_arn}"
+  uri                     = "${module.lambda.invoke_arn}"
 }
 
 resource "aws_api_gateway_deployment" "test" {
```
fractos: might want to have a think about whether provide multiple deployments as an option.

archive/terraform/api_gw_lambda/api_gateway.tf
```diff
@@ -50,7 +49,7 @@ resource "aws_api_gateway_deployment" "test" {
     "aws_api_gateway_integration.lambda_root",
   ]
 
-  rest_api_id = "${aws_api_gateway_rest_api.archive_asset_lookup.id}"
+  rest_api_id = "${aws_api_gateway_rest_api.api.id}"
   stage_name  = "prod"
 
   # keep this bit around as might require it for terraform to update the deployment
```
fractos: A note about this - without the commented code, any changes to the API gateway setup won't get deployed, so they'd have to be manually actioned in the UI or via script. The side-effect of the commented code is that every time the Terraform is applied then a deployment to API gateway will occur. This was more of an issue when I was developing a basic HTTP auth custom authorizer and had to manually deploy it after each change.

archive/terraform/api_gw_lambda/api_gateway.tf
kenoir: Quite possibly, but later!

archive/terraform/api_gw_lambda/api_gateway.tf
kenoir: Eep. That's not great, but doesn't seem like a regression.

Issue #2495, opened 2018-08-01
Prototype Archive ingest/archive/retrieval


kenoir:
- [x] Storage API - Ingest #2531 
- [x] Verification #2434 
- [x] Registration of storage manifest #2532
- [x] Storage API - Get manifest #2475

PR #2532, opened 2018-08-10
Archive registrar skeleton

kenoir:

This PR adds a skeleton for the registrar (part of the archive storage system).

There is much code duplication between this and the archiver app, as this is the second finatra-less application and follows the same patterns.

This should be factored out after merge of both apps.

### Who is this change for?

📖 ❤️ 🏬 

Depends #2434


kenoir:

Issue #1494, opened 2018-01-31
Fix nextPage/prevPage links in the staging API


alexwlchan:
Currently they're hard-coded to point to api.wellcomecollection.org; might be nice if they could point to api-stage.

(But how to do this so the switchover is seamless?)

alexwlchan:
At some point this got fixed, and I didn't notice. I think this means it's temporarily wrong when we flip the APIs though, unless I've missed something?

alexwlchan:
Closed by #2548.

PR #2548, opened 2018-08-13
Flip the prod API

alexwlchan:
And the usual round of bugs I found in the API flipping process when I did it. Notes inline.

This change is deployed.


alexwlchan:
Closes #1494.

PR #2553, opened 2018-08-14
Don't namespace the RDS security group

alexwlchan:
There's no reason for the security group to be namespaced, or created by the pipeline module.

This change will make it easier to separate RDS from this stack, as part of #2533.


alexwlchan:
This change is applied.

catalogue_pipeline/terraform/security_groups.tf
```diff
@@ -1,7 +1,7 @@
 resource "aws_security_group" "service_egress_security_group" {
-  name        = "${var.namespace}_service_egress_security_group"
+  name        = "pipeline_service_egress_security_group"
```
alicefuzier: Why taking this sg out of the pipeline module? I don't see it ever being used outside

alexwlchan: Oh yeah, I hadn't spotted we only need half of this file. I’ll roll it back.

PR #2557, opened 2018-08-15
Migrate RDS database into a separate stack

### LINKED ISSUE #2533: Move the SourceData VHS cassettes and the RDS database out of the catalogue_pipeline stack

alexwlchan:
Resolves #2533.

I, err, sorta deleted the RDS instances by accident when doing this. I think I restored correctly from the write replica, but guess we’ll find out? 🙈 

This is now applied. There's a bug in the data stack where it wants to make the first instance not public-facing, even though it already is, and it always comes up on the terraform plan. Otherwise fine.


alexwlchan:
I’m going to merge this as-is before anything else changes underneath it or breaks

PR #2559, opened 2018-08-15
Remove some Jackson-specific logic from DisplayItem

alexwlchan:
We no longer parse documents from Elasticsearch using Jackson, only Circe -- so this workaround for a Jackson bug can be removed.

Another bit of refactoring for #2163.


alexwlchan:
(I did grep for “Jackson”, and these are the only two comments that mention it)

PR #2567, opened 2018-08-16
Archive ingest api

mklander:

Create the archive-ingest API endpoint lambda API-GW -> ingest-lambda -> archivist-SNS


archive/archive_ingest/src/archive_ingest.py
```diff
@@ -0,0 +1,68 @@
+# -*- encoding: utf-8 -*-
+"""
+Receives a message to ingest a bag giving the URL and publishes the archive event to an SNS topic.
+"""
+
+import os
+
+import boto3
+import daiquiri
+from urllib.parse import urlparse
```
alexwlchan: style: convention is that Python imports go in three groups – builtins, third-party modules, first-party code. Since `urllib` is a standard library module, better would be:

```python
import os
from urllib.parse import urlparse

import boto3
import daiquiri
```

archive/archive_ingest/src/archive_ingest.py
```diff
@@ -0,0 +1,68 @@
...
+            'namespace': bucket,
+            'key': key
+        }
+    else:
+        raise ValueError(f"Unrecognised url scheme: {bag_url}")
+
+
+@log_on_error
+def handler(event, _ctx=None, sns_client=None):
+    logger.info(f"received {event}")
```
alexwlchan: We don't log the received event in our other Lambdas (except perhaps at debug) unless there's an error – that’s what the `@log_on_error` does for us. This keeps logging costs down.

Also, good practice with `logger` is to use `%`-formatters, and pass them as extra arguments to the log method, i.e.:

```python
logger.info('received %r', event)
```

If the log message isn't used (because there are no handlers, or the level is too low), you save the cost of serialising the object as a string. Probably not a big issue here, but a good habit to get into.

archive/archive_ingest/src/archive_ingest.py
```diff
@@ -0,0 +1,68 @@
...
+daiquiri.setup(level=os.environ.get('LOG_LEVEL', 'INFO'))
+logger = daiquiri.getLogger()
+
+
+def ensure_valid_request(event):
+    required_key = 'bagURL'
+    if type(event) is not dict:
+        raise ValueError(f"Invalid request not json dict: {event}")
+    if required_key not in event:
+        raise ValueError(f"Invalid request missing {required_key}")
```
alexwlchan: Python idiom is Easier to Ask For Permission, not Look Before You Leap.

It would be better to assume this is an object that implements square-bracket lookup (the `__getitem__` method), and has a key called `bagURL`, and catch the error, if any.

i.e.

```python
try:
    bag_url = event['bagURL']
except KeyError:
    raise ValueError('custom error goes here')
```

although KeyError is probably the most correct error to raise here.

Other notes:

* In general, prefer `isinstance(obj, expected_type)` to `type(obj) is expected_type` – the former allows for subclassing, the latter doesn't
* If the required key changes in `main()`, this function could pass if you forget to update `required_key`. Better to pass it in as a parameter, or catch the error upon lookup.

archive/archive_ingest/src/archive_ingest.py
```diff
@@ -0,0 +1,68 @@
...
+    logger.info(f"received {event}")
+
+    ensure_valid_request(event)
+
+    bag_url = event['bagURL']
+
+    message = archive_bag_message(bag_url)
+    logger.debug(f"sns-message: {message}")
+
+    topic_arn = os.environ['INGEST_TOPIC_ARN']
```
alexwlchan: nit: I prefer to extract environment variables at the top of a function, so if there's a config error we fail immediately.

archive/archive_ingest/src/test_archive_ingest.py
```diff
@@ -0,0 +1,48 @@
+# -*- encoding: utf-8 -*-
+
+import os
+import archive_ingest
+import pytest
+
+
+def test_sends_location_to_sns(sns_client, topic_arn):
+    os.environ['INGEST_TOPIC_ARN'] = topic_arn
```
alexwlchan: minor: prefer `os.environ.update({'INGEST_TOPIC_ARN': topic_arn})`, as that doesn't blat existing environment config.

Also, the `topic_arn` fixture auto-populates the `TOPIC_ARN` environment variable, if you wanted to be consistent with our other Lambdas.

archive/archive_ingest/src/test_archive_ingest.py
```diff
@@ -0,0 +1,48 @@
...
+        'key': 'test-bag.zip'
+    }
+
+
+def test_invalid_url_fails(sns_client, topic_arn):
+    with pytest.raises(ValueError, match="Unrecognised url scheme: invalid"):
+        os.environ['INGEST_TOPIC_ARN'] = topic_arn
+        request = {'bagURL': 'invalid'}
+
+        archive_ingest.handler(event=request, sns_client=sns_client)
```
alexwlchan: minor: I'd move the first two lines out of the `with` block – leave as little code as possible to fail inside the pytest capture.

Nice find with the `match` parameter, I haven't seen that before.

archive/archivist/trigger_archive_bag.py
```diff
@@ -3,23 +3,28 @@
 """
 Create a request to archive a bag
 
-Usage: trigger_archive_bag.py <BAG>... [--bucket=<BUCKET_NAME>]  [--topic=<TOPIC_NAME>]
+Usage: trigger_archive_bag.py [sns] <BAG>... [--bucket=<BUCKET_NAME>]  [--topic=<TOPIC_NAME>|--api=<API>]
```
alexwlchan: minor: I think a flag `--sns=(true|false)` would be more idiomatic, but that's not a strong feeling.

archive/archivist/trigger_archive_bag.py
```diff
@@ -3,23 +3,28 @@
...
        trigger_archive_bag.py -h | --help
 
 Options:
+  sns                    Send directly to SNS rather than through the API
+                         [default: false]
   --bucket=<BUCKET_NAME> The S3 bucket containing the bags.
                          [default: wellcomecollection-assets-archive-ingest]
   --topic=<TOPIC_NAME>   The archivist topic.
                          [default: archive-storage_archivist]
+   --api=<API>           The API enndpoint to use
```
alexwlchan: sp: "enndpoint"

archive/archivist/trigger_archive_bag.py
```diff
@@ -57,23 +72,36 @@ def publish_messages(topic_arn, messages):
...
+            print(f"Error: [{response.status_code}] {response.json()}")
+
+
+def main():
+    args = docopt.docopt(__doc__)
+    bags = args['<BAG>']
+    bucket_name = args['--bucket']
+    use_sns_directly = args['sns']
+
+    if(use_sns_directly):
```
alexwlchan: nit: lose the parentheses

archive/archivist/trigger_archive_bag.py
```diff
@@ -57,23 +72,36 @@ def publish_messages(topic_arn, messages):
...
-        messages=messages
+        messages=archive_bag_sns_messages(bags, bucket_name)
     )
 
 
+def call_ingest_api(bucket_name, bags, api):
+    for message in archive_bag_api_messages(bags, bucket_name):
+        response = requests.post(api, json=message)
+        if response.status_code != 200:
+            print(f"Error: [{response.status_code}] {response.json()}")
```
alexwlchan: Is continuing silently the correct response here? Or would it be better to error explicitly?

Using a requests Session might improve performance slightly – you get a shared connection pool. Similar API:

```python
sess = requests.Session()
for message in archive_bag_api_messages(bags, bucket_name):
    response = sess.post(api, json=message)
```

mklander:
Thanks @alexwlchan

PR #2571, opened 2018-08-20
Simplify label for person subjects

alicefuzier:


catalogue_pipeline/transformer/src/test/scala/uk/ac/wellcome/platform/transformer/transformers/sierra/subjects/SierraPersonSubjectsTest.scala
```diff
@@ -41,7 +41,7 @@ class SierraPersonSubjectsTest
 
     transformer.getSubjectsWithPerson(sierraBibData) shouldBe List(
       Subject(
-        label = "Sir David Attenborough",
+        label = "David Attenborough Sir",
```
alexwlchan: As discussed – this looks odd but doesn't reflect MARC, so let's make it look better!

Issue #2572, opened 2018-08-20
Extra items on a single work


alexwlchan:
Alice spotted this: https://api.wellcomecollection.org/catalogue/v2/works/prrf9sds?include=items,identifiers

Since there's only one item, we should add the location from the digital item onto the physical item, not create a new item for it. Or something – we should clarify exactly what we need to do here.

PR #2603, opened 2018-08-24
Clean up the interface to SierraItems

alexwlchan:
The responsibility for SierraItems was a bit all over the place:

* It was doing JSON decoding of the item data, whereas SierraTransformableTransformer handles the bib JSON
* It had a mishmash of public methods
* SierraTransformableTransformer was the thing that actually constructs the list of items on a work, in contrast to everything else – internal logic was leaking

So as a refactoring before #2572, this refactors the trait to present a single public method – `getItems()` – which callers can use to get a list of all items on this work.

### DIFF
--- a/catalogue_pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/sierra/SierraItems.scala
+++ b/catalogue_pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/sierra/SierraItems.scala
@@ -10,9 +10,9 @@ import uk.ac.wellcome.platform.transformer.sierra.source.{
 
 trait SierraItems extends Logging with SierraLocation {
   def getItems(
+    bibId: SierraBibNumber,
     bibData: SierraBibData,
-    itemDataMap: Map[SierraItemNumber, SierraItemData],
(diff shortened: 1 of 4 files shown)


alexwlchan:
Turns out we didn't have a test for physical location on items, so I've added on there as well.

catalogue_pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/sierra/SierraItems.scala
```diff
@@ -1,35 +1,30 @@
...
+    : List[MaybeDisplayable[Item]] = {
+    val physicalItems = getPhysicalItems(itemDataMap)
+    val digitalItems = List(getDigitalItem(bibId = bibId, bibData = bibData)).flatten
 
-  def transformItemData(itemId: SierraItemNumber,
-                        itemData: SierraItemData): Identifiable[Item] = {
+    physicalItems ++ digitalItems
+  }
+
+  private def transformItemData(
```
alicefuzier: where is this used?

alexwlchan: In getPhysicalItems

PR #2606, opened 2018-08-28
How do you solve a problem like the reindexer?

alexwlchan:
This patch started as me trying to run the reindexer against the Miro table, and grew.

* ~Run one reindexer per source table~ (see #2610)
* Use parallel scan (https://alexwlchan.net/2018/08/parallel-scan-scanamo/) in the reindexer, rather than the shard generator.

  The Miro table doesn't have the reindexVersion field, which breaks the reindexer. Switching to parallel scan was easier than adding it. (This opens the door to removing the reindex_shard_generator later.)
* ~Plug the reindexer output topics into the transformer~ (#2610)
* ~Update the IAM permissions for the Miro transformer to point at the dedicated Miro VHS~ (#2610)


kenoir:
So a combination of parallel scan using some measure to divide up a table of arbitrary size and presuming SQS will keep any incomplete work is ensuring we can keep track of which records have been processed?

alexwlchan:
Exactly.

I’ve run one complete reindex of both Miro and Sierra with this PR applied, and it seemed to work correctly. I was breaking it up into smaller PRs though, because there’s a bunch of unrelated refactoring in there.

Issue #2573, opened 2018-08-20
Ditch dynamo streams


alicefuzier:
Send messages to SNS instead of relying on dynamo streams. This has a bunch of advantages:
- dynamo stream is a bottleneck that we can't control as the parallelism is set somewhat mysteriously by aws. This should give us back control of speed and also visibility of where messages are in the pipeline.
- All applications read the same kind of message i.e `MessagePointer` which simplifies the code
- The reindexer doesn't actually need to modify anything in dynamo db which lets us get rid of reindex_request_processor and only keep the reindex_request_creator.
- This in turn lets us get rid of the reindex global secondary index on the tables

TODO:

- [x] Modify VHS to store full S3 location (bucket and key) (https://github.com/wellcometrust/scala-storage/pull/27)
- [x] Modify recorder to send message pointers with the full location in S3 (#2574)
- [x] Modify the sierra adapter to use the new VHS (#2584)
- [x] Do a full reharvest of the Sierra data (in progress)
- [x] Modify the transformer to read a MessageStream (#2595 )
- [x] Modify the reindexer
- [x] Modify the items_to_dynamo to use the new VHS  (#2584)
- [x] Migrate data from the Miro VHS to the new HybridRecord format 
- [ ] Get rid of all the DynamoDB streams config

alexwlchan:
Bring this back as the reindexer doesn't appear to be working (I think it's just a terraform issue).

kenoir:
👏 great work on this.

PR #2610, opened 2018-08-29
Run one reindexer per source table, and subscribe the transformers to their output

alexwlchan:
Spun out of #2606 as a standalone patch.

Required for #2573.


alexwlchan:
> Can we modularise these tables in the TF?

How so? They’re already coming out of the VHS module.

alexwlchan:
I’m going to merge this so I can start tidying up #2606, but happy to do further modularising in a further PR.

PR #2618, opened 2018-08-30
API for bags endpoints

mklander:

Creates an API for bags endpoints returning manifests


archive/archive_bags/src/archive_bags.py
```diff
@@ -0,0 +1,39 @@
...
+BUCKET_NAME = 'wellcomecollection-vhs-archive-manifests'
+
+daiquiri.setup(level=os.environ.get('LOG_LEVEL', 'INFO'))
+logger = daiquiri.getLogger()
+
+
+@log_on_error
+def main(event, _, dynamodb_client=None, s3_client=None):
+    dynamodb_client = dynamodb_client or boto3.client('dynamodb')
+    s3_client = s3_client or boto3.client('s3')
```
mklander: I noticed now that `dynamo_resource` is an option, and we should move to that in another PR.

archive/archive_bags/src/test_archive_bags.py
```diff
@@ -0,0 +1,92 @@
...
+    request = {'id': manifest_id}
+
+    response = archive_bags.handler(
+        request,
+        None,
+        dynamodb_client,
+        s3_client
+    )
+
+    assert response == stored_manifest
```
mklander: Should add some test for failure cases in another PR.

PR #2620, opened 2018-08-30
Add a Miro ID parser for Wellcome Images URLs

alexwlchan:
A key part of #1408 is being able to extract Miro IDs from Wellcome Images URLs as they've been stored in the Sierra catalogue.

I went through the entire VHS, and extracted every instance of MARC tag 962. It looks like we want subfield $u, so I built a series of regexes that should cover all the Sierra data. This turns out to be moderately fiddly, so I put it in its own trait for testing, and it's a standalone component we can merge immediately.


alexwlchan:
> Any way of matching extracted ids against actual ids for the whole dataset?

How do you know what the result should be? :-)

If it helps, this is the direct transcription of some regexes written in a Jupyter notebook which have run against the entire set.

PR #2624, opened 2018-08-31
[WIP] Ensure tagmanifest is included in archived bags

kenoir:

The archive system currently does not transfer tagmanifest files as they are not themselves described in a manifest including a checksum.

However tagmanifests are _implicitly_  correct if all other files are verified, therefore we can safely transfer them.

In order to move files without checksums the archivist service needs to distinguish these checksum free files and transfer them where appropriate.

### Who is this change for?

Folk who want a complete Bagit bag archived.


kenoir:
Closing as needs rethink.

Issue #2366, opened 2018-07-06
Implement ingest interface


jtweed:
- [x] POST to /ingest
- [x] GET /ingest/:id
- [x] Write ingest events
- [x] Callback on success
~- [ ] Callback on failure~ ** Doing this after conversation with @jtweed .

kenoir:
Done for now.

PR #2626, opened 2018-08-31
Track archive ingest progress

mklander:
Track archive progress during ingest.  Available at GET /ingest/:id

part of #2366


archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/progress/monitor/ArchiveProgressMonitor.scala
```diff
@@ -25,36 +28,59 @@ class ArchiveProgressMonitor @Inject()(dynamoDbClient: AmazonDynamoDB,
...
+      case Right(a) =>
+        debug(s"Successfully updated Dynamo record: ${progress.id} $a")
     }
+    progress
   }
 
-  def addEvent(id: String, description: String) = {
-    val progressTable = Table[ArchiveIngestProgress](dynamoConfig.table)
-
+  def addEvent(id: String,
```
kenoir: Adds event and status. Needs renaming?

PR #2629, opened 2018-09-05
Assorted fixes for the latest reindex

alexwlchan:
Not all our services support VPC endpoints, but it's an improvement. I’m starting a reindex to see if this makes a difference.

### DIFF
--- a/terraform-modules/egress_security_group/main.tf
+++ b/terraform-modules/egress_security_group/main.tf
@@ -1,0 +1,60 @@
+resource "aws_security_group" "security_group" {
+  name        = "${var.name}"
+  description = "${var.description}"
+  vpc_id      = "${var.vpc_id}"
+
+  egress {
+    from_port   = 0
+    to_port     = 0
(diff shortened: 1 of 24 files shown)


alexwlchan:
I’ve tried to put in the route table associations; I couldn't work out how to do them for gateway endpoints.

terraform-modules/egress_security_group/main.tf
```diff
@@ -0,0 +1,50 @@
...
+
+  tags {
+    Name = "${var.name}"
+  }
+}
+
+data "aws_vpc_endpoint_service" "logs" {
+  service = "logs"
+}
+
```
mklander: Is this is the right file, `egress_security_group`?

alexwlchan: I’m not convinced it’s the best file long-term, but it was where I put it to test with the catalogue pipeline. (It can't go with the VPC because you need the security group IDs.)

PR #2633, opened 2018-09-05
If the date string in the RFC is wrong, update it automatically

### LINKED ISSUE #2108: [WIBNI] RFC linting task corrected the "Last updated" date in the RFC header

alexwlchan:
Resolves #2108, and would have fixed Silver's problem just now.


alexwlchan:
Probably, but most of the rest of our build scripts aren't tested, and I don't want to add that infra right now. 🙈

PR #2639, opened 2018-09-06
Only store a message pointer in S3 if the message is too large to send via S3/SNS

alexwlchan:
A non-trivial cost is the number of PutObject calls we make for messaging. Yesterday we spent $65 on this API operation alone.

Since a lot of our messages are much smaller than the 256KB cap on SNS/SQS, wouldn't it be nice if we only sent an S3 pointer when we actually needed to? So… let’s do that.

(Estimated savings are ~$250/month based on back-of-the-napkin calculations yesterday. This took a little over an hour for the initial patch, plus some bugfixing and deployment. Throw in the tax on $250, and I think this is pretty good value.)

### DIFF
--- a/sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/message/MessageStream.scala
+++ b/sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/message/MessageStream.scala
@@ -7,10 +7,11 @@ import com.amazonaws.services.s3.AmazonS3
 import com.amazonaws.services.sqs.AmazonSQSAsync
 import com.amazonaws.services.sqs.model.Message
 import com.google.inject.Inject
+import io.circe.Decoder
 import uk.ac.wellcome.messaging.sns.NotificationMessage
 import uk.ac.wellcome.messaging.sqs.SQSStream
 import uk.ac.wellcome.monitoring.MetricsSender
-import uk.ac.wellcome.storage.{ObjectLocation, ObjectStore}
+import uk.ac.wellcome.storage.ObjectStore
 import uk.ac.wellcome.json.JsonUtil._
 
 import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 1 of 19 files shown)


sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/message/MessageStream.scala
```diff
@@ -32,35 +33,44 @@ class MessageStream[T] @Inject()(actorSystem: ActorSystem,
...
 
-  def foreach(streamName: String, process: T => Future[Unit]): Future[Done] =
+  def foreach(streamName: String, process: T => Future[Unit])(
+    implicit decoder: Decoder[MessageNotification[T]]): Future[Done] =
     sqsStream.foreach(
       streamName = streamName,
       process = (notification: NotificationMessage) =>
         for {
-          obj <- readFromS3(notification.Message)
+          obj <- getBody(notification.Message)
```
mklander: ` obj <- getBody(...` ?

sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/message/MessageWriter.scala
```diff
@@ -39,19 +40,43 @@ class MessageWriter[T] @Inject()(
...
+      //
+      // Max SNS message size:
+      // https://aws.amazon.com/sns/faqs/
+      //
+      notification: MessageNotification[T] <- if (encodedString.length > 250 * 1000) {
+        writeToS3(encodedString)
+      } else {
+        Future.successful(InlineNotification(message))
+      }
+
```
mklander: This relies on the SNSWriter using the same JSON serialization as `encodedString <- Future.fromTry(toJson(message))`.  Would it be possible to build the whole notification as a string before the test and then use the `writeMessage(message: String, subject: String)` method?

This still assumes the underlying byte serialisation used by AmazonSNS, which we could make a bit more explicit `string.getBytes("UTF-8").length`.

sbt_common/messaging/src/test/scala/uk/ac/wellcome/messaging/message/MessageWriterTest.scala
```diff
@@ -22,102 +21,156 @@ class MessageWriterTest
...
-  val subject = "message-writer-test-subject"
+  def createMessage(size: Int) = ExampleObject("a" * size)
 
-  it("sends messages") {
-    withLocalSnsTopic { topic =>
-      withLocalS3Bucket { bucket =>
-        withExampleObjectMessageWriter(bucket, topic) { messageWriter =>
-          val eventualAttempt = messageWriter.write(message, subject)
+  val smallMessage: ExampleObject = createMessage(size = 100)
+  val largeMessage: ExampleObject = createMessage(size = 300000)
```
mklander: Fair enough, also might be interested in a test at the cusp (~250KB).

PR #2641, opened 2018-09-06
Allow testing of Python containers

kenoir:

Do you like Python? Do you like tests? This is the PR for you!

Adds a task template to test python containers!

### Who is this change for?

@tomcrane


functions.Makefile
```diff
@@ -346,6 +346,9 @@ define __ecs_target_template
 $(1)-build:
 	$(call build_image,$(1),$(2))
 
+$(1)-test:
+	$(call test_lambda,$(STACK_ROOT)/$(1))
```
tomcrane: Should this be test_ecs or test_image or something like that?

kenoir: yup, it definitely should :)

Issue #2635, opened 2018-09-06
Run METS transform on all METS files


tomcrane:
The archive bagging component and METS transformer has a flag that allows it to skip the fetch of asset files and only process the METS files. This is much faster to run than the full version, which has to collect files from various places, assemble the bag and then put the bag in S3.

We can process ALL 250,000 METS files relatively quickly with this flag, to identify gaps in the implementation (e.g., unhandled format types and other missing info in https://github.com/wellcometrust/platform/blob/master/archive/bagger/mappings.py)

- [x] Containerise 
- [x] Accept messages for b numbers
- [x] process messages from queue
- [ ] send success messages to success queue
- [x] log failures to a description in an S3 bucket and send message to failure queue

tomcrane:
Problem:

The bagging process needs a lot of scratch disk space per running instance.

The current deployed AWS version has a theoretical 4GB limit.
While each instance cleans up its working directory after each b number...

https://github.com/wellcometrust/platform/blob/6458c617cd0798f2b7190fccc85dcb135cf62a87/archive/bagger/src/bagger.py#L107

... 4GB is nowhere near enough for very large digital objects, or any reasonable-sized video.

So it needs to run with access to a lot more temporary disk space than the current limit. 

It needs to be an ECS task running with access to an EBS volume on an EC2 instance.

Scaling up:
Most of the I/O is to and from S3, so scaling up shouldn't really cause a concern.

The code does query the DLCS for each asset, to determine the origin. A very aggressive scaling up could in theory have an effect here, but it would need to be very, very scaled up.

@fractos

PR #2646, opened 2018-09-07
Bagger in production environment

tomcrane:

The METS bagger running on AWS and listening for jobs
https://github.com/wellcometrust/platform/issues/2635

### Who is this change for?

Platform team, for anyone to be able to run sets of METS through the archive bagging service for testing


archive/terraform/main.tf
```diff
@@ -55,3 +55,53 @@ module "registrar" {
...
+    DROP_BUCKET_NAME_ERRORS     = "${var.bagger_drop_bucket_name_errors}"
+    CURRENT_PRESERVATION_BUCKET = "${var.bagger_current_preservation_bucket}"
+    DLCS_SOURCE_BUCKET          = "${var.bagger_dlcs_source_bucket}"
+    BAGGING_QUEUE               = "${module.bagger_queue.name}"
+
+    # aws creds that can see Preservica as well as Platform estate
+    AWS_ACCESS_KEY_ID     = "${var.bagger_aws_access_key_id}"
+    AWS_SECRET_ACCESS_KEY = "${var.bagger_aws_secret_access_key}"
+    AWS_DEFAULT_REGION    = "${var.aws_region}"
+
```
mklander: Can this be done by a role?

archive/bagger/.gitignore
```diff
@@ -106,6 +106,10 @@ venv.bak/
 *.bat
 bag.sh
 bnums.sh
+local_run.sh
+queue_for_bagging.sh
+local_run_queue_listener.sh
+error_report.sh
 .vscode
```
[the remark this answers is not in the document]
tomcrane: Yes, they are for running various parts of this code locally, they provide the python code with its environment vars.

archive/terraform/main.tf
tomcrane: Yes, we should adapt that bit to provide those vars following the convention of the rest of the platform.

archive/terraform/main.tf
```diff
@@ -55,3 +55,53 @@ module "registrar" {
...
+    DLCS_API_KEY     = "${var.bagger_dlcs_api_key}"
+    DLCS_API_SECRET  = "${var.bagger_dlcs_api_secret}"
+    DLCS_CUSTOMER_ID = "${var.bagger_dlcs_customer_id}"
+    DLCS_SPACE       = "${var.bagger_dlcs_space}"
+
+    # DDS credentials
+    DDS_API_KEY      = "${var.bagger_dds_api_key}"
+    DDS_API_SECRET   = "${var.bagger_dds_api_secret}"
+    DDS_ASSET_PREFIX = "${var.bagger_dds_asset_prefix}"
+  }
```
mklander: Similar here, we should check with @kenoir whether it would be better to manage these values in another way.

kenoir:
@tomcrane can we get this mergeable and add anything else in a different PR?

tomcrane:
It should be mergeable now, still need at address @mklander's concern above, will look at that with @fractos

archive/terraform/main.tf
fractos: @mklander These keys are for an external service (the DLCS), so it won't be possible to carry them by IAM role.

Issue #2632, opened 2018-09-05
September 2018 AWS Costs reduction


kenoir:

alexwlchan:
Related: #2629.

PR #2647, opened 2018-09-07
Reduce persistent logging in this Lambda as well

alexwlchan:
Another tiny logging change spotted while investigating #2632.

Not a big saver but not necessary either.

Already live.


alexwlchan:
Passed Lambda and Python formatting, so skipping the Scala tests and merging by admin superpowers.

PR #2655, opened 2018-09-10
…another Lambda does the rest

alexwlchan:
Follows up #2654, by cleaning up the GET methods from the "start an ingest" Lambda.

Note: now I’ve seen the patterns in use here, I think I see how to get 400/404s etc out of the other Lambda. I think we should pass in those "BadRequest" strings as config – env vars to the Lambda, template vars in the API Gateway Swagger – so they’re always in sync as we add new error codes.


archive/archive_start_ingest/src/archive_start_ingest.py
```diff
@@ -1,6 +1,27 @@
 # -*- encoding: utf-8 -*-
 """
-Receives a message to ingest a bag giving the URL and publishes the archive event to an SNS topic.
+Receives a POST request to ingest a bag giving the URL and publishes the
```
kenoir: 📖 👍

archive/archive_start_ingest/src/archive_start_ingest.py
```diff
@@ -35,39 +54,24 @@ def post_ingest_request(event, sns_client):
     logger.debug('ingest_request_id: %r', ingest_request_id)
 
     message = archive_bag_message(ingest_request_id, upload_url, callback_url)
-    logger.debug(f"sns-message: {message}")
+    logger.debug("sns-message: %r", message)
```
kenoir: y no f-string?

archive/archive_start_ingest/src/test_archive_start_ingest.py
```diff
@@ -5,7 +5,7 @@
 
 import pytest
 
-import archive_start_ingest as archive_ingest
+import archive_start_ingest as start_ingest
```
kenoir: Bit nit-picky, but really this is "request_ingest".

archive/archive_start_ingest/src/archive_start_ingest.py
mklander: premature optimisation ;)
Saves the cost of string building if it's never logged.

mklander:
The [Bad request] error handling is just convention. It keeps the http elements in the API GW and the lambda is a function that can throw, not completely happy with the regexp approach though.

I like the idea of passing them in if we are going to stick to it, but there may be a better way.

PR #2656, opened 2018-09-11
Define API Gateway Lambda errors in configuration, not magic strings

alexwlchan:
Picking up on a comment from #2654:

The API Gateway config detects whether to throw a particular error code based on the presence of certain strings in the error message (e.g. "BadRequest" becomes a 400 error). This patch moves those special strings out into configuration, so they're always consistent – and adds better error-handling elsewhere to boot.


archive/archive_request_ingest/src/test_archive_request_ingest.py
```diff
@@ -56,7 +54,7 @@ def test_sends_request_to_sns_with_callback(sns_client, topic_arn):
 def test_invalid_url_fails(sns_client, topic_arn):
     request = ingest_request('invalidUrl')
 
-    with pytest.raises(ValueError, match="\[BadRequest\] Unrecognised url scheme: invalid"):
+    with pytest.raises(BadRequestError, match="\[BadRequest\] Unrecognised url scheme: 'invalidUrl'"):
         request_ingest.main(event=request, sns_client=sns_client)
```
mklander: Are the []s still needed?  It's a bit confusing because of the regexp, now it's all more consistent perhaps it can just be a prefix?

archive/terraform/archive_api_swagger.json
```diff
@@ -87,12 +87,18 @@
...
             },
-            ".*[BadRequest].*": {
+            ".*[${error_bad_request}].*": {
               "statusCode": "400",
               "responseTemplates": {
                 "application/json": "#set($inputRoot = $input.path('$')){\"message\" : \"$inputRoot.errorMessage\"}"
               }
             },
+            ".*[${error_method_not_allowed}].*": {
+              "statusCode": "405",
```
mklander: The broad match here is partly to get it working, we could now anchor it ^${error_bad_request}.*? Might prevent something unexpected if someone breaks the conventions for the error strings.

alexwlchan:
Closing in favour of #2660.

PR #2672, opened 2018-09-20
Progress Monitor Flow

kenoir:

Adds a flow to send events to the progress monitor.

### Who is this change for?

📶 🐫

### DIFF
--- a/archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/progress/flows/ProgressUpdateFlow.scala
+++ b/archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/progress/flows/ProgressUpdateFlow.scala
@@ -1,0 +1,21 @@
+package uk.ac.wellcome.platform.archive.common.progress.flows
+
+import akka.NotUsed
+import akka.stream.scaladsl.Flow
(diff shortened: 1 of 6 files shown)


archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/progress/flows/ProgressUpdateFlow.scala
```diff
@@ -0,0 +1,21 @@
...
+
+object ProgressUpdateFlow {
+  def apply(progressMonitor: ProgressMonitor):
+  Flow[ProgressUpdate, Either[FailedProgressUpdate, ProgressUpdate], NotUsed] = {
+
+    Flow[ProgressUpdate].map(progressUpdate => {
+      Try(progressMonitor.update(progressUpdate)).toEither
+        .left.map(e => FailedProgressUpdate(e, progressUpdate))
+        .right.map(_ => progressUpdate)
+    })
```
mklander: I wonder if the progressMonitor should return an `Either` and deal in these type directly?

kenoir: Not sure really, the `progressMonitor` doesn't really _need to_ use an `Either`, whereas the stream does. 

Does @alicefuzier have an opinion?

Issue #2676, opened 2018-09-24
The snapshot generator queue alarm keeps firing…


alexwlchan:
…but snapshots are being created successfully!

Something funky is happening – we should find out why this alarm is misfiring, and fix it.

alexwlchan:
No alarms yesterday, but snapshots were both created. I was still experimenting, so I'll leave it another day before calling it done.

alexwlchan:
Didn't fire today either, so closing. 🎉

PR #2684, opened 2018-09-24
Improve Slack error reports from the snapshot scheduler

alexwlchan:
Part of #2676.

In particular, the Slack alarms now give a more helpful message and include the dates of the last two snapshots (if available):

screen shot 2018-09-24 at 11 44 14

For example, here we can see that the queue is filling up, but a snapshot was created yesterday – so the snapshot generator hasn't stalled.

### DIFF
--- a/monitoring/ecs.tf
+++ b/monitoring/ecs.tf
@@ -1,0 +1,3 @@
+resource "aws_ecs_cluster" "cluster" {
+  name = "monitoring"
+}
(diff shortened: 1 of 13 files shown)


monitoring/ecs.tf
```diff
@@ -0,0 +1,3 @@
+resource "aws_ecs_cluster" "cluster" {
```
alexwlchan: Previously the cluster was loaded from an output in the catalogue_pipeline stack, but that output no longer exists.

PR #2694, opened 2018-09-25
Add more mappings and ignore thumbs.db

tomcrane:

Reduce the error count on METS transformations for the new archive service

### Who is this change for?

Consumers of the new storage service

This adds some missing (but acceptable) file mappings, makes the transform process able to ignore certain named files, starting with `thumbs.db`


kenoir:
@tomcrane do you mind fixing the linting issues here so we can get this merged?

PR #2695, opened 2018-09-25
More improvements to the Archive API

alexwlchan:
* Load the Error model from the Catalogue API Swagger
* Point the health check endpoint at the right path
* Remove the nginx proxy container
* Redirect HTTP to HTTPS in Terraform
* Reduces the allocated CPU/memory
* Rename the `ns` variable to support having multiple namespaces


archive/archive_api/src/models.py
```diff
@@ -67,13 +69,5 @@ def fieldType(name, **kwargs):
...
-    'errorType': fields.String(description='The type of error', enum=['http']),
-    'httpStatus': fields.Integer(description='The HTTP response status code'),
-    'label': fields.String(
-        description='The title or other short name of the error'
-    ),
-    'description': fields.String(description='The specific error'),
-    'type': fieldType(name='Error'),
-})
+catalogue_swagger = json.load(open('catalogue/v2_swagger.json'))
+Error = api.schema_model('Error', catalogue_swagger['definitions']['Error'])
```
mklander: Nice to load this model, but not sure it should be coupled to the catalogue stack like this.

PR #2701, opened 2018-09-26
WC API load balancer infra

mklander:
Move the load balancer out of the catalogue_api stack into `infra_critical` (renamed from `catalogue_pipeline_data`) so that it can be shared among stacks to route from `api.wellcomecollection.org`.


catalogue_pipeline/terraform/pipelines/v2_pipeline/vhs_recorder.tf
```diff
@@ -1,5 +1,5 @@
 module "vhs_recorder" {
-  source = "../../../../catalogue_pipeline_data/terraform/vhs"
+  source = "..\/..\/..\/..\/critical_infra/terraform/vhs"
```
alexwlchan: Are these backslashes correct?

PR #2705, opened 2018-09-27
Improve the layout of the archive_api source tree

alexwlchan:
Probably the easiest way to review this is to clone it locally and poke around a bit. What's happened:

* The tests are now in a dedicated "tests" subdirectory, and we can get requirements.txt out of the "src" directory (this requires https://github.com/wellcometrust/dockerfiles/pull/45)
* The ingest API related code is in a dedicated "ingests" directory
* Previously `create_ingest_progress` would return correctly unless it hit a conditional check failed exception (all other errors would be swallowed) – now we raise other errors correctly.
* More validation tests.


alexwlchan:
I’ve merged https://github.com/wellcometrust/dockerfiles/pull/45; I’ll wait for those to rebuild and then re-run tests here.

PR #2707, opened 2018-09-27
Archive API /bags/xx-xxx endpoint

fractos:
This adds bag lookup capability to the archive API Flask app, which will fetch the contents of a bag using an identifier.

The endpoint on the API is /bags/xx-xxx  as defined in the RFC.

### DIFF
--- a/archive/archive_api/src/archive_api.py
+++ b/archive/archive_api/src/archive_api.py
@@ -160,6 +160,7 @@ def default_error_handler(error):
         'label': getattr(error, 'name', 'Internal Server Error'),
         'type': 'Error',
     }
+    logger.warn(error)
     if error_response['httpStatus'] != 500:
         if hasattr(error, 'data'):
             error_response['description'] = ', '.join(error.data.get('errors', {}).values())
--- a/archive/archive_api/src/archive_api.py
+++ b/archive/archive_api/src/archive_api.py
@@ -13,7 +13,6 @@ from werkzeug.exceptions import NotFound as NotFoundError
 import config
 import validators
 
-from bag_request import bag_request
 from bags import fetch_bag
 from ingests import (
     IngestProgress,
--- a/archive/archive_api/src/archive_api.py
+++ b/archive/archive_api/src/archive_api.py
@@ -14,6 +14,7 @@ import config
 import validators
 
 from bag_request import bag_request
+from bags import fetch_bag
 from ingests import (
     IngestProgress,
     create_ingest_progress,
@@ -139,7 +140,7 @@ class BagResource(Resource):
     def get(self, id):
         """Get the bag associated with an id"""
         try:
-            result = bag_request(
+            result = fetch_bag(
                 dynamodb_resource=app.config['DYNAMODB_RESOURCE'],
                 table_name=app.config['BAG_VHS_TABLE_NAME'],
                 s3_client=app.config['S3_CLIENT'],
--- a/archive/archive_api/src/archive_api.py
+++ b/archive/archive_api/src/archive_api.py
@@ -145,7 +145,7 @@ class BagResource(Resource):
             )
             return jsonify(result)
         except ValueError as error:
-            raise NotFoundError(error)
+            raise NotFoundError(str(error))
 
 
 @app.route('/storage/v1/healthcheck')
--- a/archive/archive_api/src/archive_api.py
+++ b/archive/archive_api/src/archive_api.py
@@ -8,6 +8,7 @@ from flask import Flask
 from flask_restplus import Api, Resource
 from flask import jsonify, make_response, request
 from werkzeug.exceptions import BadRequest as BadRequestError
+from werkzeug.exceptions import NotFound as NotFoundError
 
 import config
 import validators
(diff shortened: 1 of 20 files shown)


archive/archive_api/src/bag_request.py
```diff
@@ -0,0 +1,26 @@
...
+    table = dynamodb_resource.Table(table_name)
+    item_response = table.get_item(Key={'id': id})
+
+    try:
+        response = s3_client.get_object(
+            Bucket=bucket_name,
+            Key=item_response['Item']['s3key']
+        )
+    except KeyError:
+        raise NotFoundError(f'No bag found for id={id!r}')
```
alexwlchan: Minor: We shouldn't be raising a Werkzeug error here, because it's very HTTP specific – it's better to raise a regular Python exception, and catch/re-raise it in the context of an HTTP Flask app.

archive/archive_api/src/conftest.py
```diff
@@ -60,3 +76,16 @@ def create_table(dynamodb_client, table_name):
...
+
+@pytest.fixture
+def bag_id():
+    return str(uuid.uuid4())
+
+
+@pytest.fixture
+def bucket_bag(s3_client):
+    bucket_name = 'test-python-bucket-bag'
+    os.environ.update({'BAG_VHS_BUCKET_NAME': bucket_name})
```
alexwlchan: Don't forget to clean up this environment variable later.

archive/archive_api/src/conftest.py
```diff
@@ -60,3 +76,16 @@ def create_table(dynamodb_client, table_name):
...
+
+
+@pytest.fixture
+def bag_id():
+    return str(uuid.uuid4())
+
+
+@pytest.fixture
+def bucket_bag(s3_client):
+    bucket_name = 'test-python-bucket-bag'
```
alexwlchan: Does this bucket name need a random suffix so we can run tests in parallel?

archive/archive_api/src/test_bag_request.py
```diff
@@ -0,0 +1,87 @@
...
+        table_name_bag,
+        s3_client,
+        bucket_bag,
+        bag_id
+    )
+
+    assert response == stored_bag
+
+
+def run_around_tests(dynamodb_client, s3_client, bucket_bag, table_name_bag):
```
alexwlchan: Where's this method used?

fractos: @alexwlchan This was copy pasted in from the test_archive_bags.py file, but it looks like I missed the decorator `@pytest.yield_fixture(autouse=True)` so it won't be called at all. Worth adding do you think?

archive/archive_api/src/bag_request.py
fractos: @alexwlchan Copied from existing report_ingest_status.py code

archive/archive_api/src/conftest.py
fractos: If so then you probably want to update `bucket` in shared_conftest.py too: 
```@pytest.fixture
def bucket(s3_client):
    bucket_name = 'test-python-bucket'
    s3_client.create_bucket(Bucket=bucket_name)
    yield bucket_name```

File: archive/archive_api/src/bag_request.py (hunk 1 above)
Comment (mklander) [reply archive/archive_api/src/bag_request.py]: Yes, I think we should keep the http errors in the api layer.  I may have written it originally in the `report_ingest_status.py` code feel free to change it.

Comment (fractos) [reply archive/archive_api/src/bag_request.py]: I'll change the other files in a separate commit. Thanks.

File: archive/archive_api/src/archive_api.py [hunk 5]
```diff
@@ -141,6 +170,7 @@ def default_error_handler(error):
         'label': getattr(error, 'name', 'Internal Server Error'),
         'type': 'Error',
     }
+    logger.warn(error)
```
alexwlchan: 👍

archive/archive_api/src/bags/bag_request.py
```diff
@@ -0,0 +1,26 @@
...
+    table = dynamodb_resource.Table(table_name)
+    item_response = table.get_item(Key={'id': id})
+
+    try:
+        response = s3_client.get_object(
+            Bucket=bucket_name,
+            Key=item_response['Item']['s3key']
+        )
+    except KeyError:
+        raise NotFoundError(f'No bag found for id={id!r}')
```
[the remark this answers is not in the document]
fractos: hmm, thought I'd done this.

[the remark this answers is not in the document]
fractos: ah! I did. strike my last.

archive/archive_api/src/bags/bag_request.py
```diff
@@ -0,0 +1,26 @@
+# -*- encoding: utf-8
+import daiquiri
+import json
+
+from werkzeug.exceptions import NotFound as NotFoundError
+
+logger = daiquiri.getLogger()
+
+
+def bag_request(dynamodb_resource, table_name, s3_client, bucket_name, id):
```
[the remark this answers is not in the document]
fractos: ah, I guess this is a `bag` as opposed to a `bag_request` so that would be `fetch_bag` instead ?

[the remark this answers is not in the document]
fractos: (plus appropriate renames)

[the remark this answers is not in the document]
fractos: have made some renaming changes that are hopefully appropriate.

archive/archive_api/src/bags/fetch_bag.py
```diff
@@ -0,0 +1,22 @@
...
+    Fetch the contents of a Bag.
+    """
+
+    table = dynamodb_resource.Table(table_name)
+    item_response = table.get_item(Key={'id': id})
+
+    try:
+        response = s3_client.get_object(
+            Bucket=bucket_name,
+            Key=item_response['Item']['s3key']
```
alicefuzier: Err, only now realised that #2671 meant a change in the format of stuff stored in the VHS.
Each record has a `location` instead of an `s3Key`. The location is a json object:
```
{    "key" : { "S" : "archive/80/digitised/b22454408/c4efd0b3447783b68ef1dfd5c4acd4499bc5074a86f5a41de6a9b4ebe96df40a" },    "namespace" : { "S" : "wellcomecollection-vhs-archive-manifests" }  }
```
Basically `location` now contains both the key and the bucket name

fractos: Righto! Working on it...

alexwlchan:
This will need one more rebase against master for the messaging fixes, and then I’ll merge it.

Issue #2556, opened 2018-08-14
Add MARC 610 subjects from Sierra


jtweed:
```
first indicator 0
-----------------
Concept [-- Concept ]-- Place -- Period
a       [-- b       ]-- c     -- d

first indicator 1 or 2
----------------------
Organisation [-- Concept ]-- Place -- Period
a b          [-- b       ]-- c       -- d
```

silveroliver:
Proposal:

      /* Populate wwork:subjects organisation. Rules:
   * 
   * 1. For all bibliographic records use "marcTag": "610".
   * 2. Populate platform "label" from "610" concatenated with subfields a", "b", "c", "d", "e"  type as "Subject".
   * 3. Populate platform "concepts" "label" with subfield "a" and "b" type as "Organisation"
   * 4. If subfield 0 contains a value use it to populate "identifiers". For orgnisations the "identifierType" will be lc-names.
   *
   *
   *
   *
   *{
   *  "subjects": [
   *    {
   *      "label": "Wellcome Trust, sponsor.",
   *      "type": "Subject",
   *      "concepts": [
   *        {
   *          "id": "hd2891f",
   *          "label": "Wellcome Trust.",
   *          "type": "Organisation",
   *          "identifiers" : [
   *              {
   *              "indentifierType":
   *                 {
   *                 "id" : "lc-names" ,
   *                 "label" : "Library Of Congress Names" ,
   *                    "type" : "IdentifierType"
   *              } ,
   *              "value": "n81290903210",
   *              "type": "Identifier"
   *              }
   *        ] 
   *        }
   *      ]
   *    }
   *  ]
   *}
   */

alexwlchan:
This is live! Example here:

https://search.wellcomelibrary.org/iii/encore/record/C__Rb1128518?marcData=Y&lang=eng
https://api.wellcomecollection.org/catalogue/v2/works/ecdebckk?include=identifiers,subjects

PR #2708, opened 2018-09-27
Add the transformation for MARC 610 subjects

alexwlchan:
Implements the transform described in #2556.


catalogue_pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/sierra/subjects/SierraOrganisationSubjects.scala
```diff
@@ -0,0 +1,84 @@
...
+    // enough information to build the Organisation, so erroring out here is
+    // the best we can do for now.
+    if (label == "") {
+      throw TransformerException(
+        s"Not enough information to build a label on $varField")
+    }
+
+    getIdentifierSubfieldContents(varField) match {
+      case Seq(identifier) => {
+        val sourceIdentifier = SourceIdentifier(
```
alicefuzier: Because we don't expect any authority other than library on congress, for person subjects we only extract identifiers if the second indicator is 0. I know the 610 rules don't explictly say that but it feels like they should behave in the same way in this particular case...

alicefuzier: Or maybe it's worth checking with @silveroliver if that's the case here

PR #2719, opened 2018-10-01
Cap the concurrency in the Sierra item merger

alexwlchan:
An attempt to get the Sierra item merger working cleanly without tearing up the entire pipeline.

This change is published and being tested live! :D

### DIFF
--- a/reindexer/terraform/locals.tf
+++ b/reindexer/terraform/locals.tf
@@ -1,6 +1,7 @@
 locals {
-  vhs_sierra_table_name = "${data.terraform_remote_state.catalogue_pipeline_data.vhs_sierra_table_name}"
-  vhs_miro_table_name   = "${data.terraform_remote_state.catalogue_pipeline_data.vhs_miro_table_name}"
+  vhs_sierra_table_name       = "${data.terraform_remote_state.catalogue_pipeline_data.vhs_sierra_table_name}"
+  vhs_miro_table_name         = "${data.terraform_remote_state.catalogue_pipeline_data.vhs_miro_table_name}"
+  vhs_sierra_items_table_name = "${data.terraform_remote_state.catalogue_pipeline_data.vhs_sierra_items_table_name}"
 
   vpc_id          = "${data.terraform_remote_state.shared_infra.catalogue_vpc_id}"
   private_subnets = "${data.terraform_remote_state.shared_infra.catalogue_private_subnets}"
--- a/reindexer/terraform/main.tf
+++ b/reindexer/terraform/main.tf
@@ -25,3 +25,17 @@ module "sierra_reindexer" {
   service_egress_security_group_id = "${aws_security_group.service_egress_security_group.id}"
   namespace_id                     = "${aws_service_discovery_private_dns_namespace.namespace.id}"
 }
+
+module "sierra_items_reindexer" {
+  source = "./reindex_worker"
+
+  namespace      = "sierra_items"
+  vhs_table_name = "${local.vhs_sierra_table_name}"
+
+  reindex_worker_container_image = "${local.reindex_worker_container_image}"
+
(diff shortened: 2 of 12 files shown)


alexwlchan:
If you can work out how to do that, be my guest. 😄 

I had a look around, but I couldn't work out how to do it without a TypeSafe config file, and getting a TS config file into our app felt like a similarly sized piece of work.

alicefuzier:
We can change the default dispatcher in a similar way as explained here https://doc.akka.io/docs/akka/2.5.4/scala/dispatchers.html. [This](https://github.com/wellcometrust/platform/blob/master/archive/archivist/src/main/resources/application.conf) is an example of a typesafe config added to one of our apps (it's just the default akka one copied and pasted).
I'd very much like to try this approach first

alicefuzier:
I think we might want to keep using the fork-join-executor (it is the default executor according to [this](https://doc.akka.io/docs/akka/2.5/dispatchers.html). We only want to limit the maximum parallelism so something like this could work (also copied from that same page, but with parallelism-max set to 50):
```
my-dispatcher {
  # Dispatcher is the name of the event-based dispatcher
  type = Dispatcher
  # What kind of ExecutionService to use
  executor = "fork-join-executor"
  # Configuration for the fork join pool
  fork-join-executor {
    # Min number of threads to cap factor-based parallelism number to
    parallelism-min = 2
    # Parallelism (threads) ... ceil(available processors * factor)
    parallelism-factor = 2.0
    # Max number of threads to cap factor-based parallelism number to
    parallelism-max = 50
  }
  # Throughput defines the maximum number of messages to be
  # processed per actor before the thread jumps to the next actor.
  # Set to 1 for as fair as possible.
  throughput = 100
}
```

alexwlchan:
I think I've got the dispatcher config right, but I can't work out where we're creating the Actor it gets injected into… 😕

alicefuzier:
I don't see the config in this PR, so I'm not too sure. The dispatcher is created when the ActorSystem is created, based on the config file, so [here](https://github.com/wellcometrust/platform/blob/master/sbt_common/finatra_akka/src/main/scala/uk/ac/wellcome/finatra/akka/AkkaModule.scala). The ExecutionContext that we inject through Guice come directly from the actor system through [this module](https://github.com/wellcometrust/platform/blob/master/sbt_common/finatra_akka/src/main/scala/uk/ac/wellcome/finatra/akka/ExecutionContextModule.scala). It's not directly injected into an Actor (actually we never instantiate actors directly, I think). So, changing the default dispatcher in the application.conf should be all we need to do. There should be no scala code change for this

alicefuzier:
To clarify what I mean, I think [this](https://github.com/wellcometrust/platform/commit/2edd0d90e948f26147d61fba49ce8985ba82473e) should remove the need of any scala change.

The reason why I'm averse to the LazyFuture solution is because [ExecutionContext](https://docs.scala-lang.org/overviews/core/futures.html) is what the scala standard library provides to control Threads concurrency and the LazyFutures thing feels a bit of a hacky reinvention of the wheel. The whole point of Futures is that they provide an abstraction layer on top of Threads so you don't have to think about how they execute concurrently, so things like that LazyFuture are exactly the kind of thing that you shouldn't write with Futures. 

If I could be bothered creating a user on stackoverflow I would downvote that answer. The second answer is the correct one in my opinion.

We can pass an execution context that satisfies our needs either by creating one (like the old GlobalExecutionContext that we used to have) and importing it either implicitly or via Guice. Because we use akka, I think the akka streams config is the simplest way and  means we only have one excution context hanging around. But I'm not married to the akka streams config solution if it doesn't work. What I definitely would like to do, though, is use ExecutionContexts instead of LazyFuture.

alexwlchan:
Closing due to inactivity.

PR #2703, opened 2018-09-26
Archivist reports progress

alicefuzier:
👩‍🏫

PR #2725, opened 2018-10-02
ProgressUpdate takes a list of events

alicefuzier:


alicefuzier:
I'm going to fix thin in #2703

Issue #2718, opened 2018-10-01
A small number of Sierra bibs windows aren't working


alexwlchan:
```
$ python sierra_adapter/report_adapter_progress.py

===============================================================================
bibs windows
===============================================================================
1999-01-01T00:00:00 -- 2018-09-26T07:50:41.661400
2018-09-26T09:17:28.186653 -- 2018-10-01T10:32:10.389905
```

I've run the "build missing windows" script once already, and they all failed.

alexwlchan:
This is fixed, though I’m not sure why it worked on that attempt. (Was the Sierra API just having a bad moment? Have all the records updated in that window been updated again? Library gremlins?)

Anyway, it's sorted, and we have better progress reporting in Slack to let us know when it fails next time:

![screen shot 2018-10-03 at 13 08 53](https://user-images.githubusercontent.com/301220/46409443-85e9e280-c70d-11e8-9156-8100f31aef9c.png)

PR #2726, opened 2018-10-02
Add a Lambda that checks Sierra reader progress

alexwlchan:
This adds a crude Lambda that drops a message in our Slack whenever it discovers the Sierra reader has failed to fetch windows for at least six hours. Part of #2718, as the fact we noticed that was sheer luck!

Reporting is crude (it only says “there was an error”) because this is hopefully rare, and I don't want to spend any more time on it right now.


alexwlchan:
This also contains the fix for the failing tests on master

PR #2740, opened 2018-10-04
Add the @context field to every response from the Archive API

alexwlchan:
This uses [a custom Response class](https://blog.miguelgrinberg.com/post/customizing-the-flask-response-class) to ensure that we always include the `@context` parameter on our responses (even if it doesn't go anywhere yet!).

Importantly, it means we can keep this parameter out of our internal models, and never have to worry about it.


alexwlchan:
Ah, that might be why some of the tests are failing. Investigating.

PR #2743, opened 2018-10-05
Some archivist improvements

alicefuzier:


archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/messaging/SnsPublishFlow.scala
```diff
@@ -25,5 +26,8 @@ object SnsPublishFlow extends Logging {
         .flatMap(r => Try(snsClient.publish(r)))
 
     ProcessLogDiscardFlow[T, PublishResult]("sns_publish")(publish)
+      .withAttributes(
+        ActorAttributes.dispatcher(
+          "akka.stream.materializer.blocking-io-dispatcher"))
```
kenoir: Can we do this in a config file?

alicefuzier: I don't think so...Maybe, for flows like the SnsPublishFlow ( that we know is always blocking) we could add something to the configuration. But I couldn't find out how to do it. Anyway, in some other places it's not so easy. 
Also not sure whether this should be a config thing at all because it's basically telling akka that a certain flow performs blocking io operations. This info should live alongisde the code I think

alicefuzier:
Not merging yet, so travis doesn't override the container @mklander and I pushed with separate changes

PR #2746, opened 2018-10-05
Some other fixes for archive_api

alicefuzier:


alexwlchan:
I’m in the middle of a fairly major refactor of archive_api. Please can we keep these out of master, and I’ll fold them into my PRs as appropriate?

alicefuzier:
ook, but these are part of the reason why things don't talk to each other, so if they stay out of master it means what's in master stays broken

PR #2753, opened 2018-10-08
Adds callback-stub-server

kenoir:


.travis.yml
```diff
@@ -171,6 +171,8 @@ jobs:
     - env: TASK=progress_http-test
     - env: TASK=archive_api-test
     - env: TASK=notifier-test
+    - env: TASK=callback_stub_server-publish
```
alexwlchan: Even on pull requests?

kenoir: No, but this has not tests so?

archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/progress/models/Progress.scala
```diff
@@ -82,7 +70,10 @@ trait StatusConverters {
     )
 }
 
-object Progress extends URIConverters with StatusConverters {
+object Progress
+  extends URIConverters
+    with UUIDConverters
```
kenoir: ? Is this causing a bug ?

Issue #2634, opened 2018-09-05
Storage service MVP


kenoir:
In order to make the storage service available to clients for testing we need to ensure it can run an a bag ingest end to end with documented interfaces and error handling.

jtweed:
TODO:

# Storage service testing

## API

### Common

- Should be on api.wellcomecollection.org, over SSL
- Should it really be going through Cloudfront?
- Should have a base path of `/storage/v1`, not `/prod/storage/v1`
- Swagger should be available at `/storage/v1/swagger.json`
- Context should be available at `/storage/v1/context.json`
- Errors should use same model as the Catalogue API
- We need to agree a plan for authentication

### Ingests

- POST should not return a response body
- POST should require `type`, `ingestType` and `uploadUrl` to be present
- Validation on `type` and `ingestType` (only support `create` for now)
- Incorrect serialisation of `uploadUrl` (also check `callbackUrl`)
- Change `createdAt` to `createdDate`
- Change `modifiedAt` to `lastModifiedDate`
- Valid `id`s for `IngestResult` are `success` and `failure`
- Add `description`containing any overall ingest status or error message
- Remove detailed `events` for now, we can add as an enhancement later
- Add a link to the bag that was affected by the ingest (property name TBD)

### Bags

- Serialisation shouldn’t create objects with values
- Generate an `id` for the bag, don’t use the source identifier
- Implement `source` to show the source of the bag
- Implement `identifiers` to include the external IDs supplied with the bag
- Suggestion: change `Manifest` to `FileManifest` to differentiate it from IIIF manifest
- Implement `tagManifest`
- Add `description` from bag info
- Add `size` from bag info
- Remove `version`, we can add as an enhancement later
- Suggestion: remove `locations` and replace with `archiveUrl`, `replicaUrl` and `accessUrl`

## S3

### Common

- Namespace by content type, i.e. `digitised` for digitised content from Goobi (note, we will have other types from Goobi)
- Shard using the same scheme as VHS
- Consider moving buckets to separate storage account

### Archive

- Change bucket name to `wellcomecollection-storage-archive`
- Lifecycle objects to Glacier storage

### Access

- Store access copy in bucket called `wellcomecollection-storage-access`

alexwlchan:
Several things are ticked here which don't seem to be done – e.g. “Change createdAt to createdDate”. The Progress model still uses "createdAt" and "updatedAt".

Or is there something else I'm missing?

kenoir:
[...]

PR #2764, opened 2018-10-10
Align archive API with internal progress/ingest API model

mklander:
Update the archive api to return the  DispalyIngest model from progress http
Align the internal ingest/progress model to match the specification because the notifier should `post` the same model during callbacks.

part of #2634


archive/progress_http/src/main/scala/uk/ac/wellcome/platform/archive/progress_http/Router.scala
```diff
@@ -40,7 +42,8 @@ class Router @Inject()(monitor: ProgressMonitor, config: HttpServerConfig) {
         get {
           // TODO add test for what happens if id is not a valid UUID
           monitor.get(UUID.fromString(id)) match {
-            case scala.Some(progress) => complete(progress)
+            case scala.Some(progress) =>
```
alicefuzier: scala.?

functions.Makefile
```diff
@@ -180,9 +180,10 @@ endef
 #   $1 - Path to the docker-compose file.
 #
 define docker_compose_up
-    $(ROOT)/docker_run.py --dind --sbt --root -- \
-    		--net host \
-    		docker/compose:1.21.0 -f $(ROOT)/$(1) up
+	$(ROOT)/docker_run.py --dind --sbt --root -- \
+		--net host \
+		wellcome/sbt_wrapper \
+		"project $(1)" "dockerComposeUp"
```
alicefuzier: why this? I intentionally bypassed sbt when I wrote this, because sbt wants to resolve the dependencies etc, before starting the containers, which makes it super slow. Eventually I just found that I was using docker-compose directly from command line cause sbt was too painful

alexwlchan: If the directory structure doesn't match the sbt project name (as is the case for archive_common and archive/common), this task fails.

alicefuzier: yeah I noticed that. I'm not sure this is less painful than having to call docker-compose directly just form archive_common. But we shall see.

Issue #2691, opened 2018-09-25
Sierra item merger is broken


alicefuzier:
The sierra item merger has intermittent but very frequent failures that cause lots of messages to go in the DLQ. Redriving them makes some of them succeed but, because failure are more frequent than successes, a lot of them go back in the DLQ. As a result the sierra item merger can't keep up with updates from sierra and the DLQ keeps growing.

![cloudwatch management console](https://user-images.githubusercontent.com/26736746/46005413-1d678980-c0ad-11e8-9f89-ea92028598fb.png)
The blue line is the number of failures reported by the sierra item merger and the green line is the number of successes. As the graph shows, failures far outweigh successes.

Checking logs, the errors seem to be 

> com.amazonaws.SdkClientException: Unable to execute HTTP request: Timeout waiting for connection from pool

when trying to get an s3 connection. 
This could be due to two reasons:
* We don't close the object input stream when reading from s3 (which was fixed in https://github.com/wellcometrust/scala-storage/pull/31)
* We have more threads running than the number of connections available in the pool (the default number of connections is 50)

alicefuzier:
Quick update: I picked one of the failing messages (itemId: 1503297) and it has way more than 50 bibRecords. Because the sierra item merger updates the VHS for each bibRecord, this message always fails.

alexwlchan:
@alicefuzier How many more?

pollecuttn:
That item is attached to this bib http://catalogue.wellcomelibrary.org/record=b1579198 with links to more (I can't easily see in Sierra) which must be other bibs from this phonebox cards ephemera collection. There's probably a number of things like this in the ephemera collection.

alicefuzier:
@alexwlchan Don't know exactly, that's how far my digging went. But I redrove the queue a bunch of times and I think there are ~500 messages which just keep going back to the DLQ eventually

alicefuzier:
@alexwlchan ah I see you meant how many more bibrecords. I couldn't be bothered to count them, I just scrolled the json. We can always retrieve that item from the sierra_items_to_dynamo VHS to get a precise number. For now, I just wanted to record that item number here so that we can investigate when we pick up this issue

alexwlchan:
Fiddling with futures in #2719 seems to have worked. I'm going to have another go at that with Akka configuration for Alice, see if works the same.

PR #2769, opened 2018-10-11
Support reindexing the sierra_items VHS

alexwlchan:
These changes allow us to run a reindex on the sierra_items VHS, for example if the sierra_item_merger fails and loses records. It doesn't modify the item merger itself.

Part of #2691. Spun out of #2719.


alexwlchan:
It passed formatting, so merging without waiting for the rest of Travis to run.

PR #2771, opened 2018-10-11
Add space to bags-api request endpoint

kenoir:

Allow the space to be present in the URL `/bags/space/external-id`

### Who is this change for?

🎤 💇‍♂️

### DIFF
--- a/archive/archive_api/src/apis/bags.py
+++ b/archive/archive_api/src/apis/bags.py
@@ -20,7 +20,7 @@ api.add_model("Source", definition=Source)
 
 
 @api.route("/<space>/<id>")
-@api.param("space", "The space!")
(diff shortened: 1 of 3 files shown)


archive/archive_api/src/apis/bags.py
```diff
@@ -19,27 +19,30 @@
 api.add_model("Source", definition=Source)
 
 
-@api.route("/<id>")
+@api.route("/<space>/<id>")
+@api.param("space", "The space!")
```
alexwlchan: “space” isn’t an especially helpful term here, and this description isn’t helping!

What’s wrong with “namespace”?

kenoir: It is the will of @jtweed. OBEY.

PR #2849, opened 2018-10-16
[WIP] DDS IAM role to access storage S3

fractos:
This will allow an AWS API user to be created that the DDS can use to access the S3 bucket for archival storage.

This change is for the DDS application.


fractos:
@kenoir This will currently not work as the variable holding the DDS ARN isn't defined yet. I have this working now, but only if the assume role policy references the DDS user specifically, though based on the conversation we had about the underlying unique ID then this may be far less of a concern.

fractos:
This is no longer required.

Issue #2844, opened 2018-10-15
Provide bag manifests as per RFC


kenoir:
https://github.com/wellcometrust/platform/tree/master/docs/rfcs/002-archival_storage#api-1

**TODO:**

- [ ] Update registrar to store all data necessary (tag-manifest excluded).
- [x] Create `registrar-http` service mirroring `progress-async` pattern to provide translation to display body and control access to progress dynamo table.
- [x] /bags API should present bag manifest with JSON-LD context and types.

PR #2851, opened 2018-10-16
Creating registrar_http service

alicefuzier:
That exposed the display model for the bags endpoint. For #2844


alicefuzier:
This is deployed and tested

PR #2859, opened 2018-10-17
Make archive_api call registrar_http

alicefuzier:


archive/archive_api/src/bags_manager.py
```diff
@@ -0,0 +1,51 @@
...
+    """Raised if we get an unexpected error from the progress service."""
+
+
+class BagNotFoundError(Exception):
+    """Raised if we try to look up a progress that doesn't exist."""
+
+
+class BagsManager:
+    """
+    Handles requests to/from the progress service.
```
kenoir: 💥 Comments 💥

alicefuzier: oops

archive/archive_api/src/apis/bags.py
```diff
@@ -32,22 +32,17 @@ def get(self, space, id):
...
-                bucket_name=app.config["BAG_VHS_BUCKET_NAME"],
-                id=bag_id,
-            )
+            result = bags_manager.lookup_bag(space, id)
 
             # TODO: Remove the necessity to do this
-            result["id"] = result["id"]["value"]
+            result[
+                "id"
+            ] = f'{result["id"]["space"]}/{result["id"]["externalIdentifier"]}'
```
alexwlchan: I think I might just turn off black if this is the formatting it's going to apply, because this is bizarre.

alicefuzier: It is, but it's probably my fault for putting too much into the same line. I don't think it can break that line differently?

PR #2861, opened 2018-10-18
Enrich service status JSON with task definition & update dashboard

kenoir:

In order to have better visibility of the version of a deployed service this change adds more task information to the service status JSON used by the ECS dashboard.

sc

Now with added _drop shadow_!

### Who is this change for?

Developers!


monitoring/ecs_dashboard/update_service_list/src/update_service_list.py
```diff
@@ -26,58 +26,66 @@ def _create_event_dict(event):
...
-    return cluster_list
+def get_service_list(ecs_client, cluster_arn):
+    return [
+        _enrich_service(ecs_client, cluster_arn, service_arn)
+        for service_arn in get_service_arns(ecs_client, cluster_arn)
+    ]
+
+
+def _enrich_cluster_list(ecs_client, cluster_arn):
+    cluster = (describe_cluster(ecs_client, cluster_arn),)
```
alexwlchan:
```suggestion
    cluster = describe_cluster(ecs_client, cluster_arn)
```

monitoring/ecs_dashboard/update_service_list/src/update_service_list.py
```diff
@@ -26,58 +26,66 @@ def _create_event_dict(event):
...
-        service_list.append(_create_service_dict(service))
+    zipped = zip(service["deployments"], task_definitions)
 
-    return service_list
+    enriched_deployments = [
+        _enrich_deployment(deployment, task_definition)
+        for deployment, task_definition in zipped
+    ]
 
+    enriched_events = [_create_event_dict(e) for e in service["events"][:5]]
```
alexwlchan: Why are you only getting the first five events?

kenoir: Because there could be _a lot_ of them (and only the recent ones are useful for the dashboard).

monitoring/ecs_dashboard/update_service_list/src/update_service_list.py
```diff
@@ -6,12 +6,10 @@
 It does not use the event data from the event.
 """
 
+import boto3
```
alexwlchan: nit: Depending on which style guide you adhere to, the original was “more“ correct, insofar as import order matters.

boto3 isn't a standard library module, so it goes in the second group of imports.

monitoring/ecs_dashboard/update_service_list/src/update_service_list.py
alexwlchan: Ah, I see the original comment – can we keep that please? :-)

PR #2863, opened 2018-10-18
Add queue-watcher lambda to provide custom metrics on queue work

kenoir:

In order to allow application autoscaling to prevent scale down while messages are still being worked on, we need to react to in-flight & visible message counts.

A prototype has been running for the last couple of weeks and the data looks ok:

q_data

### Who is this change for?

⚖️ 🤖


monitoring/queue_watcher/README.md
```diff
@@ -0,0 +1,3 @@
+# slack_budget_bot
+
```
alicefuzier: Budget bot!

alicefuzier:
Actually one question before merging: have you checked the impact on cloud watch costs? If I read the code correctly this is sending metrics every minute without batching the metric data...

kenoir:
> Actually one question before merging: have you checked the impact on cloud watch costs? If I read the code correctly this is sending metrics every minute without batching the metric data...

Costs don't seem significant, I think because the once a minute send rate is relatively low.

Issue #2865, opened 2018-10-19
Reindexer should be able to publish to a particular topic


alexwlchan:
So Harrison can run reindexes without triggering an entire pipeline reindex.

kenoir:
closing as superseded by infra updates in https://github.com/wellcometrust/platform/pull/2871

PR #2867, opened 2018-10-19
Break out message-writing logic from SNSWriter

alexwlchan:
A bit of speculative refactoring for #2865 while I wait for scripts to run.

Our current SNSWriter is tied to the idea that it publishes to a single topic – if we want Harrison to be able to reindex to a different topic, we could either create even more services configured with new topics, or break this assumption. The latter seems easier, so this patch:

* Adds a new class which takes the topic ARN as a parameter:

    ```scala
    class SNSMessageWriter {
        def writeMessage(message: String,
                         subject: String,
                         topicArn: String): Future[PublishAttempt] = ...
    }
    ```

* Modifies SNSWriter to use the new class when sending messages.

There should be no change to existing services, but it would a future patch to send to different topics within the same reindexer service.

[Note: I have no spare local CPU capacity, so this patch is currently untested.]

### DIFF
--- a/sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/sns/SNSMessageWriter.scala
+++ b/sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/sns/SNSMessageWriter.scala
@@ -1,0 +1,42 @@
+package uk.ac.wellcome.messaging.sns
+
+import com.amazonaws.services.sns.AmazonSNS
+import com.amazonaws.services.sns.model.PublishRequest
+import com.google.inject.Inject
+import grizzled.slf4j.Logging
+import io.circe.Encoder
+import uk.ac.wellcome.json.JsonUtil._
+
+import scala.concurrent.{ExecutionContext, Future, blocking}
+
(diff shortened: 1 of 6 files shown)


sbt_common/messaging/src/main/scala/uk/ac/wellcome/messaging/sns/SNSMessageWriter.scala
```diff
@@ -0,0 +1,45 @@
...
+import scala.concurrent.{blocking, ExecutionContext, Future}
+
+/** Writes messages to SNS.  This class can write to a different topic on
+  * every request.
+  *
+  */
+class SNSMessageWriter @Inject()(snsClient: AmazonSNS)(
+  implicit ec: ExecutionContext)
+    extends Logging {
+  def writeMessage(message: String,
```
kenoir: Might be worth allowing currying here, i.e.

```scala
def writeMessage(topicArn: String)(message: String, subject: String) = {
  // ...
}
```

Then you can create a writer like:

```scala
val writer = writer.writeMessage(topicArn) _

writer("subject", "message")
writer("anotherSubject", "message")
```

PR #2869, opened 2018-10-19
Callback when completed or failed

mklander:

Make callbacks when the progress status is Completed or Failed and the callback status is Pending.  This fixes a problem that has arisen because callback status is now separated from the processing status leading to multiple callbacks.

However this change still assumes only one message is sent with the final status change, which should also be fixed.


archive/progress_async/src/main/scala/uk/ac/wellcome/platform/archive/progress_async/flows/CallbackNotificationFlow.scala
```diff
@@ -35,9 +39,18 @@ object CallbackNotificationFlow extends Logging {
...
+      case progress @ Progress(
+            id,
+            _,
+            _,
+            Some(callback),
+            progressStatus,
+            _,
+            _,
+            _,
+            _) =>
```
kenoir: W
a
a
a
a
a
a
a
a
a
a
a
h
h
h
h
.
.
.
.

PR #2889, opened 2018-10-24
Allow Chemist and Druggist to be bagged

tomcrane:

Remove the artificial block on creating a bag for Chemist and Druggist

### Who is this change for?

Anyone who wants C&D to survive in the new storage


kenoir:
😱

PR #2891, opened 2018-10-24
Mets checksum map

tomcrane:

Add a more useful name for the algorithm used to generate a checksum in METS


archive/bagger/src/identifiers.py
```diff
@@ -9,7 +9,7 @@ def normalise_b_number(b_str):
     if len(b_str) != 8:
         raise ValueError("Not a valid b number")
 
-    if b_str == "19974760":
-        raise ValueError("Will not process Chemist and Druggist right now")
+    # if b_str == "19974760":
+    #     raise ValueError("Will not process Chemist and Druggist right now")
```
kenoir: This is going to conflict with the enable c&d branch no?

tomcrane: accidentally added that file, have reverted

PR #2897, opened 2018-10-26
Implement filters for workType

### LINKED ISSUE #2798: Filter works by work type

alexwlchan:
While waiting for the Tandem Vault APIs to stabilise, I started working on this, and it fell out rather neatly. Closes #2798.

Note: the filter is implemented on both the V1 and the V2 APIs because that’s easiest and I couldn’t see a reason not to, but it wouldn’t be much more complicated to make it V2 only. @jtweed – thoughts?

### DIFF
--- a/catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
+++ b/catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
@@ -227,6 +227,11 @@ abstract class WorksController[M <: MultipleResultsRequest[W],
         |To search for any of these special characters, they should be escaped with \.""".stripMargin,
         required = false
       )
+      .queryParam[String](
+        "workType",
+        "Filter by the workType of the searched works",
+        required = false
+      )
       .parameter(includeSwaggerParam)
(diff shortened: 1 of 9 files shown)


jtweed:
Two things from me:

1. What's with the duplication? It might be ok now, but it's going to be completely untenable as we add more filters.
2. The term query should be added in filter context, not query context, to ensure that it is non-scoring.

alexwlchan:
@jtweed 

1. Duplication across the V1/V2 API, or somewhere else?
2. I’ve switched it to use a filter context, same for the IdentifiedWork type.

alexwlchan:
@jtweed The repetition is a moderately fiddly change that I was hoping to put off for a bit longer; I’ve put it in a separate PR.

kenoir:
Im ok with the repetition for now, ([rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming))).

alexwlchan:
I think the code complexity is similar for V1/V2 and V2-only.

@jtweed Any strong opinions?

alexwlchan:
I’m YOLO’ing this PR and merging with V1 and V2 filters.

The code complexity is similar whether you do V1+V2 or V2-only, but the former is simpler and more useful.

PR #2900, opened 2018-10-28
Use function currying to reduce repetition in {Works,Elasticsearch}Service

alexwlchan:
Follows #2897.

Since @jtweed pointed out we have a quite a bit of repetition in these two services, and that’s going to make adding new filters harder, I had a stab at cleaning it up. I wanted to try currying and partial functions again, and I think it’s finally “clicked”.

This patch:

* Introduces two new case classes `WorksSearchOptions` and `ElasticsearchQueryOptions`, which absorb the shared logic of things like filters, page size, index name, and so on
* Modifies the list/search methods in the services to use currying, so you pass the options separately from method-specific config
* Updates tests and so on to match

Separate from #2897 because we can add workType filtering without fixing the repetition, and I think this is substantial enough to merit its own review.

It won’t run in Travis until we merge the original PR, but I've run tests locally and it’s all green. 💚

### DIFF
--- a/catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
+++ b/catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
@@ -105,12 +105,12 @@ abstract class WorksController[M <: MultipleResultsRequest[W],
       pageNumber = request.page
     )
 
-    val curriedSearch: (WorksSearchOptions) => Future[ResultList] = request.query match {
+    def searchFunction: (WorksSearchOptions) => Future[ResultList] = request.query match {
       case Some(queryString) => worksService.searchWorks(queryString)
       case None => worksService.listWorks
     }
 
-    curriedSearch(worksSearchOptions)
+    searchFunction(worksSearchOptions)
   }
 
   private def generateSingleWorkResponse[T <: DisplayWork](
--- a/catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
+++ b/catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
@@ -105,10 +105,11 @@ abstract class WorksController[M <: MultipleResultsRequest[W],
       pageNumber = request.page
     )
 
-    def searchFunction: (WorksSearchOptions) => Future[ResultList] = request.query match {
-      case Some(queryString) => worksService.searchWorks(queryString)
-      case None => worksService.listWorks
-    }
+    def searchFunction: (WorksSearchOptions) => Future[ResultList] =
+      request.query match {
+        case Some(queryString) => worksService.searchWorks(queryString)
+        case None              => worksService.listWorks
+      }
 
     searchFunction(worksSearchOptions)
   }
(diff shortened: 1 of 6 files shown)


catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/services/WorksService.scala
```diff
@@ -1,19 +1,24 @@
...
+case class WorksSearchOptions(
+  workTypeFilter: Option[String],
+  indexName: String,
+  pageSize: Int,
+  pageNumber: Int
+)
+
 @Singleton
-class WorksService @Inject()(
-  apiConfig: ApiConfig,
```
alexwlchan: One small nicety: we get to ditch the dependency between WorksService and ApiConfig, which was always a bit odd – the default page size now gets passed higher up, in the WorksController.

catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
```diff
@@ -102,27 +97,20 @@ abstract class WorksController[M <: MultipleResultsRequest[W],
...
-        )
+  private def getWorkList(request: M, pageSize: Int): Future[ResultList] = {
+    val worksSearchOptions = WorksSearchOptions(
+      workTypeFilter = request.workType,
+      indexName = request._index.getOrElse(indexName),
+      pageSize = pageSize,
+      pageNumber = request.page
+    )
+
+    val curriedSearch: (WorksSearchOptions) => Future[ResultList] = request.query match {
```
kenoir: This is a *partial function*, rather than a *partially applied function*.  

To make things more complicated a curried function is a partially applied function, but not all partially applied functions are curried!

- **Currying:** "Decomposition of a function with multiple arguments into a chain of single-argument functions."
- **Partial application of function:** "Passing to a function less arguments than it has in its declaration."
- **Partial function:** "A partial function of type PartialFunction[A, B] is a unary function where the domain does not necessarily include all values of type A."

catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/services/ElasticsearchService.scala
```diff
@@ -5,12 +5,21 @@ import com.sksamuel.elastic4s.http.ElasticDsl._
...
+import com.sksamuel.elastic4s.searches.SearchDefinition
 import com.sksamuel.elastic4s.searches.queries.BoolQueryDefinition
 import com.sksamuel.elastic4s.searches.queries.term.TermQueryDefinition
+import com.sksamuel.elastic4s.searches.sort.FieldSortDefinition
 import uk.ac.wellcome.elasticsearch.DisplayElasticConfig
 
 import scala.concurrent.Future
 
+case class ElasticsearchQueryOptions(
+  workTypeFilter: Option[String],
```
kenoir: Perhaps not here, but can we make this even more generic, with `ElasticsearchQueryOptions` having a list of `WorkFilter`, which are themselves typed?

e.g.

```scala
// Filters
sealed trait WorkFilter

case class WorkTypeFilter(workType: WorkType) extends WorkFilter
// ...
```

catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/controllers/WorksController.scala
alicefuzier: I'd say this is just a function. Since `Function` is a first class type in scala, this is just a function value

alicefuzier: Actually, why is this a val, rather than a normal def?

PR #2909, opened 2018-10-29
Implement multiple workType filters, e.g. ?workType=a,b

alexwlchan:
Follows #2901, another part of #2798.

This patch:

* Adds a generic `WorkFilter` trait, with one instance `WorkTypeFilter`
* Modifies `WorkTypeFilter` to support matching on multiple values
* Modifies ElasticsearchQueryOptions and WorksSearchOptions to have a more general `List[WorkFilter]` parameter
* Adds a bunch of tests for filtering against multiple values

### DIFF
--- a/catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkFilter.scala
+++ b/catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkFilter.scala
@@ -1,0 +1,14 @@
+package uk.ac.wellcome.platform.api.models
+
+import uk.ac.wellcome.models.work.internal.WorkType
+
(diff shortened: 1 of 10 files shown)


catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/models/WorkFilter.scala
```diff
@@ -0,0 +1,10 @@
+package uk.ac.wellcome.platform.api.models
+
+sealed trait WorkFilter
+
+case class WorkTypeFilter(workTypeIds: Seq[String]) extends WorkFilter
```
kenoir: We've got `WorkType`. Why not use it?

PR #2912, opened 2018-10-29
Fix the ALB redirect to developers.wellcomecollection.org

alexwlchan:
@kenoir I’ve not applied this because it might break your stuff, but here’s a fix for the redirects on api.wc.org.


kenoir:
Did not even realise it was broken! But this is fine to apply.

Issue #2908, opened 2018-10-29
Progress HTTP service replaces flask archive API


kenoir:
- [x] Accept POST /storage/ingests in `DisplayIngest` model
- [x] Send to SNS topic as `IngestBagRequest` 
  - [x] Subscribe archivist queue

PR #2915, opened 2018-10-30
Send IngestBagRequest to SNS when getting a POST in progress_http

alicefuzier:
Part of #2908

### DIFF
--- a/archive/progress_common/src/main/scala/uk/ac/wellcome/platform/archive/common/progress/monitor/ProgressTracker.scala
+++ b/archive/progress_common/src/main/scala/uk/ac/wellcome/platform/archive/common/progress/monitor/ProgressTracker.scala
@@ -21,10 +21,10 @@ class ProgressTracker @Inject()(
 )(implicit ex: ExecutionContext) extends Logging {
   import Progress._
 
-  def get(id: UUID): Option[Progress] = {
-    Scanamo.get[Progress](dynamoClient)(dynamoConfig.table)(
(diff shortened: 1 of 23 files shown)


archive/progress_common/src/main/scala/uk/ac/wellcome/platform/archive/common/progress/monitor/ProgressTracker.scala
```diff
@@ -12,18 +12,19 @@ import grizzled.slf4j.Logging
...
   dynamoClient: AmazonDynamoDB,
   dynamoConfig: DynamoConfig
-) extends Logging {
+)(implicit ex: ExecutionContext) extends Logging {
   import Progress._
 
-  def get(id: UUID): Option[Progress] = {
-    Scanamo.get[Progress](dynamoClient)(dynamoConfig.table)(
+  def get(id: UUID): Future[Option[Progress]] = Future {
+    blocking(Scanamo.get[Progress](dynamoClient)(dynamoConfig.table)(
```
kenoir: Why `blocking`?

alicefuzier: https://www.cakesolutions.net/teamblogs/demystifying-the-blocking-construct-in-scala-futures
Basically, it's to tell the execution context that this operation is going to block, so that the execution context can treat it differently. Not all execution contexts do something special for blocking operations but it's good practice to add it, so the ones that do can be notified

PR #2917, opened 2018-10-30
Adele – Someone Like You(nicode)

alexwlchan:
Here’s a tiny fix for some borked Unicode data in the Miro exports.

Here’s the Sierra record with the correct spelling (and also what looks like a dodgy reference to a Miro image): https://search.wellcomelibrary.org/iii/encore/record/C__Rb1568629?lang=eng

### DIFF
--- a/catalogue_pipeline/transformer/transformer_miro/src/main/scala/uk/ac/wellcome/platform/transformer/miro/source/MiroTransformableData.scala
+++ b/catalogue_pipeline/transformer/transformer_miro/src/main/scala/uk/ac/wellcome/platform/transformer/miro/source/MiroTransformableData.scala
@@ -45,16 +45,25 @@ case object MiroTransformableData {
    *
    * We need to fix them up before we decode as JSON.
    */
-  private def unescapeHtml(data: String): String =
-    StringEscapeUtils.unescapeHtml(data)
+  private def unescapeHtml(s: String): String =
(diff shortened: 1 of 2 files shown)


catalogue_pipeline/transformer/transformer_miro/src/main/scala/uk/ac/wellcome/platform/transformer/miro/source/MiroTransformableData.scala
```diff
@@ -45,16 +45,24 @@ case object MiroTransformableData {
    *
    * We need to fix them up before we decode as JSON.
    */
-  private def unescapeHtml(data: String): String =
-    StringEscapeUtils.unescapeHtml(data)
+  private def unescapeHtml(s: String): String =
+    StringEscapeUtils.unescapeHtml(s)
```
mklander: 👍

alexwlchan:
> Can we fix these more generally?

The issues is not knowing where they are, unfortunately. Once we have the reporting pipeline, we can scan for more general classes of Unicode WTFs, and maybe then write a more general fix (or just fix the metadata).

The “best” fix is to create Sierra records that get merged with the Miro work, so we can override the old metadata.

mklander:
👍 Thanks, yes let's find them with reporting 😄

Issue #2800, opened 2018-10-11
Filter works by digitised image availability


jtweed:
This is so that the Experience can filter for images that actually have an associated image available to view online. Following the pattern in the RFC, matching that use case should look like:

?locations.locationType=iiif-image

PR #2921, opened 2018-10-30
Implement filters for Item LocationType

alexwlchan:
This patch allows you to make queries of the form:

```
GET /catalogue/v2/works?items.locations.locationType=iiif-image
```

and retrieve all the works that have an item with a `iiif-image` location – which should be everything that includes a Miro record.

Additionally, this patch removes the workType filter from the V1 API, because every V1 work has the same workType so it's not very useful, and turns out to be less complicated than I thought.

Reviewers: @jtweed for behaviour, @wellcometrust/platform-devs for code.

Part of #2800.


catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/requests/Requests.scala
```diff
@@ -19,6 +19,7 @@ trait MultipleResultsRequest[W <: WorksIncludes] extends ApiRequest {
   val include: Option[W]
   val query: Option[String]
   val workType: Option[String]
+  val itemLocationType: Option[String]
```
alicefuzier: if workType and locationType are not available for V1 resquests why do we add them here instead of V2MultipleResultsRequest?

alexwlchan: Because the logic for building the filters is inside WorksController, and is quite generic about the request type.

I want to keep it generic because I think some of the filters will get backported to V1 (contributors and subjects), and it leaves the door open for generic filters in V3. But not strongly tied to it.

jtweed: Just to confirm, there will be no requirement for filtering in V1.

Issue #2907, opened 2018-10-29
Progress HTTP service replaces flask archive API:


kenoir:
- [ ] Accept POST /storage/ingests in `DisplayIngest` model
- [ ] Send to SNS topic as `IngestBagRequest` 
  - [ ] Subscribe archivist queue

kenoir:
DUPE!

PR #2924, opened 2018-10-30
[WIP] Make progress_http accept DisplayIngest in POST

alicefuzier:
Part of #2907


archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/models/DisplayIngest.scala
```diff
@@ -8,21 +10,33 @@ import uk.ac.wellcome.platform.archive.common.progress.models.{
...
+                                 events: Seq[DisplayProgressEvent] = Seq.empty,
+                                 createdDate: String,
+                                 lastModifiedDate: String,
+                                 @JsonKey("type")
+                                 ontologyType: String = "Ingest")
+    extends DisplayIngest
 
 case class DisplayCallback(uri: String,
-                           status: String,
+                           status: Option[String],
```
kenoir: When can this be `None`?

archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/progress/models/Progress.scala
```diff
@@ -40,12 +41,13 @@ case object Progress extends URIConverters {
     override def toString: String = failedString
   }
 
-  def apply(createRequest: ProgressCreateRequest): Progress = {
+  def apply(createRequest: RequestDisplayIngest): Progress = {
```
kenoir: Apropos of this, do we want to rename the apps to something less generic? Happy to concede that they should be named for their purpose (which is now `Ingest`).

archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/models/DisplayIngest.scala
alicefuzier: When it's supplied as part of a POST. I should have adopted the same approach as DiplayIngest to be honest and split this in two classes

archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/progress/models/Progress.scala
alicefuzier: Yeah, I think possibly the names should be revised

PR #2930, opened 2018-10-31
Reformat reporting transformer

harrisonpim:
Cleans up the transformer example, moving the `transform()` function into a separate file. Modifications to the transformer shouldn't affect the rest of the process (moving records from VHS to the transformer, and from the transformer to the ES index)

### Who is this change for?
📄📊


reporting/transformer_example/src/transformer_example.py
```diff
@@ -50,82 +39,82 @@ class ElasticsearchRecord(object):
...
+    Returns
+    -------
+    data_dict : dict
+        object data in malleable dict format
+    """
+    object_location = hybrid_record.location
+    s3_object = s3.get_object(Bucket=object_location.namespace, Key=object_location.key)
+    record = s3_object["Body"].read().decode("utf-8")
+    record_data = json.loads(record)["data"]
+    return record_data
```
alexwlchan: Suggestion: don’t combine these two functions into one. Previously you had a function that could read from S3, another that unpacked that as an ElasticsearchRecord. Keeping them separate makes it easier to reuse parts of them later.

reporting/transformer_example/src/transformer_example.py
```diff
@@ -50,82 +39,82 @@ class ElasticsearchRecord(object):
...
-    hybrid_records = [
-        _extract_hybrid_record(raw_record) for raw_record in raw_hybrid_records
+    hybrid_records = [HybridRecord(**record) for record in raw_hybrid_records]
+    data_dicts = [
+        hybrid_record_to_data_dict(s3, hybrid_record)
+        for hybrid_record in hybrid_records
     ]
-    return [_build_es_record(s3, hybrid_record) for hybrid_record in hybrid_records]
-
+    return data_dicts, hybrid_records
```
alexwlchan: You’re returning both of these because you want to combine them later, I think?

A dict would be a nicer data structure for this. And possibly not doing it all at once – how familiar are you with Python generators? Might be nice to go through some of those together at some point, it would clean up a bunch of this code.

(If you get a “TypeError: unhashable type: ‘HybridRecord’”, that’s fixable. I can go through it with you.)

reporting/transformer_example/src/transformer_example.py
```diff
@@ -50,82 +39,82 @@ class ElasticsearchRecord(object):
...
+    an elasticsearch index
+    """
     s3_client = s3_client or boto3.client("s3")
-
-    elasticsearch_client = elasticsearch_client or elasticsearch_client(
-        url=os.environ["ES_URL"],
-        username=os.environ["ES_USER"],
-        password=os.environ["ES_PASS"],
+    index = index or os.environ["ES_INDEX"]
+    doc_type = doc_type or os.environ["ES_DOC_TYPE"]
```
alexwlchan: 👍 for moving this stuff up, and getting environment parsing out of the way early.

reporting/transformer_example/src/transformer_example.py
harrisonpim: wrt generators, were you thinking something along these lines?

```python
def extract_records(s3, event):
    for record in event["Records"]:
        raw_hybrid_record = json.loads(record["Sns"]["Message"])
        hybrid_record = HybridRecord(**raw_hybrid_record)
        s3_object = fetch_object_from_s3(s3, hybrid_record)
        data_dict = s3_object_to_data_dict(s3_object)
        yield data_dict, hybrid_record
```

 happy to wait until friday to chat in person if you'd prefer

reporting/transformer_example/src/transformer_example.py
kenoir: Also though possibly @alexwlchan would object, having separate functions with descriptive names goes some way to having the code be self documenting.

reporting/transformer_example/src/transformer_example.py
alexwlchan: Along the right lines, but I’d decompose it into a bunch of separate generators that you chain together. More like:

```python
def extract_sns_messages_from_event(event):
    for record in event["Records"]:
        yield json.loads(record["Sns"]["Message"])


def get_hybrid_objects_from_messages(s3, messages):
    for msg in messages:
        hybrid_record = HybridRecord(**raw_hybrid_record)
        yield hybrid_record, fetch_object_from_s3(s3, hybrid_record)


def get_data_dicts_from_hybrid_objects(hybrid_objects):
    for hybrid_record, s3_obj in hybrid_objects:
        data_dict = s3_object_to_data_dict(s3_object)
        yield data_dict, hybrid_record


if __name__ == "__main__":
    messages = extract_sns_messages_from_event(event)
    hybrid_objects = get_hybrid_objects_from_messages(s3, messages)
    data_dicts = get_data_dicts_from_hybrid_objects(hybrid_objects)
```

Each of the individual generators is much easier to iterate on or test, and then you have reusable generators you can glue together into a different variant of the pipeline later.

harrisonpim: Much neater now, I hope! Thanks for the advice

PR #2933, opened 2018-11-01
Add a script for listing all of our Elasticsearch indexes

alexwlchan:
This adds a small script for listing our Elasticsearch indexes, because we have quite a lot! 85 at the current count.

And we're starting to fill up the storage on our cluster – 75% when I checked yesterday.

It is **not** a tool for deleting indexes. I considered that, but getting the heuristics right felt trickier than I fancied at 8am on a Thursday morning. It just gives you information, you decide what to do with it.

One thing it tells me, for example, is that we have a bunch of indexes that aren't prefixed with a date or version:

```
2018-06-21-alex-testing
better-sierra-ids
credit-awards-legacy-ids-061117
deleteme
items-with-innopac-ids
miro-preproc-2-november-rerun
miro-preproc-monday-rerun
miro-preproc-tuesday-rerun
miro-preproc-wednesday-rerun
miro-reindex
sierra-bibs-title
sierra-creators-with-identifiers
using-credit-not-copyright
works-from-miro-preprocessor
works-from-miro-preprocessor-with-exceptions
```

which has been our practice for nearly six months. I thus propose that these be the first against the wall when the ~revolution~ ~purge~ cleanup comes!

### DIFF
--- a/misc/list_all_es_indexes.py
+++ b/misc/list_all_es_indexes.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python
+# -*- encoding: utf-8
+
+import boto3
+import hcl
+import requests
+
+
+def get_terraform_vars():
+    s3_client = boto3.client("s3")
+    tfvars_body = s3_client.get_object(
+        Bucket="wellcomecollection-platform-infra",
+        Key="terraform.tfvars"
+    )["Body"]
+    return hcl.load(tfvars_body)
+
+
+def build_url(es_credentials):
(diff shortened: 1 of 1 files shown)


kenoir:
What is the motivation for having a script to do this? It's possible to do this easily in Kibana.

alexwlchan:
Yeah, I guess so.

The original idea was the script could also work out which indexes were in use and which weren't, and offer to delete them for you (which Kibana can’t do), then I chickened out of writing that bit!

PR #2939, opened 2018-11-01
Adds migrator lambda

kenoir:

In order to move existing content into the archive service easily, a mechanism to automate ingestion of bagged content (from the `bagger` service) is desirable.

This lambda acts as a bridge between the `bagger` service and the archive ingest api.

### Who is this change for?

👓 🌾


archive/migrator/src/migrator.py
```diff
@@ -0,0 +1,63 @@
...
+def lambda_handler(event, _):
+    ingest_api_url = os.getenv("INGEST_API_URL")
+
+    for record in event["Records"]:
+        message = record["Sns"]["Message"]
+        decoded_message = json.loads(message)
+
+        print(decoded_message)
+
+        if decoded_message["upload_location"]:
```
alexwlchan: Do you want to log something if there isn't an upload_location? Is that an expected/support use case?

archive/migrator/src/migrator.py
```diff
@@ -0,0 +1,63 @@
...
+
+    for message in archive_bag_api_messages(bags, bucket_name):
+        response = session.post(api, json=message)
+        status_code = response.status_code
+
+        if status_code != 201:
+            print_result(f"ERROR calling {api}", response)
+        else:
+            print(f"{message} -> {api} [{status_code}]")
+            location = response.headers.get("Location")
```
alexwlchan: Don't we always expect the Location header to be present? Should we error here if it's absent?

kenoir:
@alexwlchan in both cases I agree, but I want to stop working on this for now though as i've got a few hundred bags archived. We'll need to revisit this for the migration so i don't want to lose the work.

kenoir:
That said, both those comments are useful for a later refactor.

PR #2940, opened 2018-11-01
Use blocking dispatcher in UploadInputStreamFlow

kenoir:

We are seeing the following errors in the archivist service when under load:

"Unable to execute HTTP request: Timeout waiting for connection from pool"

This is symptomatic of all connections made available from the client (in this case via the AWS Client) being used up which is itself likely caused by the number of threads being made available to perform this operation being higher than those available in the AWS Client.

In this case the AWS client builder provides a default connection pool of 50. Unless otherwise specified the default Akka streams dispatcher providing threads does not by default have a cap on the number of threads it will hand out to run streams.

So 50 < infinity, we run out of connections.

This can be mitigated by using a `blocking-dispatcher` which by default allows a maximum number of threads for all streams making use of it.

We did use this dispatcher in all places the archivist uses streams from S3, however in `UploadAndGetChecksumFlow.scala` the dispatcher is attached to `S3UploadFlow` which is reused for all uploads. This is likely resulting in the thread restriction being inappropriately applied at too high a level.

This change moves the blocking dispatcher to `UploadInputStreamFlow.scala` which is created for each upload.

### Who is this change for?

🔬 ⚓️

### DIFF
--- a/archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/bag/BagInfoParser.scala
+++ b/archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/bag/BagInfoParser.scala
@@ -46,6 +46,8 @@ object BagInfoParser {
       extractInternalSenderDescription(bagInfoLines)
     ).mapN(BagInfo.apply)
 
+    inputStream.close()
+
     validated.toEither.leftMap(list => InvalidBagInfo(t, list.toList))
   }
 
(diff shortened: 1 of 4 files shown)


archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/bag/BagInfoParser.scala
```diff
@@ -46,6 +46,8 @@ object BagInfoParser {
       extractInternalSenderDescription(bagInfoLines)
     ).mapN(BagInfo.apply)
 
+    inputStream.close()
```
kenoir: Bonus stream closure.

PR #2945, opened 2018-11-02
Clean up the fields on NotificationMessage

alexwlchan:
This removes all the fields we weren't using, and renames the final field from `Message` to `body`, as that's a more accurate representation of what it means!


kenoir:
Kind of turns this into a value class? This can probably be further refactored away at some point.

alexwlchan:
We still need it for parsing from JSON, so I don't think we can turn it into a value class? But maybe Circe is smarter than I realise.

Issue #2906, opened 2018-10-29
Progress/Registrar services should add `@context` to GET responses


kenoir:
For endpoints: 
- [ ] GET /storage/ingests
- [ ] GET /storage/bags

PR #2948, opened 2018-11-02
Add reference to context in storage service HTTP apps

alicefuzier:
For #2906


archive/archive_api/src/responses.py
```diff
@@ -35,23 +33,12 @@ def __init__(self, response, *args, **kwargs):
         # The @context may already be provided if we've been through the
         # force_type method below.  We also don't add a context if we're
         # looking at the healthcheck endpoint.
-        if ("@context" in rv) or (rv == {"status": "OK"}):
-            return super().__init__(response, *args, **kwargs)
-        else:
-            rv["@context"] = self.context_url
-            json_string = json.dumps(rv)
-            return super().__init__(json_string, *args, **kwargs)
+        return super().__init__(response, *args, **kwargs)
```
alexwlchan: At this point I suspect we can ditch this entire class.

archive/registrar_http/src/test/scala/uk/ac/wellcome/platform/archive/registrar/http/RegistrarHttpFeatureTest.scala
```diff
@@ -103,6 +104,8 @@ class RegistrarHttpFeatureTest
                           "Location"),
                         createdDateString,
                         "Bag") =>
+
+                      actualContectUrl shouldBe "http://api.wellcomecollection.org/storage/v1/context.json"
```
alexwlchan:
```suggestion
                      actualContextUrl shouldBe "http://api.wellcomecollection.org/storage/v1/context.json"
```

archive/registrar_http/src/test/scala/uk/ac/wellcome/platform/archive/registrar/http/RegistrarHttpFeatureTest.scala
```diff
@@ -81,6 +81,7 @@ class RegistrarHttpFeatureTest
 
                   inside(displayBag) {
                     case DisplayBag(
+                      actualContectUrl,
```
alexwlchan:
```suggestion
                      actualContextUrl,
```

archive/registrar_http/src/test/scala/uk/ac/wellcome/platform/archive/registrar/http/RegistrarHttpFeatureTest.scala
alicefuzier: oops

archive/archive_api/src/responses.py
alicefuzier: I think so but I couldn't figure out how to do that and I couldn't be bothered anyway cause all this code is going away entirely. We still have very out of date models in the models folder for example

alicefuzier:
This is deployed and tested both via the archive_api route and the api gateway route and it works fine

PR #2950, opened 2018-11-02
Reporting pipeline miro transforms

harrisonpim:
A bit of initial cleaning and transformation of the miro data. As we're actually starting to apply transforms now, we also replace the `transformer_example` names throughout with `miro_transformer`.

### Who is this change for?
📄📊


reporting/miro_transformer/src/transform.py
```diff
@@ -0,0 +1,83 @@
+import json
+from datetime import datetime
+
+
+def transform(record):
```
alexwlchan: minor: a good name for this variable would be `miro_transformable`, as the type we use in the catalogue transformer is `MiroTransformable`.

reporting/miro_transformer/src/transform.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def clean_dates(data):
+    """
+    elasticsearch's default sort is alphabetical, and sorting dd/mm/yyyy dates
+    alphabetically is rubbish
+    """
+    for key, value in data.items():
+        if "date" in key and value is not None:
+            if type(value) == str:
```
alexwlchan: minor: when comparing types in Python, it's better to use `isinstance` rather than `type()`, because then you cover inheritance as well.
```suggestion
            if isinstance(value, str):
```

reporting/miro_transformer/src/transform.py
```diff
@@ -0,0 +1,83 @@
...
+                data[key] = convert_date_to_iso(value)
+            elif type(value) == list:
```
alexwlchan:
```suggestion
            elif isinstance(value, list):
```

reporting/miro_transformer/src/transform.py
```diff
@@ -0,0 +1,83 @@
...
+                data[key] = [convert_date_to_iso(date) for date in value]
```
alexwlchan: Don’t forget the `else:` branch!

reporting/miro_transformer/src/transform.py
```diff
@@ -0,0 +1,83 @@
...
+    return data
+
+
+def convert_date_to_iso(date_string):
+    if date_string is None or date_string is "":
```
alexwlchan: Be careful here – comparing strings with `is` in Python can return surprising results. `is` compares identity, `==` compares equality. `None` is a singleton so `is` is fine, but strings aren't always the same.
```suggestion
    if date_string is None or date_string == "":
```

or even shorter:

```suggestion
    if not date_string:
```

reporting/miro_transformer/src/transform.py
```diff
@@ -0,0 +1,83 @@
...
+    return transformed
+
+
+def drop_redundant_fields(original_data, keys_to_drop):
+    """
+    strip out any of the original VHS fields which are either empty, sparse,
+    redundantly filled, or just generally unwanted in our ES index
+    """
+    clean_data = {
+        key: value for key, value in original_data.items() if key not in keys_to_drop
```
alexwlchan: Ooh, be careful here.

Your `keys_to_drop` is a list, and list lookups are O(n) – Python has to check every element of the list to see if it compares.

Then this check is O(k * n), where k is the length of `original_data`.

Since you don't care about order or duplicate elements, `keys_to_drop` could be a set instead, and set lookups are O(1) – it uses hashing to identify elements, so lookups are faster.

Make your list a set instead!

reporting/miro_transformer/src/test_data/raw_data.json
```diff
@@ -0,0 +1,253 @@
+{
```
alexwlchan: I should have mentioned this on a previous review, but oh well: we probably want to be a little careful putting raw Miro records here, as some of them contain names/metadata we shouldn't be making public.

harrisonpim:
@alexwlchan checks are now passing and we've run a successful complete reindex here - happy to merge?

alexwlchan:
I’ll have another quick look.

reporting/miro_transformer/src/miro_transformer.py
```diff
@@ -0,0 +1,9 @@
+from reporting_pipeline.hybrid_record_pipeline import process_messages
+from transform import transform
+
+
+def main(event, _, s3_client=None, es_client=None, index=None, doc_type=None):
+    try:
+        process_messages(event, transform, s3_client, es_client, index, doc_type)
+    except Exception as e:
+        print("Error processing MIRO messages:", e)
```
alexwlchan: Errors should never pass silently; this exception should be thrown so it isn’t lost.

reporting/miro_transformer/src/reporting_pipeline/hybrid_record_pipeline.py
```diff
@@ -61,14 +58,17 @@ def unpack_json_from_s3_objects(s3_objects):
         yield hybrid_record_id, json.loads(record)
 
 
-def transform_data_for_es(data):
+def transform_data_for_es(data, transform):
```
alexwlchan: What is this transform parameter meant to be?

reporting/miro_transformer/src/test_miro_pipeline.py
```diff
@@ -0,0 +1,69 @@
...
+                            \"22/04/2017\",
+                            \"19/12/2011\"],
+                \"string_field\": \"a string\",
+                \"null_field\": null,
+                \"single_date\": \"24/04/2004\",
+                \"dict_field\": {
+                      \"a\": \"a_value\"
+                  },
+                \"image_source\": \"ignored field\"
+            }"""
```
alexwlchan: Note, for future reference: an alternative would be `json.dumps(json.dumps(my_dict))`, which might be a little easier to read.

reporting/miro_transformer/src/transform.py
```diff
@@ -0,0 +1,95 @@
...
+keys_to_parse = (
+    "_from",
+    "_to",
+    "_tech",
+    "all_amendment_date",
+    "image_iap_catalog_date",
+    "image_innopac_id_date",
+)
+
+keys_to_drop = (
```
alexwlchan: Close, but no cigar. This is now a tuple, not a set.

Quick guide:

```python
# List. Mutable, ordered, allows duplicates. O(n) lookups.
# Square brackets.
[1, 1, 2, 3]

# Tuple. Immutable, ordered, allows duplicates. O(n) lookups.
# Round brackets.
(1, 1, 2, 3)

# Set. Mutable, unordered, all elements unique. O(1) lookups.
# Curly braces.
{1, 2, 3}
```

You want curly braces (`{}`) not parentheses (`()`).

harrisonpim: 🤦‍♂️

PR #2957, opened 2018-11-05
Add context.json and serve it via api gateway out of s3

alicefuzier:
Various things I'm not sure about and that I would like people' opinion on:

- where to put context.json in our repo. It's in archive/resources/context.json in this PR. Does that sound right?
- What should be the name of the s3 bucket we store it in (or whether we should use an already existing one). In this PR it's a new bucket called `wellcomecollection-assets-archive-static-content`
- Should the static content bucket live in this stack or in the infra critical stack?


alexwlchan:
> where to put context.json in our repo. It's in archive/resources/context.json in this PR. Does that sound right?

Seems fine to me.

The precedent here is the catalogue API, which keeps its context at catalogue_api/api/src/main/resources/context.json. It has to be in `src` because it's served through the Scala app. I think that's fine.

> What should be the name of the s3 bucket we store it in (or whether we should use an already existing one). In this PR it's a new bucket called wellcomecollection-assets-archive-static-content

Not a fan of that name.

For me a bucket called “-assets” contains precious, irretrievable content, and the context definitely isn’t. Additionally the name doesn’t mention the fact that this content is public, which our other buckets do (e.g. wc-data-public, wc-miro-images-public).

I think we should really put the “-public” bit at the front, so how about **wellcomecollection-public-archive-api**?

> Should the static content bucket live in this stack or in the infra critical stack?

As above, this isn’t precious, irretrievable content or a database that would be difficult to rebuild. I think keeping it out of infra critical is best.

PR #2961, opened 2018-11-06
Reduce the size of services in the archive stack

alexwlchan:
💸 💸 💸


kenoir:
👍 I'd try it first though!

alexwlchan:
It’s applied, so let’s see if it breaks anything. :-)

alicefuzier:
Is there a way of reducing it only for some of them? I think the archivist needs loads of resurces, but only the archivist

PR #2972, opened 2018-11-06
Status values as per RFC

alicefuzier:

### DIFF
--- a/archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/progress/models/Callback.scala
+++ b/archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/progress/models/Callback.scala
@@ -33,12 +33,4 @@ case object Callback extends URIConverters {
   case object Failed extends CallbackStatus {
     override def toString: String = failureString
   }
-
-  def parseStatus(string: String): CallbackStatus = {
-    string match {
-      case `processingString`   => Pending
-      case `successString` => Succeeded
-      case `failureString`    => Failed
-    }
-  }
 }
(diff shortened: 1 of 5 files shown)


archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/progress/models/Callback.scala
```diff
@@ -18,27 +18,19 @@ case object Callback extends URIConverters {
...
-  val pendingString = "pending"
-  val succeededString = "succeeded"
-  val failedString = "failed"
+  val processingString = "processing"
+  val successString = "success"
+  val failureString = "failure"
 
   case object Pending extends CallbackStatus {
-    override def toString: String = pendingString
+    override def toString: String = processingString
```
mklander: Maybe rename the class to `Processing` to match the string?  Personally I prefer `pending` for the callback status, because processing leads me to think it is waiting for a response.

archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/progress/models/Callback.scala
```diff
@@ -18,27 +18,19 @@ case object Callback extends URIConverters {
...
   }
 
   case object Succeeded extends CallbackStatus {
-    override def toString: String = succeededString
+    override def toString: String = successString
```
mklander: As with `pending`, I think it's a bit confusing that the string doesn't match the status.

alicefuzier: I didn't rename them out of laziness to be honest, because that's a refactoring change that affects all apps potentially. But one nice side effect of not renaming the scala object is that the stuff that it's currently stored in dynamo is still readable because scanamo serialises using the object name, not the toString method.
I'm tempted to leave them as is, I'm not sure the strings should explicitly match the object name given that they are more external serialisation than anything else.

Issue #2967, opened 2018-11-06
Archive services config cleanup 


kenoir:
The Archive services can use a simpler configuration strategy.

See https://github.com/wellcometrust/platform/pull/2745 for a reference implementation.

PR #2980, opened 2018-11-07
More Progress on squeezing the Guice out of our archive stack

alexwlchan:
Part of #2967. Follows #2978.


archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/config/models/OldHttpServerConfig.scala
```diff
@@ -1,7 +1,8 @@
 package uk.ac.wellcome.platform.archive.common.config.models
+
 import java.net.URL
 
-case class HttpServerConfig(
+case class OldHttpServerConfig(
```
alicefuzier: I'm sure there's a reason but why this?

alexwlchan: Aha, this is a bit of a fudge.

The new HTTPServerConfig doesn’t include the ContextURL, because I wanted to be able to configure that separately (in the notifier). But Scala doesn’t let you have classes with the same name that differ only in case, as they break on case-insensitive filenames.

I renamed this temporarily, but it eventually goes away when I migrated whichever other app was using this old case class.

PR #2983, opened 2018-11-08
Reporting pipeline miro transforms

harrisonpim:
A few small changes before starting work on the sierra transformer
- removes `reporting_pipeline` subdir - this stuff is now importable from the [wellcome_aws_utils package](https://github.com/wellcometrust/aws_utils)
- splits tests by theme
- makes `keys_to_parse` an explicit set, rather than a confusing list of conditions
- ignores, rather than drops, badly formatted dates

### Who is this change for?
📄 📊


reporting/miro_transformer/src/transform.py
```diff
@@ -26,7 +26,7 @@ def clean_dates(data, keys_to_parse):
     alphabetically is rubbish
     """
     for key, value in data.items():
-        if value_should_be_parsed(key, value, keys_to_parse):
+        if any(key == substring for substring in keys_to_parse):
```
alexwlchan: I think the following is equivalent, and simpler:
```suggestion
        if key in keys_to_parse:
```

PR #2993, opened 2018-11-08
Finatra was the name of the monster, not the doctor

alexwlchan:
So having removed all the Guice and dependency injection from the archive apps, why not try this in the rest of the repository? I give you… a Finatra-free bib merger.

Still needs a bit of work to get tests passing, etc, and I have to go home now, but how’s this as a first pass?


kenoir:
Can't find a gif of the monster waving as he disappears off into the distance on an iceberg, so you'll just have to imagine it.

Also the monster has the face of Nic Cage.

PR #2998, opened 2018-11-09
[WIP] Have the reindexer emit an instance of VHSIndexEntry instead of HybridRecord

alexwlchan:
This is a change to the way the reindexer works. It’s also still quite broken (I’m fighting implicit encoders in Circe, yay), so skip the code review for now.

Here’s the intended behaviour:

* If the table doesn’t have any extra metadata (e.g. the Sierra or Miro migration VHS), just send the raw HybridRecord
* If the table does have any metadata, send that in a VHSIndexRecord

Thoughts?

### DIFF
--- a/reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex/reindex_worker/services/ReindexWorker.scala
+++ b/reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex/reindex_worker/services/ReindexWorker.scala
@@ -1,6 +1,6 @@
 package uk.ac.wellcome.platform.reindex.reindex_worker.services
 
-import akka.actor.ActorSystem
+import akka.actor.{ActorSystem, Terminated}
 import com.google.inject.Inject
 import com.twitter.app.Flag
 import uk.ac.wellcome.messaging.sns.NotificationMessage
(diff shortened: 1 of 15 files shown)


reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex/reindex_worker/services/ReindexWorker.scala
```diff
@@ -1,31 +1,109 @@
...
+    */
+  private def processEmptyMetadata(
+    items: List[util.Map[String, AttributeValue]]): Future[Unit] =
+    bulkSNSWriter
+      .sendToSNS(items.map { av =>
+        // AWLC: if you're changing this code to create an instance of
+        // `EmptyMetadata` to include in a `VHSIndexEntry`, don't bother
+        // parsing as a case class; just create a fresh value.
+        //
+        // When I tried to call `parseAsCaseClass[EmptyMetadata]`, I got
```
alexwlchan: Note to self: this turned out to be caused by a Circe problem (specifically, a null decoder was being passed). Revisit this?

kenoir:
Let's have a whiteboard chat about this.

PR #3003, opened 2018-11-11
Ye gods, he’s still at it

alexwlchan:
One more for tonight. Follows #3002, part of #2967.

🔻🔺🔻


alexwlchan:
This one mostly works, but it relies on the Jackson object mapper used by the API, and configuring that is fiddly. I’m going to leave this one for now, and revisit it if/when I look at the API.

PR #3014, opened 2018-11-13
[WIP] Update to use the latest version of api-gw modules

kenoir:

Deployable Catalogue API being API Gateway.

### Who is this change for?

🍶


archive/terraform/storage_api/gateway.tf
```diff
@@ -0,0 +1 @@
+
```
alicefuzier: ?

Issue #3006, opened 2018-11-12
Add swagger docs to storage service


alicefuzier:
We want swagger docs to be available for the storage service and we would like them to be automagically generated in code so that model and endpoint descriptions are more likely to be kept up to date.

The main problem is that the endpoints exposed by storage service (ingests and bags) are served by two separate services (progress_http and registrar_http respectively). We can easily generate swagger docs for each of the services but we have to figure out a way to munge them together into a cohesive doc that includes both. Specifically, we need a swagger documentation that:
- includes endpoints for both services with the paths as exposed by API Gateway
- documents the data model. The data model exposed by the generated docs for each service will differ but they will partially overlap.

A proposed solution is to:
- expose a automagically generated swagger.json endpoint in each app. 
- add a hardcoded swagger.json to API Gateway at https://api.wellcomecollectcion.org/storage/v1/swagger.json that looks approximately like this:
```
{
  "swagger": "2.0",
  "info": {
    "description": "Storage",
    "version": "v1",
    "title": "Catalogue"
  },
  "host": "api.wellcomecollection.org",
  "schemes": [
    "https"
  ],
  "produces": [
    "application/json",
    "application/ld+json"
  ],
  "paths": {
    "/storage/v1/bags/{id}": {
      "$ref": "https://api.wellcomecollection.org/storage/v1/bags/swagger.json#/paths/registrar"
    },
    "/storage/v1/ingests/{id}": {
      "$ref": "https://api.wellcomecollection.org/storage/v1/ingests/swagger.json#/paths/progress"
    }
  }
}
``` 
This leverages the use of [Json Reference](https://swagger.io/docs/specification/using-ref/) to bring into the main storage swagger.json the bits of each service swagger.json docs that are relevant to it.

The main advantages of this approach are that:
- it keeps the documentation of things likely to change (like the data model or request parameters) close to the code 
- the above hardcoded JSON is not likely to change very often

Disadvantages are that:
- to resolve the full swagger.json a user has to make 3 HTTP requests. Nota a major issue
- the internal swagger endpoints (https://api.wellcomecollection.org/storage/v1/bags/swagger.json) need to be publicly resolvable which means that the separation in two services will be visible to an external user. Not nice but not dangerous 😐 
- Some shared model classes will be repeated in both internal swagger endpoints. Not nice but also probably not particularly problematic, especially because the bits of the model that are shared are defined in a shared module in scala, so the exposed documentation will effectively always be the same 😐 

[...]

PR #3018, opened 2018-11-13
[WIP] Add swagger docs to storage service

alicefuzier:
For #3006


alicefuzier:
Turns out this approach doesn't work

Issue #2984, opened 2018-11-08
Create a VHS with all Miro data, not just stuff in the Catalogue API


alexwlchan:

alexwlchan:
The magic number we're waiting for in vhs-miro-migration is 421,608.

alexwlchan:
![screen shot 2018-11-16 at 07 45 53](https://user-images.githubusercontent.com/301220/48605072-a85d5580-e973-11e8-86d9-237227a5fea6.png)

🎉

PR #3023, opened 2018-11-14
Have the reindexer emit the entire row as a JSON string, not just case classes

alexwlchan:
This is a required piece for #2984 – we can now put metadata in the VHS and have it be available to the downstream applications via the reindexer.

This returns a flat structure, i.e.

```json
{"id": "...", "location": "...", "version": "...", "miroMetadata: "..."}
```

Do we particular care about that vs nested results? (i.e. VHSEntry replicas)

Closes #2998.


alicefuzier:
> Do we particular care about that vs nested results? (i.e. VHSEntry replicas)
I think it depends on what's going to read them, doesn't it? I guess we can create a case class representing the flat structurein the miro transformer

alexwlchan:
Yep. Or you can just parse the case class with Circe twice – once as HybridRecord, once as MiroMetadata.

Issue #2964, opened 2018-11-06
Annotate S3 objects with checksums


kenoir:
The archivist should annotate S3 objects with their checksum for future proofing.

mklander:
@jtweed confirmed that this should be added as object metadata rather than tags and a header key of `x-amz-meta-sha-256` is acceptable, i.e., `x-amz-meta-$digest-algorithm`

Issue #2782, opened 2018-10-11
Annotate S3 objects with checksums


jtweed:

kenoir:
Closing as dupe of https://github.com/wellcometrust/platform/issues/2964

PR #3027, opened 2018-11-14
Add digests to S3 objects

mklander:

Add the sha-256 digest as userMetadata when uploading to S3

part of #2782 #2964

### DIFF
--- a/archive/archivist/src/main/scala/uk/ac/wellcome/platform/archive/archivist/flow/S3UploadFlow.scala
+++ b/archive/archivist/src/main/scala/uk/ac/wellcome/platform/archive/archivist/flow/S3UploadFlow.scala
@@ -7,7 +7,7 @@ import akka.stream._
 import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
 import akka.util.ByteString
 import com.amazonaws.services.s3.AmazonS3
-import com.amazonaws.services.s3.model.{ObjectMetadata => S3ObjectMetadata, _}
+import com.amazonaws.services.s3.model._
 import grizzled.slf4j.Logging
 import uk.ac.wellcome.platform.archive.archivist.models.storage.ObjectMetadata
 import uk.ac.wellcome.storage.ObjectLocation
@@ -246,7 +246,7 @@ class S3UploadFlow(
             new InitiateMultipartUploadRequest(
               uploadLocation.namespace,
               uploadLocation.key,
-              toS3ObjectMetadata(objectMetadata)
+              objectMetadata.toS3ObjectMetadata
             )
         }
         Try(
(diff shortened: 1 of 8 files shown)


archive/archivist/src/main/scala/uk/ac/wellcome/platform/archive/archivist/models/storage/ObjectMetadata.scala
```diff
@@ -0,0 +1,3 @@
+package uk.ac.wellcome.platform.archive.archivist.models.storage
+
+case class ObjectMetadata(userMetaData: Map[String, String])
```
alexwlchan: nit: let's be consistent with capitalisation.
```suggestion
case class ObjectMetadata(userMetadata: Map[String, String])
```

archive/archivist/src/main/scala/uk/ac/wellcome/platform/archive/archivist/flow/S3UploadFlow.scala
```diff
@@ -7,20 +7,25 @@ import akka.stream._
...
 
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 import scala.util.{Failure, Success, Try}
 
 object S3UploadFlow extends Logging {
-  def apply(uploadLocation: ObjectLocation)(implicit s3Client: AmazonS3) =
-    new S3UploadFlow(uploadLocation)(s3Client)
+  def apply(uploadLocation: ObjectLocation,
+            maybeUploadMetadata: Option[ObjectMetadata] = None)(
```
alexwlchan: minor: I wonder if it would be cleaner to make this a parameter that defaults to an empty map? Something like:

```suggestion
            uploadMetadata: ObjectMetadata = ObjectMetadata(Map.empty))(
```

The `toS3ObjectMetadata` method should work fine (you just get an empty metadata blob), and it reduces the number of Options you have to pass around.

Actually, what are the scenarios when we wouldn't want metadata? (Aren't we always annotating uploads with hashes?)

mklander: The scenario where we don't want metadata is uploading the tagManifest does not include a checksum and hence no metadata.  We could default to an empty map.

alicefuzier: I also think the SUploadFlow is fairly generic and we should eventually move into scala-storage, so I wouldn't enforce metadata

mklander: yup, thanks @alicefuzier.

archive/archivist/src/main/scala/uk/ac/wellcome/platform/archive/archivist/flow/S3UploadFlow.scala
```diff
@@ -224,21 +230,39 @@ class S3UploadFlow(uploadLocation: ObjectLocation)(implicit s3Client: AmazonS3)
...
+            )
+        }
         Try(
           s3Client
             .initiateMultipartUpload(initiateRequest)
             .getUploadId
         )
       }
+
+      private def toS3ObjectMetadata(
```
alicefuzier: perhaps this should live in the `ObjectMetadata` class?

PR #3028, opened 2018-11-14
Reporting pipeline sierra

harrisonpim:
adds sierra bibs to the reporting pipeline

### Who is this change for?
📄 📊


reporting/sierra_transformer/src/transform.py
```diff
@@ -0,0 +1,59 @@
...
+
+    # unpack material types
+    if "materialType" in bib_record:
+        bib_record["materialType"] = bib_record["materialType"]["code"]
+
+    # unpack orders
+    if "orders" in bib_record:
+        for order in bib_record["orders"]:
+            bib_record["order_locations"] = order["location"]
+            bib_record["order_dates"] = order["date"]
```
alexwlchan: I might be reading this wrong (it’s late), but won’t this discard all but the first order?

You repeatedly overwrite the same `order_{locations, dates}` key, so only the last order will be preserved.

reporting/terraform/main.tf
```diff
@@ -30,6 +31,26 @@ module "lambda_miro_inventory_transformer" {
...
 
+  vhs_read_policy       = "${local.miro_inventory_vhs_read_policy}"
   trigger_topic_arn     = "${local.miro_inventory_topic_arn}"
   error_alarm_topic_arn = "${local.lambda_error_alarm_arn}"
 }
+
+module "lambda_sierra_transformer" {
+  source = "./reporting_lambda"
+
+  name        = "sierra_transformer"
```
alexwlchan: Minor: I'd like us to be better about namespacing our applications. We already have a “sierra_transformer” in the catalogue pipeline – maybe all these Lambdas could have their name prefixed with “reporting-”, or something?

(I checked the module, but couldn’t see anything doing this.)

mklander: 👍yes we also talked about this

harrisonpim:
closing this PR, as its contents are now encompassed by #3110

PR #3029, opened 2018-11-14
“What do you mean, even smaller?”

alexwlchan:
Despite turning down the memory assigned to each of our Scala apps, they're still wasting memory. I looked at the metrics, and they're barely using 30% of what we're giving them. WASTEFUL.

This turns them down another notch, although I think any more than this and we'll start having out-of-memory errors.


alexwlchan:
These changes are applied and will affect tomorrow morning’s Sierra runs.

PR #3045, opened 2018-11-16
Yes, there is plenty more red to squeeze out of our Typesafe config

alexwlchan:
This patch introduces a new trait `WellcomeApp`, which is extended by the `Main` object in all of our apps. It cleans up a bit of boilerplate and repetitive code, and makes simpler Main objects in each app.

It also removes a couple of now-defunct stop() methods, and fixes the Sierra transformer.


sbt_common/common/src/main/scala/uk/ac/wellcome/WellcomeApp.scala
```diff
@@ -0,0 +1,35 @@
...
+
+ import scala.concurrent.duration.Duration
+ import scala.concurrent.{Await, Future}
+
+ trait WorkerService {
+   def run(): Future[Done]
+ }
+
+ trait WellcomeApp extends App with Logging {
+   def buildWorkerService(config: Config): WorkerService
```
alicefuzier: can we avoid overriding this function and have `run` take a `WorkerService`?

alexwlchan: I was trying to ditch having this line in every app:

```scala
val config: Config = ConfigFactory.load()
```

How about if `run()` looks a bit like this:

```scala
def run(buildWorkerService: (config: Config) => WorkerService) {
    val config: Config = ConfigFactory.load()
    val workerService = buildWorkerService(config)
    workerService.run()
}
```

Would that work for you?

alicefuzier: Yeah ok. 
`config` is never used in `WellcomeApp` other than to build the `WorkerService ` so it feels to me like it shouldn't be the concern of `WellcomeApp` to instantiate it. The fact that it does feels a bit artificial and I don't think it's quite in the right place. 
But it's better than overriding, I would be happy to go with that for now

alexwlchan:
And hey, we just kicked Scalaz and cats out of our common classpath. 😜

catalogue_pipeline/id_minter/src/universal/conf/application.conf.template
```diff
@@ -2,9 +2,9 @@ aws.rds.host="${cluster_url}"
 aws.rds.identifiers.database="identifiers"
 aws.rds.identifiers.table="identifiers"
 aws.rds.password="${db_password}"
-aws.rds.port="${db_port}"
+aws.rds.port=${db_port}
```
alicefuzier: ?

alicefuzier:
I think the dependencies changes might be too restrictive: registrar_http, progress_http and notifier use circe optics in the test assertions, I believe. Also, archivist and registrar need cats definitely, not sure if that comes from somewhere else

kenoir:
Yo @alexwlchan, what does this need to be closed?

kenoir:
Closing as abandoned.

PR #3058, opened 2018-11-20
Make sierra transformer (and others) run

alicefuzier:


alicefuzier:
Not sure why this doesn't show as gree cause it is green in travis. I'm going to merge it as it is deployed and puts the pipeline in a slightly less broken state

PR #3072, opened 2018-11-26
Move to new storage stack

kenoir:

Remove previous storage service infra, move to new stack with more sensible split of resources.

This should enable us to bring up further parallel storage services when required for testing without disturbing the live service.

### Who is this change for?

✂️


kenoir:
> Plan: 1 to add, 0 to change, 212 to destroy.

archive/terraform/ecr.tf
```diff
@@ -0,0 +1,48 @@
+data "aws_ecr_repository" "ecr_repository_nginx_services" {
+  name = "uk.ac.wellcome/nginx_services"
+}
+
+module "ecr_repository_archivist" {
+  source = "git::https://github.com/wellcometrust/terraform.git//ecr?ref=v1.0.0"
+  name   = "archivist"
+}
```
mklander: Is it intentional to lose `uk.ac.wellcome` in the names?

archive/terraform/iam.tf
```diff
@@ -0,0 +1,125 @@
...
+  statement {
+    effect = "Allow"
+
+    principals {
+      type = "AWS"
+
+      identifiers = [
+        "arn:aws:iam::653428163053:user/echo-fs",
+        "arn:aws:iam::653428163053:user/api",
+      ]
```
mklander: Can these ids be in variables in the private vars?

archive/terraform/main.tf
```diff
@@ -2,7 +2,7 @@
 
 resource "aws_api_gateway_base_path_mapping" "api-gw-mapping-241118" {
   api_id      = "${module.stack-241118.api_gateway_id}"
-  domain_name = "api-stage.wellcomecollection.org"
+  domain_name = "stage.wellcomecollection.org"
   base_path   = "storage"
```
mklander: I think I preferred `api-stage` & `api` rather than `stage` & `api`, but I expect there is a reason this is now different.

archive/terraform/ecr.tf
kenoir: I think V1 of the modules prepends that for you.

archive/terraform/iam.tf
kenoir: Bit late now! Seriously I don't think these are dangerous to have in the wild?

Issue #3038, opened 2018-11-16
Sort out the Miro VHS split brain


alexwlchan:
Follows #2984.

A brief note: the current state of the world is that we have two VHS cassettes for Miro:

* vhs-sourcedata-miro contains all the records cleared for the Catalogue API, but the S3 objects are that weird wrapper
* vhs-miro-complete contains every Miro record, but the S3 objects are the raw Miro objects

The long-term goal is to have a single VHS cassette containing all the Miro records, with the S3 objects being raw Miro metadata, and some metadata in DynamoDB that tells us whether it's cleared for the Catalogue API.

Remaining tasks:

- [x] Add a boolean field "isClearedForCatalogueAPI" to the `vhs-miro-complete` table (populated based on the vhs-sourcedata-miro table)
- [x] Modify the transformer to read this table, and only transform records with "isClearedForCatalogueAPI: true"
- [ ] Modify the reindexer to support FilterExpression (you can do this by adding withFilterExpression to the ScanSpec)
- [x] Copy all the data from vhs-miro-complete into vhs-sourcedata-miro
- [x] Delete vhs-miro-complete

alexwlchan:
The last item is its own ticket: #3120.

PR #3081, opened 2018-11-27
[WIP] Miro transformer can read records from the unified VHS

alexwlchan:
This is a rather chunky patch for the Miro transformer. The key change is that it reads from a VHS whose entries are the Miro source data, rather than the weird MiroTransformable wrapper we created.

Unfortunately that wrapper is everywhere, so getting it out is rather tricky!

I might break this into multiple patches.

Part of #3038.


alexwlchan:
Closing in favour of #3084.

PR #3085, opened 2018-11-27
Receive HybridRecord and MiroMetadata from the VHS, not a MiroTransformable

alexwlchan:
This allows the Miro transformer to read from the new VHS.

Once merged, I'll deploy it and run a small reindex before exposing it to the entire data set.

Follows #3084. Part of #3038.


kenoir:
LGTM, do we still have notion of a transformable? I'm not sure it makes sense now.

alexwlchan:
> do we still have notion of a transformable?

Not really. The sealed trait got ditched a while back, and `SierraTransformable` has hung around because:

1. Nobody has given it a better name
2. It doesn't really map to anything in Sierra, so we can't call it a "record" or anything like that.

alexwlchan:
I’m treating Robert’s “LGTM” as implicit approval. Merge!

PR #3087, opened 2018-11-27
Change the values of ingest/callback status

### LINKED ISSUE #3074: Changes to ingest/callback status
The current values for `Status` on `Ingest` and `Callback` are:

- `initialised`, `processing`, `success`, `failure`

To improve consistency of language, these should be changed to:

- `accepted`, `processing`, `succeeded`, `failed`

alexwlchan:
Resolves #3074.


archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/progress/models/Progress.scala
```diff
@@ -19,25 +19,25 @@ case class Progress(id: UUID,
...
   private val processingString = "processing"
-  private val successString = "success"
-  private val failureString = "failure"
+  private val succeededString = "succeeded"
+  private val failedString = "failed"
 
-  case object Initialised extends Status {
-    override def toString: String = initialisedString
+  case object Accepted extends Status {
+    override def toString: String = acceptedString
```
mklander: Will this break when deserialising existing stored progress?  As it is progress maybe that's fine...

alexwlchan: Yes.

I thought the database was only test data, so that was safe, but perhaps I've missed something?

mklander: I think then something to be aware of.

PR #3094, opened 2018-11-28
Refactor the reindexer core to pass around table name/topic ARN

alexwlchan:
This is the first step towards a reindexer wish list item: being able to pass the table name/topic ARN in the message sent to the reindexer.

It doesn't change the reindexer interface (except renaming "ReindexJob" to "ReindexParameters"), but the internal classes now expect to get DynamoDB/SNS config with each message, rather than configured once upfront and never changed. Later we can pass the parameters directly from the queue message into the classes.

### DIFF
--- a/reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex/reindex_worker/Main.scala
+++ b/reindexer/reindex_worker/src/main/scala/uk/ac/wellcome/platform/reindex/reindex_worker/Main.scala
@@ -43,7 +43,7 @@ object Main extends App with Logging {
   )
 
   val hybridRecordSender = new BulkSNSSender(
-    snsWriter = SNSBuilder.buildSNSWriter(config)
+    snsMessageWriter = SNSBuilder.buildSNSMessageWriter(config)
   )
 
   val workerService = new ReindexWorkerService(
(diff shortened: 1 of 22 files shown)


mklander:
I wonder if we should whitelist the tables and topics that the reindexer can send to?

alexwlchan:
@mklander I was thinking we could do that with IAM permissions?

We can whitelist the VHS tables, and allow pushing to any SNS topic whose name is of the form `reindexer_*`.

mklander:
Using IAM is a good idea, perhaps we could do both?

alexwlchan:
Yep, no reason not to. :-)

kenoir:
I'd be tempted to have an id/arn mapping, that way we dont explicitly pass around arns in messages and the whitelist is implicit from the available mappings. 

In this situation all reindexers would get config for all possible topic/tables, but the message sent contains a ref that lets you pull the correct ones out of the config.

alexwlchan:
Yeah, and then the script just sends that ID.

We could very easily do something like that.

I like it!

On 28 Nov 2018, at 14:57, Robert Kenny <notifications@github.com<mailto:notifications@github.com>> wrote:


I'd be tempted to have an id/arn mapping, that way we dont explicitly pass around arns in messages and the whitelist is implicit from the available mappings.

In this situation all reindexers would get config for all possible topic/tables, but the message sent contains a ref that lets you pull the correct ones out of the config.

PR #3095, opened 2018-11-28
Ditch the duplication between StorageSpace/Namespace

alexwlchan:
We had two internal models, both of which wrap a single underlying val `String`, and which get mapped to `DisplayStorageSpace` in a slightly ugly way. I spotted it while pulling a different thread, and it became a scandalone thing.

tl;dr: Slightly cleaning up the archive models.


alexwlchan:
Closing due to inactivity.

PR #3096, opened 2018-11-28
Only create an instance of Progress once

alexwlchan:
This is the strand I was originally pulling on in #3095.

We have a test somewhere that creates an instance of Progress, then asserts the created and modified dates are the same. Which is fine… until `Instant.now` is called over a millisecond boundary, and the test fails. Really hard to hit, but caused me a flaky failure on master. So this fixes that.


alexwlchan:
Closing due to inactivity.

Issue #3046, opened 2018-11-16
Subjects/genres/contributors normalisation


jtweed:
For all data sources, on `Person`, `Organisation`, `Agent`, `Concept`, `Place`, `Period`, `Subject` and `Genre`:

* Remove trailing punctuation from `label` 

For Miro, on `Concept`, `Subject` and `Genre`:

* Normalise `label` to sentence case

mklander:
Is punctuation to include just full stops and commas?  What about abbreviations and contractions, for example: `...` `&Co.` `B.C.` `etc.`

jtweed:
on person remove trailing `.`
on period remove trailing `.`
on place remove trailing ` :`
on concept remove trailing `.`

mklander:
Objects can be created, such as Organisation/Subject/Genre as a compound of multiple Marc/Miro fields.  In these cases which rules should apply to each part of the compound label?

For example 610 bib data creates an Organisation with a label from MarcSubfields a & b.

A Miro subject label is created from multiple keywords from `image_keywords` and/or `image_keywords_unauth`.

jtweed:
I think for Organisation, we should combine and then remove trailing punctuation from the combined label.

For subjects, we should normalise each part, then combine.

jtweed:
Nothing for Sierra, sentence case for Miro

PR #3097, opened 2018-11-28
Normalise label text

mklander:

Normalise label text by stripping punctuation.

part of #3046


kenoir:
, 💥

catalogue_pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraAgents.scala
```diff
@@ -27,11 +27,18 @@ trait SierraAgents {
       val prefixString =
         if (prefixes.isEmpty) None else Some(prefixes.mkString(" "))
 
-      Person(
-        label = label,
-        prefix = prefixString,
-        numeration = numeration
-      )
+      if (Set("100", "700").contains(marcTag))
```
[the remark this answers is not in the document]
alexwlchan: I would rather pass in a parameter `normalisePerson: Boolean` than the MARC tag, that defaults to false – otherwise we have to know about what a particular MARC tag means in multiple places.

catalogue_pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/SierraContributors.scala
```diff
@@ -48,7 +48,7 @@ trait SierraContributors extends MarcUtils with SierraAgents {
     persons
       .flatMap { subfields: List[MarcSubfield] =>
         val roles = getContributionRoles(subfields)
-        val maybePerson = getPerson(subfields)
+        val maybePerson = getPerson(marcTag, subfields)
```
alexwlchan: Following the above, here you'd have

```scala
val maybePerson = getPerson(subfields, normalisePerson = true)
```

catalogue_pipeline/transformer/transformer_sierra/src/main/scala/uk/ac/wellcome/platform/transformer/sierra/transformers/subjects/SierraPersonSubjects.scala
```diff
@@ -44,7 +44,7 @@ trait SierraPersonSubjects extends MarcUtils with SierraAgents {
       .filterNot { _.indicator2.contains("7") }
       .flatMap { varField: VarField =>
         val subfields = varField.subfields
-        val maybePerson = getPerson(subfields)
+        val maybePerson = getPerson("600", subfields)
```
alexwlchan: And with the flag I suggested, you could leave this code as-is.

sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
```diff
@@ -1,28 +1,65 @@
...
+import uk.ac.wellcome.models.work.text.TextNormalisation._
+
 sealed trait AbstractRootConcept {
   val label: String
 }
 
 sealed trait AbstractConcept extends AbstractRootConcept
 
 case class Concept(label: String) extends AbstractConcept
+object Concept {
```
alexwlchan: minor: elsewhere we've used `case object` to highlight that these are companion objects, although I confess I'm not entirely sure what the distinction means!

sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
```diff
@@ -1,28 +1,65 @@
...
+object Period {
+  def normalised(label: String): Period = {
+    Period(trimTrailing(label, '.'))
+  }
+}
 
 case class Place(label: String) extends AbstractConcept
+object Place {
+  def normalised(label: String): Place = {
+    Place(trimTrailing(label, ':'))
```
alexwlchan: Is there a spec or ticket that explains what normalisations we're meant to be doing? A comment linking to that would be helpful.

sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/text/TextNormalisation.scala
```diff
@@ -0,0 +1,9 @@
+package uk.ac.wellcome.models.work.text
+
+object TextNormalisation {
+  def trimTrailing(s: String, c: Char): String =
+    s.replaceAll(s"[\\s]*[$c][\\s]*$$", "")
```
alexwlchan: I'd like a brief comment to explain this regex, or better no regex at all! The `$$` is especially odd.

Or possibly something like:

```scala
s.trim.stripSuffix(s).trim
```

?

sbt_common/internal_model/src/test/scala/uk/ac/wellcome/models/work/internal/text/textNormalisationTest.scala
```diff
@@ -0,0 +1,46 @@
...
+    }
+  }
+
+  it("converts to sentence case") {
+    val examples = Table(
+      ("in", "out"),
+      ("text", "Text"),
+      ("TEXT", "TEXT"),
+      ("teXT", "TeXT"),
+      ("text text", "Text text")
```
alexwlchan: Other test cases that might be worth considering:

```
Text Text
Text teXT
```

sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/internal/AbstractRootConcept.scala
alicefuzier: I don't think case is useful for object. The only effect it has is addin methods like copy, equals, hash and so on. It makes little sense for objects and I generally prefer to leave it out

sbt_common/internal_model/src/main/scala/uk/ac/wellcome/models/work/text/TextNormalisation.scala
mklander: My understanding is that `s.trim` removes whitespace at the beginning and end which is not the same.  I don't believe there is an `rtrim`.  I considered breaking it down, but that means multiple regexp calls.

alexwlchan: Ack, yes. Okay, a comment explaining what the regex is meant to do then.

mklander: 👍

PR #3109, opened 2018-11-30
Fixes for SNS fixtures

alexwlchan:
We often want to get messages from SNS, then decode them as a case class – this cleans up some repeated code for doing that.


alexwlchan:
Closing due to inactivity.

PR #3110, opened 2018-11-30
New reporting sources

harrisonpim:
adds further reporting channels from source data &rarr; kibana
- sierra
- calm
- switches to the new-format miro VHS

### Who is this change for?
📄 📊


reporting/reporting_calm_transformer/src/reporting_calm_transformer.py
```diff
@@ -0,0 +1,33 @@
+"""
+Basic transformer, which cleans up the static calm data before sending it off
+to an elasticsearch indesx. The raw data can be downloaded by running:
+
```
mklander: index ? spelling!

reporting/reporting_calm_transformer/src/reporting_calm_transformer.py
```diff
@@ -0,0 +1,33 @@
...
+import pandas as pd
+from transform import transform
+from tqdm import tqdm_notebook as tqdm
+from elasticsearch import Elasticsearch
+
+es_username = ""
+es_password = ""
+es_url = ""
+path_to_calm_json = ""
+
```
mklander: maybe take these from the command line or load from a local . file?

reporting/reporting_calm_transformer/src/test_date_transforms.py
```diff
@@ -0,0 +1,21 @@
+from transform import transform
+
+
+def test_cleans_simple_date():
+    test_dict = {"UserDate1": "01/01/1970"}
+    assert transform(test_dict) == {"UserDate1": "1970-01-01"}
+
```
mklander: make the month and day different so this tests the formatting is as you expect.

reporting/reporting_calm_transformer/src/test_transform.py
```diff
@@ -0,0 +1,29 @@
...
+
+def raw_data():
+    return {
+        "quote_before": "'some text",
+        "quote_after": "some text'",
+        "quote_both": "'some text'",
+        "single_element_list": ["single list item"],
+        "nan_field": math.nan,
+    }
+
```
mklander: maybe clearer to put these cases in the respective tests.

reporting/terraform/main.tf
```diff
@@ -30,6 +31,26 @@ module "lambda_miro_inventory_transformer" {
...
+  source = "./reporting_lambda"
+
+  name        = "reporting_sierra_transformer"
+  description = "Transform sierra source data and send to ES."
+
+  environment_variables = {
+    ES_URL      = "${var.reporting_es_url}"
+    ES_USER     = "${var.reporting_es_user}"
+    ES_PASS     = "${var.reporting_es_pass}"
+    ES_INDEX    = "sierra"
```
mklander: I think we should move this to a more specific index and include the creation data to avoid confusion and name clashes.

reporting/reporting_calm_transformer/src/reporting_calm_transformer.py
```diff
@@ -0,0 +1,33 @@
+"""
+Basic transformer, which cleans up the static calm data before sending it off
+to an elasticsearch indesx. The raw data can be downloaded by running:
```
alexwlchan:
```suggestion
to an Elasticsearch index. The raw data can be downloaded by running:
```

reporting/reporting_calm_transformer/src/reporting_calm_transformer.py
```diff
@@ -0,0 +1,33 @@
...
+
+    python monitoring/scripts/download_oai_harvest.py
+
+from the root of this repo. The `path_to_calm_json` parameter below should be
+set to the relative path to that file from this directory. The `es_username`,
```
alexwlchan: I'm not sure I understand what this parameter should be. It's probably also not necessary – you can work it out.

```python
# Path to the root of the Git repo
subprocess.check_call([
    "git", "rev-parse", "--show-toplevel"]).strip().decode("utf8")

# Path to the current file
os.path.abspath(__file__)
```

reporting/reporting_calm_transformer/src/reporting_calm_transformer.py
```diff
@@ -0,0 +1,33 @@
...
+
+es_username = ""
+es_password = ""
+es_url = ""
+path_to_calm_json = ""
+
+df = pd.read_json(path_to_calm_json)
+es = Elasticsearch(es_url, http_auth=(es_username, es_password))
+
+for idx in tqdm(range(len(df))):
```
alexwlchan: Would it be neater to iterate over the records directly? A quick Google suggests something like:

```python
for record in tqdm(df.iterrows(), total=len(df))
```

reporting/reporting_calm_transformer/src/test_date_transforms.py
```diff
@@ -0,0 +1,21 @@
...
+
+def test_handles_empty_datestring():
+    test_dict = {"UserDate1": ""}
+    assert transform(test_dict) == {"UserDate1": None}
+
+
+def test_handles_badly_formatted_date():
+    test_dict = {"UserDate1": "a badly formatted date"}
+    assert transform(test_dict) == {"UserDate1": None}
```
alexwlchan: Do you want to be throwing this data away? Some of the Calm dates are super funky – maybe stash the raw string in `UserDate1.raw` or something?

reporting/reporting_calm_transformer/src/transform.py
```diff
@@ -0,0 +1,45 @@
+import math
+from copy import deepcopy
+from dateutil.parser import parse
+
+
+def convert_date_to_iso(date_string):
+    try:
+        return parse(date_string).date().isoformat()
```
alexwlchan: You should check if the Calm date ever uses "day-first" like in the UK in a way that could be ambiguous.

See https://notebook.alexwlchan.net/2018/10/beware-ambiguous-dates-with-dateutil-parse/

reporting/reporting_calm_transformer/src/transform.py
```diff
@@ -0,0 +1,45 @@
...
+                new_value = new_value[:-1]
+
+        if key in keys_to_parse:
+            new_value = convert_date_to_iso(new_value)
+
+        transformed_record[key] = new_value
+    return transformed_record
+
+
+keys_to_parse = {
```
alexwlchan: 👍 for set

reporting/reporting_calm_transformer/src/transform.py
```diff
@@ -0,0 +1,45 @@
...
+                new_value = None
+
+        if isinstance(new_value, list) and len(value) == 1:
+            new_value = record[key][0]
+
+        if isinstance(new_value, str):
+            if new_value.startswith("'"):
+                new_value = new_value[1:]
+            if new_value.endswith("'"):
+                new_value = new_value[:-1]
```
alexwlchan: What if a value is only partially quoted? Should that still be stripped?

reporting/reporting_miro_transformer/src/transform.py
```diff
@@ -40,7 +38,7 @@ def convert_date_to_iso(date_string):
     try:
         return parse(date_string).date().isoformat()
     except (ValueError, TypeError):
-        return date_string
+        return None
```
alexwlchan: minor: Good Python practice is to throw an exception rather than return sentinel values, so you're not relying on the caller to check `if foo is None`

reporting/reporting_sierra_transformer/src/test_transform.py
```diff
@@ -0,0 +1,121 @@
...
+        "multiple_dicts_multiple_subfields": [
+            {"a": "b", "x": "y"},
+            {"a": "c", "x": "z"},
+        ],
+        "orders_date": ["1970-01-01T00:00:00"],
+        "publishYear": 1970,
+        "empty_string": "",
+    }
+
+    if full:
```
alexwlchan: Rather than an ambiguously named `full` parameter, how about splitting this into two methods?

```python
def build_bib_record()

def build_sierra_transformable()
```

reporting/reporting_sierra_transformer/src/test_transform.py
```diff
@@ -0,0 +1,121 @@
...
+        "string_field": "some string",
+        "boolean_field": True,
+        "single_dict_single_subfield": {"a": "b"},
+        "single_dict_multiple_subfields": {"a": "b", "x": "y"},
+        "multiple_dicts_single_subfield": [{"a": "b"}, {"a": "c"}],
+        "multiple_dicts_multiple_subfields": [
+            {"a": "b", "x": "y"},
+            {"a": "c", "x": "z"},
+        ],
+        "orders_date": ["1970-01-01T00:00:00"],
```
alexwlchan: As before, different day/month.

reporting/reporting_sierra_transformer/src/transform.py
```diff
@@ -0,0 +1,165 @@
...
+        should the field be deleted from the transformed record
+
+    Returns
+    -------
+    edit_record : dict
+        the modified record with an unpacked form of the specified field
+    """
+    try:
+        data_to_unpack = deepcopy(view_record[field_name])
+        del edit_record[field_name]
```
alexwlchan: I think these are the only two lines that can throw a `KeyError`? In general it's good practice to keep the body of your `try` small, so you don't catch and lose unexpected errors.

reporting/reporting_miro_transformer/src/test_date_transforms.py
```diff
@@ -0,0 +1,21 @@
+from transform import transform
+
+
+def test_cleans_simple_date():
+    test_dict = {"all_amendment_date": "01/01/1970"}
+    assert transform(test_dict) == {"all_amendment_date": "1970-01-01"}
```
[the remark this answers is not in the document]
harrisonpim: Possibly, although i'm not sure it's worth it at this point. Will keep in mind though

reporting/reporting_calm_transformer/src/transform.py
```diff
@@ -0,0 +1,44 @@
...
+
+        if isinstance(new_value, (int, float, complex)):
+            if math.isnan(value):
+                new_value = None
+
+        if isinstance(new_value, list) and len(value) == 1:
+            new_value = record[key][0]
+
+        if isinstance(new_value, str):
+            if new_value.startswith("'") and new_value.endswith("'"):
```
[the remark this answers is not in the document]
harrisonpim: haven't seen any instances of ugly doubles, so leaving them be for now

PR #3116, opened 2018-12-02
Use the latest version of scala-json

alexwlchan:
The new version of scala-json has built-in support for URI and UUID decoders/encoders (well, URIs – apparently Circe already handles UUIDs).

See https://github.com/wellcometrust/scala-json/pull/5

Plus some better error messages from JsonAssertions when a test fails.

### DIFF
--- a/archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/json/URIConverters.scala
+++ b/archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/json/URIConverters.scala
@@ -1,10 +1,9 @@
 package uk.ac.wellcome.platform.archive.common.json
 
-import java.net.{URI, URISyntaxException}
+import java.net.URI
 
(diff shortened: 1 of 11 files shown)


archive/common/src/main/scala/uk/ac/wellcome/platform/archive/common/json/URIConverters.scala
```diff
@@ -1,10 +1,9 @@
...
-import java.net.{URI, URISyntaxException}
+import java.net.URI
 
 import com.gu.scanamo.DynamoFormat
-import io.circe.{Decoder, DecodingFailure, Encoder}
-import uk.ac.wellcome.json.JsonUtil.{fromJson, toJson}
+import uk.ac.wellcome.json.JsonUtil._
 
 trait URIConverters {
   implicit val fmtUri =
```
mklander: Is this still needed?

PR #3126, opened 2018-12-03
[WIP] Upgrade the pipeline and API to use Elasticsearch 6

alexwlchan:
Which should unblock us using a fixed version of elastic4s.

I'm deploying this into a new catalogue pipeline and indexing into a new cluster.


alexwlchan:
Replaced by #3127.

PR #3127, opened 2018-12-03
Support Elasticsearch 6 in the pipeline

### LINKED ISSUE #3123: Upgrade to Elasticsearch 6

alexwlchan:
Which should let us back out #3125. Resolves #3123.


catalogue_pipeline/ingestor/src/main/scala/uk/ac/wellcome/platform/ingestor/services/WorkIndexer.scala
```diff
@@ -49,19 +50,26 @@ class WorkIndexer(
...
         bulk(inserts)
       }
-      .map { bulkResponse: BulkResponse =>
-        val actualFailures = bulkResponse.failures.filterNot {
-          isVersionConflictException
-        }
+      .map { response: Response[BulkResponse] =>
+        if (response.isError) {
+          warn(s"Error from Elasticsearch: $response")
+          Right(works)
```
alicefuzier: Why right here? doesn't that cause messages to be deleted from the queue even though there was an elasticsearch failure?

alexwlchan: Because I am a goofball and meant `Left`, I think. Oops. Will fix when I get in.

catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchServiceTest.scala
```diff
@@ -191,49 +199,77 @@ class ElasticsearchServiceTest
           .sortBy(_.canonicalId)
 
         insertIntoElasticsearch(indexName, works: _*)
-        withElasticsearchService { searchService =>
+
+        val index = Index(indexName)
+
+        // Get the initial ordering from Elasticsearch
+        val initialResponse = searchService
```
alicefuzier: why this?

catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchServiceTest.scala
```diff
@@ -191,49 +199,77 @@ class ElasticsearchServiceTest
...
           (1 to 10).foreach { _ =>
             val searchResponseFuture = searchService
-              .simpleStringQueryResults("A")(
-                createElasticsearchDocumentOptionsWith(indexName),
-                createElasticsearchQueryOptions)
+              .simpleStringQueryResults("A")(index, defaultQueryOptions)
 
             whenReady(searchResponseFuture) { response =>
-              searchResponseToWorks(response) shouldBe works
+              searchResponseToWorks(response) shouldBe initialWorks
```
alicefuzier: This was asserting on works because all these documents should have the same score against the query, so they should be ordered by canonicalld (which is the secondary ordering when doing a query). So you don't need the initial query to get the desired result order. You already know by the way they are constructed

sbt_common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ElasticsearchIndexCreator.scala
```diff
@@ -0,0 +1,74 @@
...
+      .execute {
+        createIndex(indexName).mappings {
+          mappingDefinition
+        }
+      }
+      .map { response: Response[CreateIndexResponse] =>
+        if (response.isError) {
+          if (response.error.`type` == "resource_already_exists_exception" || response.error.`type` == "index_already_exists_exception") {
+            info(s"Index $indexName already exists")
+            update(indexName, mappingDefinition = mappingDefinition)
```
alicefuzier: Doesn't this return a Future[Future[_]]? are we sure we wait on the result of the uopdate? Is there a test for it?

sbt_common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ElasticsearchIndexCreator.scala
```diff
@@ -0,0 +1,74 @@
...
+          } else {
+            Future.failed(
```
alicefuzier: Same here

alicefuzier:
I think there may be a but in ElastisearchIndexCreator where we don't wait on the result of a nested future. I also don't see any tests for ElasticsearchIndexCVreator, but it seems non trivial so perhaps it should have some?
Also in the test about consistent ordering in ElasticsearchServiceTest, there should be no need to make an initial request: the expected order of the works should be deterministic when they get created. Was that test failing?

sbt_common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ElasticsearchIndexCreator.scala
alexwlchan: I don't think there's a test that we wait for it, and I'm not sure we can write one – the nature of Elasticsearch indexes is that sometimes they take a while even when the API completes. We do have a test that it successfully updates a mapping, but there's a 500ms sleep first.

But it is returning `Future[Future[Unit]]`; good catch. Will fix.

alexwlchan:
@alicefuzier 

> I think there may be a bug in ElastisearchIndexCreator where we don't wait on the result of a nested future.

Yep, pushed a fix as eb86d9b.

> I also don't see any tests for ElasticsearchIndexCreator, but it seems non trivial so perhaps it should have some?

They're in ElasticsearchIndexTest, but the name is a bit iffy – I ditched the old `ElasticsearchIndex` trait because it was fiddly in favour of the class, but didn't rename the test.

> Also in the test about consistent ordering in ElasticsearchServiceTest, there should be no need to make an initial request: the expected order of the works should be deterministic when they get created. Was that test failing?

Yes.

catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchServiceTest.scala
alicefuzier: It's not a default thing in elasticsearch. It depends on the fact that we specify canonicalId as a sort secondary fieldsort in ElsaticsearchService. This test would probably still fail if you remove the secondary ordering because we run the same query 10 times and you can only get the same results if you have secondary ordering, but this shouldn't be needed

alexwlchan: Ah, I hadn't seen that. I'll have another poke around.

alexwlchan: The original test is flaky (possibly on master, possibly something in this patch is exacerbating it).

```
=== PASSED ===
((9ldvnwawsz,A 1VZXkEyOaU), (cd3rffnepd,A QdSIDX1xQC), (oxlnqvmdas,A gjYGndbglv), (y8giokxbla,A PuPQq6z5Iv), (ulg6lvkfcg,A 8k0sYc2Qed))

=== FAILED ===
((d5s8eno21f,A 9alYy75FON), (904oqponnm,A BGRiLhzVSJ), (cso9aepm4o,A fmJhU20iqc), (rhkzqshbo8,A ORHvutgQt3), (wxukcbmkw1,A YWz7jLIQGN))
((904oqponnm,A BGRiLhzVSJ), (cso9aepm4o,A fmJhU20iqc), (d5s8eno21f,A 9alYy75FON), (rhkzqshbo8,A ORHvutgQt3), (wxukcbmkw1,A YWz7jLIQGN))

((5lnl0r01ui,A TPlbdCIsh0), (stbv6g2out,A 1zpzD0kps0), (3uhk8zeujg,A 534eEX3lQg), (csyfs7ngwf,A Rirw3M5QIZ), (pbrr1wravr,A qR6Tc4mUv4))
((3uhk8zeujg,A 534eEX3lQg), (5lnl0r01ui,A TPlbdCIsh0), (csyfs7ngwf,A Rirw3M5QIZ), (pbrr1wravr,A qR6Tc4mUv4), (stbv6g2out,A 1zpzD0kps0))

((pkz9i3gm76,A H1wszbPdWM), (290cqtanty,A q8n84XatBH), (36omhtcale,A gOs8c9zhKn), (dehxtpoviz,A mnEWjJD6ol), (qqygaq4b5l,A i5ssIQIxuV))
((290cqtanty,A q8n84XatBH), (36omhtcale,A gOs8c9zhKn), (dehxtpoviz,A mnEWjJD6ol), (pkz9i3gm76,A H1wszbPdWM), (qqygaq4b5l,A i5ssIQIxuV))

((ui8dvq7rge,A lCQRkfp6YZ), (fypnfmfkc6,A lwdEF4Enr7), (gygvt1gvap,A QW7Fozkwpv), (j512uwjbpr,A dLedGcOVD0), (zcczpz71tc,A nrIY4oIULc))
((fypnfmfkc6,A lwdEF4Enr7), (gygvt1gvap,A QW7Fozkwpv), (j512uwjbpr,A dLedGcOVD0), (ui8dvq7rge,A lCQRkfp6YZ), (zcczpz71tc,A nrIY4oIULc))

((axgyvvdhd1,A oI377vLRRv), (qfslcxwkhi,A 7CPPe63Hwl), (zrnzjqhtpj,A UggopLYwyW), (4jbfjwb8rx,A uqSaTn3KBt), (ynps5j91hw,A sXyrBHb98v))
((4jbfjwb8rx,A uqSaTn3KBt), (axgyvvdhd1,A oI377vLRRv), (qfslcxwkhi,A 7CPPe63Hwl), (ynps5j91hw,A sXyrBHb98v), (zrnzjqhtpj,A UggopLYwyW))

((aira8klpb0,A e730YM1DEe), (abskubgwyf,A e730YM1DEe), (bskovbqzxr,A e730YM1DEe), (kxlficcprx,A e730YM1DEe), (rnwjdyit8r,A e730YM1DEe))
((abskubgwyf,A e730YM1DEe), (aira8klpb0,A e730YM1DEe), (bskovbqzxr,A e730YM1DEe), (kxlficcprx,A e730YM1DEe), (rnwjdyit8r,A e730YM1DEe))

((4bxnw04xlx,A R6KL6JLLFJ), (peghwt1xtu,A R6KL6JLLFJ), (ykcctxp16s,A R6KL6JLLFJ), (e2bdrqhtqa,A R6KL6JLLFJ), (rhruixbphk,A R6KL6JLLFJ))
((4bxnw04xlx,A R6KL6JLLFJ), (e2bdrqhtqa,A R6KL6JLLFJ), (peghwt1xtu,A R6KL6JLLFJ), (rhruixbphk,A R6KL6JLLFJ), (ykcctxp16s,A R6KL6JLLFJ))
```

Examples below, including two cases where I fixed the title so it shouldn't be doing anything funky with the `_score`. Still fails.

alexwlchan: I printed the scores of all the works, and it turns out Elasticsearch 6 is assigning different scores under the hood for works *with the same title*.

```
0.2876821 / 89tjlwrkrg)
0.2876821 / dl3uc8slzc)
0.2876821 / y6lcdpsjfz)
0.18232156 / fw5xgmbnch)
0.18232156 / rfny1igf7x)

0.2876821 / d9dtrli13l)
0.18232156 / lfkuwqwrid)
0.18232156 / m7voqjvytl)
0.18232156 / wcnvrpxdzj)
0.18232156 / xfg4djvypn)

0.2876821 / 0wocur2uii)
0.2876821 / d3xn7ck6l0)
0.2876821 / gmainmep5s)
0.18232156 / 3uctxzpnin)
0.18232156 / odxaiywt5u)
```

It does so consistently, as proven by my amended test. I have no inclination to investigate the nuances of the new Elasticsearch sorting algorithm when presented with gibberish, but I suspect it is the culprit here.

catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/services/WorksService.scala
```diff
@@ -21,32 +24,31 @@ class WorksService @Inject()(searchService: ElasticsearchService)(
...
+  def searchWorks(query: String)(index: Index,
+                                 worksSearchOptions: WorksSearchOptions)
+    : Future[Either[ElasticError, ResultList]] =
     searchService
       .simpleStringQueryResults(query)(
-        documentOptions,
+        index,
         toElasticsearchQueryOptions(worksSearchOptions))
-      .map { createResultList }
+      .map { _.right.map { createResultList } }
```
alicefuzier: What happens if the either is a left? I think in general we should avoid getting the right projection in production code...It's a bit like doing .get on an Option

catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchServiceTest.scala
alicefuzier: uhm let me check the master one. If master it's not flaky we should definitely figure out what's going on here

catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/services/WorksService.scala
alicefuzier: Actually I think this is exactly the same as doing `_.map { createResultList }`, so no need to get the right projection

catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchServiceTest.scala
alicefuzier: The thing is different scores might be something to do with the dfs search. I'm running the test on master 100 times, no failures so far

alicefuzier: Yeah, master is not flaky...

catalogue_api/api/src/main/scala/uk/ac/wellcome/platform/api/services/WorksService.scala
alexwlchan: KK. It’s not a `right.get` because I was trying to map only on the Right value. :-)

alicefuzier: Yeah, I was expecting that to fail with a left but it doesn't. It's still unnecessary in scala 2.12 because Eithers are right biased, so I think you can loose it

catalogue_api/api/src/test/scala/uk/ac/wellcome/platform/api/services/ElasticsearchServiceTest.scala
alexwlchan: master is also not running with an Elasticsearch 6 container.

Can you replicate the weird behaviours you were trying to fix in the staging API?

alicefuzier: This test is supposed to replicate that behaviour. Removing either the dfs query or the secondary sort field makes it fail in master.

alexwlchan: Okay, so either we’re failing to pass that query in this branch or Elasticsearch’s behaviour has changed.

alicefuzier: Uhm, i don't see search_type in elastic4s debug logs. Anyway, I have a jupyter notebook to test this against an elasticsearch index. I'll try it against your es6 index

alicefuzier: Ok, I'm calling it a night. I'm running my script against the es6 cluster and I see inconsistent ordering issues. But I see it also with dfs search_type and secondary ordering. Maybe you're right and we should just fix this using one shard (will still need the secondary ordering). The behaviour of dfs search type seems to have changed

alicefuzier: I think using just one shard will make the original version of this test pass, which is probably what we want. This version is green even though stuff gets inconsistently ordered by es

alicefuzier: Yeah, so with one shard that test passes as it was originally. I think we should do that. This test not passing in its original form is an indication that this branch breaks the fix for the ordering thing (which is true, I checked reproducing the query the api does against the index)

alexwlchan: Blergh, that's annoying.

But one shard sounds good, especially as that's where Elasticsearch is eventually going.

PR #3129, opened 2018-12-03
Use more types from elastic4s 6

alexwlchan:
Elasticsearch 6 is deprecating the idea of separate document types and indexes, and Elasticsearch 7 will remove the separate document type entirely.

To prepare for that migration, elastic4s has:

* Recommended we switch to the index-only APIs
* Introduced some nice new case classes to pass that info around

Let's use those sweet treats, because who doesn't love some extra types?


kenoir:
Waaah! Good stuff. Why so much conflict?

Issue #2969, opened 2018-11-06
Store separate archival and access copies


kenoir:
Access/Archival storage buckets (replication)

**Bucket naming:**
- Access bucket: `wellcomecollection-storage-access`
- Archive bucket: `wellcomecollection-storage-archive`

**Tasks:**
- [x] Create new account for storage buckets
- [x] Create buckets as above
- [x] Add these buckets to dev deny policy
- [x] Ensure services do not namespace storage location (only _space_)
- [x] Test the replicator and make sure it's behaving
- [x] Reproduce the internal links to Goobi
- [x] Ensure Tom switches to using the storage service in the new account

PR #3134, opened 2018-12-04
Add archive_critical stack

mklander:
Add a new stack for infrastructure that is critical to the archive stack.  Initially contains S3 buckets for archive and access storage.

part of #2969


archive/terraform_critical/Makefile
```diff
@@ -0,0 +1,18 @@
...
+STACK_ROOT = archive/terraform_critical
+
+SBT_APPS 	=
+SBT_DOCKER_LIBRARIES =
+SBT_NO_DOCKER_LIBRARIES =
+
+ECS_TASKS 	=
+LAMBDAS 	=
+
+TF_NAME 	= archive_critical
```
alexwlchan: nit: Could we swap the name to be `critical_archive`?

Reasoning: `make archive` could autocomplete to the critical or non-critical stacks; putting `critical` first ensures it can't be accidentally autocompleted.

PR #3137, opened 2018-12-05
Use single shard to solve ordering problem

alicefuzier:


sbt_common/elasticsearch/src/main/scala/uk/ac/wellcome/elasticsearch/ElasticsearchIndex.scala
```diff
@@ -17,9 +17,12 @@ trait ElasticsearchIndex extends Logging {
...
     httpClient
-      .execute(createIndex(indexName).mappings {
-        mappingDefinition
-      })
+      .execute(
+        createIndex(indexName)
+          .mappings {
+            mappingDefinition
+          }
+          .shards(1))
```
alexwlchan: Comment to explain why we use a single shard please. Otherwise we'll forget, and somebody will remove this line in a year's time.

alicefuzier:
Closing as we're going to port this into #3127

PR #3138, opened 2018-12-06
Add an RFC about our AWS account layout

alexwlchan:


kenoir:
My suggestion is that the platform account should handle IAM users (with MFA enforced), all other accounts have roles (in the dev/readonly/budget pattern you describe) to be assumed. 

All other accounts only have a root user for which MFA is turned on and uses one of the yubikeys @wellingk donated (currently held by @jtweed). Passwords should be printed and stored in a safe place.

kenoir:
All accounts will likely have a `projectname-private` repo in which the role templates per account are held along with any other private infra.

kenoir:
Also pinging @jtweed and @wellingk to review.

kenoir:
@wellingk I'm reluctant to have asymmetric permissions between the CLI and the Web console.

Some thought experiments (file these under "based on a true story").

> A developer makes a CLI change to some infrastructure that causes an outage and wants to fix the problem as quickly as possible (and is a bit panicked about it) - the CLI is not intuitive and is harder to use for most folk than the web UI. _It takes longer to resolve the issue, people are sad._

> A senior experienced dev short on time comes up with an idea to test something that might improve the product, the test requires access to production infrastructure. The CLI is awkward to use, the dev gives up as there won't be time to run a test and do everything else. _New idea is not tried, people are sad.

I think it makes sense to have appropriate permissions for developers but to more tightly control _who_ has access to what, and make that information audit-able.

In addition we could probably improve on having multiple identities by pushing permissions into roles and have users assume those roles.

Thoughts?

kenoir:
I've created some modules over here: https://github.com/wellcometrust/terraform-modules/pull/127

PR #3139, opened 2018-12-06
Don't error in the Sierra transformer if we see an unrecognised LocationType

alexwlchan:
All we'll do is edit the spreadsheet with whatever value we see, and re-run it – let's skip the manual step and the alarm spam.


alicefuzier:
Are we doing this because of the ycons locantions? I spoke with Branwen last week and those are test records with fake locations so they shouldn't end up in the catalogue

PR #3144, opened 2018-12-06
Create a topic for edits to the Miro VHS data

alexwlchan:
The editing script then forwards the new item to the topic, so the catalogue transformer gets it immediately.

In theory the reporting pipeline could subscribe to it as well.

It doesn't cover edits of the form “toggle the isClearedForCatalogueAPI” parameter, but we can easily add that later.


alexwlchan:
@kenoir All good now?

PR #3150, opened 2018-12-07
realtime reporting

harrisonpim:
subscribes the miro and sierra reporting lambdas to topics which record updates to the source data in VHS

### Who is this change for?
📄 📊


reporting/terraform/reporting_lambda/main.tf
```diff
@@ -16,9 +16,8 @@ module "reporting_lambda" {
 }
 
 module "reporting_lambda_trigger" {
-  source = "git::https://github.com/wellcometrust/terraform.git//lambda/trigger_sns?ref=v10.2.2"
+  source = "./trigger_sns_subscriptions"
```
alexwlchan: Is there a reason you've vendored this module?

harrisonpim: only temporary - going to get it all working locally first before pushing it over to `terraform-modules`

harrisonpim: looks like, given the more recent changes to the module (particularly the propagation of the `topic_count` variable throughout), it might be worth leaving it here for the foreseeable future after all

Issue #3076, opened 2018-11-26
Improve storage API ingest validation


jtweed:
This breaks down into two things:

- Improved messages
- Additional validation

When a required property is missing, currently the errors are like:

```
The request content was malformed:
Attempt to decode value on failed cursor: DownField(sourceLocation)
```

Instead it should be something simpler, eg:

```
Required property sourceLocation not supplied
```

This should nest, so an error like:

```
The request content was malformed:
Attempt to decode value on failed cursor: DownField(bucket),DownField(sourceLocation)
```

Should become:

```
Required property bucket not supplied on sourceLocation
```

Properties that require specific values should also be validated. Right now that means that the only acceptable values for `ingestType` and `Location.Provider` are:

- `ingestType`: `create`
- `Location.Provider`: `aws-s3-standard`, `aws-s3-ia` 

Any other value should error with a message like:

```
Invalid value supplied for ingestType, valid values are: create`
```

PR #3163, opened 2018-12-11
Better validation errors in ingests endpoint

alicefuzier:
For #3076

### DIFF
--- a/archive/display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayIngestType.scala
+++ b/archive/display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayIngestType.scala
@@ -3,7 +3,7 @@ package uk.ac.wellcome.platform.archive.display
 import io.circe.CursorOp.DownField
 import io.circe.{Decoder, DecodingFailure, Encoder, Json}
 
-sealed trait DisplayIngestType{val id: String}
+sealed trait DisplayIngestType { val id: String }
 
 object CreateDisplayIngestType extends DisplayIngestType {
   override val id: String = "create"
@@ -11,7 +11,8 @@ object CreateDisplayIngestType extends DisplayIngestType {
 
 object DisplayIngestType {
 
-  implicit val decoder: Decoder[DisplayIngestType] = Decoder.instance[DisplayIngestType](cursor =>
+  implicit val decoder
+    : Decoder[DisplayIngestType] = Decoder.instance[DisplayIngestType](cursor =>
     for {
       id <- cursor.downField("id").as[String]
       ingestType <- id match {
@@ -24,9 +25,12 @@ object DisplayIngestType {
       }
     } yield {
       ingestType
-    })
+  })
 
-  implicit val encoder: Encoder[DisplayIngestType] = Encoder.instance[DisplayIngestType] { ingestType =>
-    Json.obj("id" -> Json.fromString(ingestType.id), "type" -> Json.fromString("IngestType"))
-  }
+  implicit val encoder: Encoder[DisplayIngestType] =
+    Encoder.instance[DisplayIngestType] { ingestType =>
+      Json.obj(
+        "id" -> Json.fromString(ingestType.id),
+        "type" -> Json.fromString("IngestType"))
+    }
 }
--- a/archive/display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayIngestType.scala
+++ b/archive/display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayIngestType.scala
@@ -1,0 +1,30 @@
+package uk.ac.wellcome.platform.archive.display
+
+import io.circe.CursorOp.DownField
+import io.circe.{Decoder, DecodingFailure, Encoder, Json}
+
+sealed trait DisplayIngestType{val id: String}
+
+object CreateDisplayIngestType extends DisplayIngestType {
+  override val id: String = "create"
+}
+
+object DisplayIngestType {
+
+  implicit val decoder: Decoder[DisplayIngestType] = Decoder.instance[DisplayIngestType](cursor =>
+    for {
+      id <- cursor.downField("id").as[String]
+      ingestType <- id match {
+        case CreateDisplayIngestType.id => Right(CreateDisplayIngestType)
+        case _ =>
+          val fields = DownField("id") +: cursor.history
+          Left(DecodingFailure(s"invalid value supplied, valid values are: ${CreateDisplayIngestType.id}.", fields))
+      }
+    } yield {
+      ingestType
+    })
+
+  implicit val encoder: Encoder[DisplayIngestType] = Encoder.instance[DisplayIngestType] { ingestType =>
+    Json.obj("id" -> Json.fromString(ingestType.id), "type" -> Json.fromString("IngestType"))
+  }
+}
--- a/archive/display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayIngestType.scala
+++ b/archive/display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayIngestType.scala
@@ -16,9 +16,11 @@ object DisplayIngestType {
       id <- cursor.downField("id").as[String]
       ingestType <- id match {
         case CreateDisplayIngestType.id => Right(CreateDisplayIngestType)
-        case _ =>
+        case invalidId =>
           val fields = DownField("id") +: cursor.history
-          Left(DecodingFailure(s"invalid value supplied, valid values are: ${CreateDisplayIngestType.id}.", fields))
+          Left(DecodingFailure(
+            s"""got "$invalidId", valid values are: ${CreateDisplayIngestType.id}.""",
+            fields))
       }
     } yield {
       ingestType
--- a/archive/display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayProvider.scala
+++ b/archive/display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayProvider.scala
@@ -36,7 +36,9 @@ object DisplayProvider {
       provider <- id match {
         case StandardDisplayProvider.id => Right(StandardDisplayProvider)
         case InfrequentAccessDisplayProvider.id => Right(InfrequentAccessDisplayProvider)
-        case _ => Left(DecodingFailure("", List(DownField("id"))))
+        case _ =>
+          val fields = DownField("id") +: cursor.history
+          Left(DecodingFailure(s"invalid value supplied, valid values are: ${StandardDisplayProvider.id}, ${InfrequentAccessDisplayProvider.id}.", fields))
       }
   } yield {
     provider
--- a/archive/display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayProvider.scala
+++ b/archive/display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayProvider.scala
@@ -1,13 +1,44 @@
 package uk.ac.wellcome.platform.archive.display
 
-import io.circe.generic.extras.JsonKey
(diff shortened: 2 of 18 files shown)


alicefuzier:
So, the error messages here are slightly different from the error messages suggested in #3076 :
- If a required property is missing the error looks like 
```Invalid value at .sourceLocation: required property not supplied.```
if a nested property is missing, the full path to the missing property is provided i.e. 
```Invalid value at .sourceLocation.bucket: required property not supplied.```

- If we only accept a controlled set of values as we do for `ingestType.id` and `provider.id` and the supplied value is not one of the allowed ones, the error message looks like:
```Invalid value at .sourceLocation.provider.id: invalid value supplied, valid values are: aws-s3-standard, aws-s3-ia.```

- If a property is supplied with a different format than the one supported, the error message looks like:
```Invalid value at .sourceLocation.bucket: should be a String.```

Also, if there are multiple errors in the JSON, they are all accumulated in the response message and line separated i.e. 
```
Invalid value at .sourceLocation: required property not supplied.
Invalid value at .ingestType: required property not supplied.
```
The returned error is a JSON repsonse in the format:
```
{
 "httpStatus": 400,
 "description": "...",
 "label": "Bad Request"
 "type": "Error"
}
```
which is consistent with what we do in the API

File: archive/display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayIngestType.scala [hunk 1]
```diff
@@ -0,0 +1,36 @@
...
+  implicit val decoder
+    : Decoder[DisplayIngestType] = Decoder.instance[DisplayIngestType](cursor =>
+    for {
+      id <- cursor.downField("id").as[String]
+      ingestType <- id match {
+        case CreateDisplayIngestType.id => Right(CreateDisplayIngestType)
+        case _ =>
+          val fields = DownField("id") +: cursor.history
+          Left(DecodingFailure(
+            s"invalid value supplied, valid values are: ${CreateDisplayIngestType.id}.",
```
alexwlchan: Rather than `value`, maybe this could be `id`? And include whatever the ID the user passed was. For example:

> Invalid ID 'foobar'; expected one of "create", "destroy", "preserve".

Because this message could get confusing if you pass multiple keys/values, e.g.

```json
{
    "id": "foo",
    "name": "bar"
}
```

Which of these is the invalid value?

archive/display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayProvider.scala
```diff
@@ -1,13 +1,62 @@
...
+    for {
+      id <- cursor.downField("id").as[String]
+      provider <- id match {
+        case StandardDisplayProvider.id => Right(StandardDisplayProvider)
+        case InfrequentAccessDisplayProvider.id =>
+          Right(InfrequentAccessDisplayProvider)
+        case _ =>
+          val fields = DownField("id") +: cursor.history
+          Left(DecodingFailure(
+            s"invalid value supplied, valid values are: ${StandardDisplayProvider.id}, ${InfrequentAccessDisplayProvider.id}.",
```
alexwlchan: Ditto above; be more specific about which part of the input value is incorrect.

archive/display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayProvider.scala
```diff
@@ -1,13 +1,62 @@
...
+      }
+    } yield {
+      provider
+  })
+
+  implicit val encoder: Encoder[DisplayProvider] =
+    Encoder.instance[DisplayProvider] { provider =>
+      Json.obj(
+        "id" -> Json.fromString(provider.id),
+        "type" -> Json.fromString("Provider"))
```
alexwlchan: Could you use the `ontologyType` on `DisplayProvider`? Otherwise there doesn't seem much of a reason to set it.

archive/display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayIngestType.scala
```diff
@@ -0,0 +1,36 @@
+package uk.ac.wellcome.platform.archive.display
+
+import io.circe.CursorOp.DownField
+import io.circe.{Decoder, DecodingFailure, Encoder, Json}
+
+sealed trait DisplayIngestType { val id: String }
+
+object CreateDisplayIngestType extends DisplayIngestType {
```
alexwlchan: We already use `create` as the prefix for a bunch of our generator methods, and the overlap is confusing me as I read this diff. It's a little icky, but would you consider `CreationDisplayIngestType` to disambiguate them?

archive/progress_http/src/main/scala/uk/ac/wellcome/platform/archive/progress_http/Router.scala
```diff
@@ -22,26 +27,61 @@ class Router(
...
+      case MalformedRequestContentRejection(err, causes: DecodingFailures) =>
+        val message = causes.failures.map { cause =>
+          val path = CursorOp.opsToPath(cause.history)
+
+          // Error messages returned by Circe are somewhat inconsistent and we also return our
+          // own error messages when decoding enums (DisplayIngestType and DisplayStorageProvider).
+          val reason = cause.message match {
+            // "Attempt to decode on failed cursor" seems to mean in circeworld
+            // that a required field was not present.
+            case s if s.contains("failed cursor") =>
```
alexwlchan: Could we match on the more specific string, i.e.

```scala
case s if s.contains("Attempt to decode value on failed cursor")
```

archive/progress_http/src/main/scala/uk/ac/wellcome/platform/archive/progress_http/model/ErrorResponse.scala
```diff
@@ -0,0 +1,6 @@
+package uk.ac.wellcome.platform.archive.progress_http.model
+
+case class ErrorResponse(httpStatus: Int,
```
alexwlchan: This doesn't quite match the errors we return from the catalogue API, which have an `errorType` of `http`.

We should probably put those models in a common lib.

archive/progress_http/src/main/scala/uk/ac/wellcome/platform/archive/progress_http/Router.scala
```diff
@@ -22,26 +27,61 @@ class Router(
...
+              "required property not supplied."
+            // These are errors returned by our custom decoders for enum.
+            case s if s.contains("invalid value") => s
+            // If a field exists in the JSON but it's of the wrong format
+            // (for example the schema says it should be a String but an object has
+            // been supplied instead), the error message returned by Circe only
+            // contains the expected type.
+            case s => s"should be a $s."
+          }
```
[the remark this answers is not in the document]
alexwlchan: Not sure I can make it any better, but comments in the code are really helpful. 👍

archive/progress_http/src/test/scala/uk/ac/wellcome/platform/archive/progress_http/ProgressHttpFeatureTest.scala
```diff
@@ -216,5 +285,353 @@ class ProgressHttpFeatureTest
...
+
+            val request = HttpRequest(
+              method = POST,
+              uri = url,
+              headers = Nil,
+              entity = entity
+            )
+
+            whenRequestReady(request) { response: HttpResponse =>
+              println(response)
```
alexwlchan: Should this still be here?

archive/progress_http/src/test/scala/uk/ac/wellcome/platform/archive/progress_http/ProgressHttpFeatureTest.scala
```diff
@@ -216,5 +285,353 @@ class ProgressHttpFeatureTest
...
+                  |  }
+                  |}""".stripMargin
+            )
+
+            val request = HttpRequest(
+              method = POST,
+              uri = url,
+              headers = Nil,
+              entity = entity
+            )
```
alexwlchan: Not for this PR, but I wonder if we should push this repeated POST request code into a fixture?

archive/progress_http/src/test/scala/uk/ac/wellcome/platform/archive/progress_http/ProgressHttpFeatureTest.scala
```diff
@@ -216,5 +285,353 @@ class ProgressHttpFeatureTest
...
+              response.status shouldBe StatusCodes.BadRequest
+              response.entity.contentType shouldBe ContentTypes.`application/json`
+
+              val progressFuture =
+                Unmarshal(response.entity).to[ErrorResponse]
+
+              whenReady(progressFuture) { actualError =>
+                actualError shouldBe ErrorResponse(
+                  400,
+                  "Invalid value at .ingestType.id: invalid value supplied, valid values are: create.",
```
alexwlchan: “Invalid value at, invalid value supplied”? Bit repetitive. How about:

> Invalid value at .ingestType.id: got "baboop", expected "create".

archive/progress_http/src/test/scala/uk/ac/wellcome/platform/archive/progress_http/ProgressHttpFeatureTest.scala
```diff
@@ -216,5 +285,353 @@ class ProgressHttpFeatureTest
...
+                actualError shouldBe ErrorResponse(
+                  400,
+                  "Invalid value at .sourceLocation: required property not supplied.",
+                  "Bad Request",
+                  "Error")
+                val requests =
+                  listMessagesReceivedFromSNS(topic).map(messageInfo =>
+                    fromJson[IngestBagRequest](messageInfo.message).get)
+
+                requests shouldBe empty
```
alexwlchan: Simpler: `assertSnsReceivesNothing(topic)`.

archive/display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayProvider.scala
alicefuzier: Good point, I'll delete `ontologyType` cause I don't like it anyway

archive/display/src/main/scala/uk/ac/wellcome/platform/archive/display/DisplayIngestType.scala
alicefuzier: mm I see what you mean, but the id is "create" so I'd prefer to name it as such..

archive/progress_http/src/main/scala/uk/ac/wellcome/platform/archive/progress_http/model/ErrorResponse.scala
alicefuzier: yeah, I'm not sure what's the point of the errorType field if I'm honest. Rather than putting it in hardcoded I went for leaving it out. We can always add it later. I'm not sure if they should be in the same lib though. I can see those models eventually diverging

archive/progress_http/src/test/scala/uk/ac/wellcome/platform/archive/progress_http/ProgressHttpFeatureTest.scala
alicefuzier: oops

archive/progress_http/src/test/scala/uk/ac/wellcome/platform/archive/progress_http/ProgressHttpFeatureTest.scala
alicefuzier: Yeah, you're right, it does repeat. I'll change it. How about
```Invalid value at .ingestType.id: got "baboop", valid values are: create.```
This way we can change the rejectionhandler in  `Router` to use "valid values" in the match block

PR #3169, opened 2018-12-14
bagging and archive report for test b numbers

tomcrane:
Provides links to information and preview of b numbers used for testing bagging and ingesting to archive storage

### Who is this change for?
Platform team


kenoir:
Thanks! We might actually end up putting this in a shared space in a machine readable format and use these bags in our automated testing. Pinging @mklander

Issue #3170, opened 2018-12-14
Synthetic tests for ingestion and improve archive services logging


mklander:
Ingestion appears to stall or fail without reporting errors

Add a synthetic test for ingestion
Improve logging for archive services to aid debugging (include request context)

PR #3173, opened 2018-12-14
Trigger bag ingest lambda

mklander:
A lambda to trigger bag ingests for continuous testing of storage flows

part of #3170


archive/terraform/kms_key/encryption_kms_key.tf
```diff
@@ -0,0 +1,4 @@
+resource "aws_kms_key" "encryption_key" {
```
kenoir: We generally don't bother with KMS, but obviously this is better than not. Probably should spread the pattern everywhere.

PR #3176, opened 2018-12-17
Trigger regular bag ingest

mklander:
Add trigger every 15 minutes for test bag ingest


mklander:
I don't think that module works (e.g., not all the variables used are defined).  Perhaps I should fix it.

PR #3200, opened 2019-01-07
Use transfer manager to download a file from s3

alicefuzier:


alicefuzier:
Yep, I shamelessly copied the code from the bag replicator branch. I agree that the transfer manager approach is much better than the akka streams approach. I tried that and got entangled really quickly.

PR #3207, opened 2019-01-09
Use separate s3 client for transfer manager

alicefuzier:


kenoir:
But _why_?

alicefuzier:
Mostly it's to prove the theory that these new timeouts are caused by the introduction of transfermanager and were not there before.
If we see them with this change, it will give us an indication of where the problem is because the two s3clients are now isolated

PR #3215, opened 2019-01-10
Remove unused variables from our inline Terraform modules

alexwlchan:
Just a straight clean-up using the script I added in the modules repo recently.


alexwlchan:
I’m not overly fussed about that, as they were derived entirely automatically.

alexwlchan:
Yeah, but no biggie even if we don't.

PR #3220, opened 2019-01-10
Use new versions of terraform-modules that support complete VPC endpoints

alexwlchan:


alexwlchan:
Requires 
https://github.com/wellcometrust/terraform-modules/pull/134

PR #3222, opened 2019-01-11
Try to find assets the DLCS doesn't know about

tomcrane:
The bagger uses the DLCS to determine the best place to collect the asset from. But the DLCS doesn't know about all assets. An asset record might be missing from the DLCS because of a problem with a sync operation, or it's a closed file that never goes to the DLCS for access purposes. They still need to be bagged...

### Who is this change for?
All

### DIFF
--- a/archive/bagger/src/main.py
+++ b/archive/bagger/src/main.py
@@ -41,7 +41,9 @@ def process_message(message):
     print("processing " + identifier)
     result = bagger_processor.process_bagging_message(body)
     error = result.get("error", None)
-    if error is not None:
+    if error is None:
+        aws.remove_error(identifier)
+    else:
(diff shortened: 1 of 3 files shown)


tomcrane:
This will require a redeploy of the bagger

archive/bagger/src/main.py
```diff
@@ -41,7 +41,9 @@ def process_message(message):
     print("processing " + identifier)
     result = bagger_processor.process_bagging_message(body)
     error = result.get("error", None)
-    if error is not None:
+    if error is None:
```
tomcrane: I assume the running bagger still uses this `main` function as it processes the queue?

If so, if no errors during bagging, get rid of any error previously logged to the error bucket.

PR #3326, opened 2019-01-17
Allow platform to publish storage images

kenoir:

We are in the process of moving the storage service to its own repo - for now we need to publish container images from the platform account.

### Who is this change for?

🐶 🍇


kenoir:
Requires: https://github.com/wellcometrust/terraform-modules/pull/141

PR #3340, opened 2019-01-21
Allow the migration tool to read from a fixture list

tomcrane:
A bit more flexibility in load testing - the migration tool accepts a file with a list of arbitrary b numbers (there's one included in this PR for the reference list) as well as the existing filter and single b number forms.

### Who is this change for?
Migrators


kenoir:
Superseded by #3354

Issue #3265, opened 2019-01-15
Move subject IDs to subject


jtweed:
Based on a misunderstanding of LC names and LCSH, we are currently adding the ID on subject fields to the first concept in the concepts list.

This is sometimes the case, but sometimes the ID is of a ComplexSubject in LC that is made of the separate concepts in the list, each of which then have their own IDs.

We can't fully solve this until we bring in LC names and LCSH, so in the meantime, let's move the ID from the first concept to the subject itself.

---

New transformation checklist:

- [x] Make code change (#3352)
- [x] Deploy new transformer
- [x] Start a reindex
- [x] Check the reindex completed
- [x] Update the production API index

alexwlchan:
This is turning into a much more involved patch than I originally realised. 🤯

jamesgorrie:
Just to be clear, we'll be moving from:
```JSON
{
	"subjects": [{
			"label": "Human skeleton.",
			"concepts": [{
				"id": "vegjfgd8",
				"label": "Human skeleton.",
				"type": "Concept"
			}],
			"type": "Subject"
		},
		{
			"label": "Anatomical specimens.",
			"concepts": [{
				"id": "k43gjbuy",
				"label": "Anatomical specimens.",
				"type": "Concept"
			}],
			"type": "Subject"
		},
		{
			"label": "Anatomy, Pathological.",
			"concepts": [{
				"id": "yx7v4jrq",
				"label": "Anatomy, Pathological.",
				"type": "Concept"
			}],
			"type": "Subject"
		}
	]
}
```

to 

```JSON
{
	"subjects": [{
			"label": "Human skeleton.",
		 	"id": "vegjfgd8",
			"concepts": [{
				"id": "vegjfgd8",
				"label": "Human skeleton.",
				"type": "Concept"
			}],
			"type": "Subject"
		},
		{
			"label": "Anatomical specimens.",
                         "id": "k43gjbuy",
			"concepts": [{
				"label": "Anatomical specimens.",
				"type": "Concept"
			}],
			"type": "Subject"
		},
		{
			"label": "Anatomy, Pathological.",
                         "id": "yx7v4jrq",
			"concepts": [{
				"label": "Anatomy, Pathological.",
				"type": "Concept"
			}],
			"type": "Subject"
		}
	]
}
```

Looking at it - I am assuming those are our IDs?

alexwlchan:
@jamesgorrie Yes.

[...]

Issue #3264, opened 2019-01-15
Add label to production events


jtweed:
It turns out that it's not possible to always unambiguously create a label from structured data within a production event, if the source of the data is 260. It is possible when it is 264.

To work around this, we are going to leave what we have in place, but add a `label` property to production events, that can be used for display.

Silver's docs have been updated to reflect this:

https://github.com/wellcometrust/platform/blob/f527915e013e343a97a6ca3ab1f26df6e5cca7b7/ontologies/WIP/sierratransformable#L414

But amount to: take production event label from entire field concatenating all sub-fields.

PR #3350, opened 2019-01-22
Use moar generators for Subject and Production

alexwlchan:
A bit of refactoring in anticipation of #3264 and #3265. Nothing too different, but lays the ground for useful changes later and should help reduce the diffs on those patches.


alexwlchan:
Actually this isn’t what I want.

PR #3352, opened 2019-01-22
Move IDs to subject

alexwlchan:
Code change for #3265. Turns out replumbing the way subjects are identified is quite a disruptive change! :D


alexwlchan:
I was considering removing the Displayable logic from subject concepts, but we'll have to put it back in six months so I don't think I'll bother.

PR #3360, opened 2019-01-24
Force sbt to dump a copy of the dependency tree every time, and massively simplify the build system

alexwlchan:
This PR is the result of a breakfast experiment.

Currently we have a lot of logic in `travistooling` that tries to guess whether a particular changed file affects a particular project. This is complicated and fiddly and hard to write, boo. We're not in the business of writing dependency trees…

…but sbt is. And it knows the exact dependency tree, because it compiles it in `build.sbt`.

This patch modifies `build.sbt` in two ways:

* It adds our own wrapper around the sbt DSL to make it slightly cleaner and easier to read
* Within the wrapper, it saves a small amount of information about each project. These are the JSON files in `builds/sbt_metadata`. In a future patch, we can read this instead of mucking around with Make, and ditch all our “does X depend on Y” guessing.


build.sbt
```diff
@@ -2,198 +2,221 @@ import sbt.Keys._
...
+  // future build system to determine whether to run tests based on the
+  // up-to-date project graph.
+  // See https://www.scala-sbt.org/release/docs/Howto-Generating-Files.html
+  val file = new File(s"builds/sbt_metadata/${project.id}.json")
+  val dependenciesJson = localDependencies
+    .map { p: Project => p.id }
+    .map { id: String => s""""$id""""}
+    .mkString(", ")
+
+  IO.write(file, s"""
```
kenoir: Why are we writing JSON like this? You've got the whole Scala ecosystem to play with.

alexwlchan: Running inside sbt seems to be missing a bunch of bits – in particular, I couldn’t find `FileWriter` or `JSONObject`.

Maybe they need importing if you don't have the standard library by default? It’s only just occurred to me to try – I was trying to get something done quickly.

kenoir: Also I think there are pre-built SBT plugins which do this: https://github.com/sbt/sbt-buildinfo

alexwlchan: Ooh, much nicer

alexwlchan: And now you ask, I’ve put in a proper JSON encoder.

PR #3365, opened 2019-01-24
Add a script for checking the Travis caches

alexwlchan:
One lever for improving Scala compile times is making sure we cache builds correctly.

Now we have handy metadata about all the projects, this script checks for incorrect entires in the cache list. Output as follows:

```console
$ python travistooling/check_travis_cache.py
Don't cache these directories:
 - storage/commmon/target

Do cache these directories:
 - sbt_common/common/target
 - storage/bag_replicator/target
 - storage/common/target
```

I’m keeping this separate from #3360 so I can merge that ASAP.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@ cache:
 
     - catalogue_api/api/target
 
-    - catalogue_pipeline/common/target
     - catalogue_pipeline/id_minter/target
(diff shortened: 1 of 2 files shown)


alexwlchan:
ah but #3360 just failed so I'll have to rebuild anyway

might as well merge then

PR #3367, opened 2019-01-24
A better README for bagger and tools

tomcrane:
A quick reference for the various migration tools

### Who is this change for?
Migrators


tomcrane:
I thought we might want to throttle some operations. So far I haven't used them though.

PR #3381, opened 2019-01-30
Rename the next archivist topic

mklander:
Rename the registrarTopic because it is no longer the registrar


alexwlchan:
Next what?

Issue #3285, opened 2019-01-17
Storage manifests should provide reference to access copy of assets


kenoir:
In order for the storage service to present the expected location for assets it should provide links to the access bucket.

We should:
- [ ] Ensure assets are available when the storage service is available.
- [ ] Check with DDS/DLCS that those assets are accessible.

PR #3410, opened 2019-02-05
Get the destination BagLocation from BagStorage

alexwlchan:
Some more refactoring for #3285. This modifies the BagStorage class to return a list of the BagLocation of the copied bag, not an S3 CopyResult.

I ended up doing quite a bit of refactoring in BagStorage (and adding a couple of extra tests!), so this is a standalone patch that can be merged separately.


storage/bag_replicator/src/test/scala/uk/ac/wellcome/platform/archive/bagreplicator/BagStorageTest.scala
```diff
@@ -26,33 +24,50 @@ class BagStorageTest
     with IntegrationPatience
     with BagReplicatorFixtures {
 
-  it("should duplicate a bag into a given location") {
+  val bagStorage = new BagStorage(s3Client = s3Client)
```
kenoir: `withBagStorage`?

alexwlchan: I considered it, but it doesn’t have any state and keeps the indentation cleaner. So I took it out, but I don't feel strongly about it.

Issue #3406, opened 2019-02-05
Break out the storage code into a proper repo, for realsies


alexwlchan:
- [ ] Pull the Scala library release code into a Docker container.
- [x] Create a "scala-fixtures" library that just contains our fixture code (from the current `common` lib). Update our existing Scala libraries to use this code.
- [x] Create a "scala-typesafe-app" library that contains the rest of the current `common` code and our existing `config_core` library. Update the platform to use this code.
- [x] Find a way to build multiple libraries in a single repo; add it to the release code.
- [x] Shift config_{storage,messaging,monitoring} to be distinct libraries published alongside their main code, release as typesafe_* libraries.

PR #3412, opened 2019-02-06
Use the new scala-fixtures library

alexwlchan:
Part of #3406.


alexwlchan:
This is failing because the API tests keep crashing with out-of-memory errors. 😕 

See #3415.

PR #3413, opened 2019-02-06
Retry with upper case X in bnum checksum if ALTO fetch fails

tomcrane:
Fix a case mismatch between what is in the METS file for a given ALTO path and what's actually on disk

### Who is this change for?
Storage migration


alexwlchan:
Closing in favour of https://github.com/wellcometrust/storage-service/pull/24

PR #3414, opened 2019-02-06
Remove distinction between archive and access copies

jtweed:

Remove the distinction between archive and access copies, to simplify the storage service.

### Who is this change for?

People who want the storage service to be as dumb as possible.


docs/rfcs/002-archival_storage/README.md
```diff
@@ -20,15 +20,15 @@ We will build a storage service based on Amazon S3 and DynamoDB.
 ![archival storage service - page 1](storageservice-20190109.png)
```
jtweed: @kenoir Can you update this as part of this PR?

docs/rfcs/002-archival_storage/README.md
```diff
@@ -530,18 +531,18 @@ Response:
...
+    {
+      "type": "Location",
+      "provider": {
+        "type": "Provider",
+        "id": "aws-s3-ia",
+        "label": "AWS S3 - Infrequent Access"
+      },
+      "bucket": "bucketname",
+      "path": "/digitised/b24923333",
+      "url": "http://bucketname.s3-eu-west-1.amazonaws.com/digitised/b24923333"
```
alexwlchan: If I’m a consumer like DLCS or DDS, how do I know which of these locations I should request? Look at the provider?

tomcrane: This is how the DDS does it atm:
```
...
bucketName = storageManifest["accessLocation"]["bucket"].Value<string>();
path = storageManifest["accessLocation"]["path"].Value<string>();
...
```

Let me know if this needs changing, it's easily done and redeployed.

Issue #3284, opened 2019-01-17
Storage stack application version tracking


kenoir:
In order to properly identify what services are deployed where and keep them current with master, we should publish released versions in a way that makes them accessible and easily visible in the storage service.

We should:

- [x] Update publish_service container to write to SSM
- [x] Update Storage stack to use SSM as data source for app versions

PR #3430, opened 2019-02-07
Better builds for a better life (TM)

kenoir:

Use the new release tooling format path.

### Who is this change for?

👨 + 🐶


kenoir:
For #3284

PR #3459, opened 2019-02-25
add script to populate dummy calm VHS

harrisonpim:
#3248 , following #3456. Added script populates the new Calm VHS table with a static copy of the source data

### DIFF
--- a/monitoring/scripts/populate_dummy_calm_vhs.py
+++ b/monitoring/scripts/populate_dummy_calm_vhs.py
@@ -10,9 +10,10 @@ from download_oai_harvest import fetch_calm_records
 CALM_TABLE = "vhs-calm"
 CALM_BUCKET = "wellcomecollection-vhs-calm"
 
+
 def calm_records():
     """
-    Yields calm records line by line, either from a local file downloaded 
+    Yields calm records line by line, either from a local file downloaded
     previously, or by downloading the records individually from the source.
     """
     dir_path = os.path.dirname(os.path.realpath(__file__))
@@ -24,12 +25,12 @@ def calm_records():
             for record in json.load(f):
                 yield record
     else:
-       print(
-           "Can't find calm_records.json locally. "
-           "Using data downloaded from source instead. "
-       )
-       for record in fetch_calm_records():
-           yield record
+        print(
+            "Can't find calm_records.json locally. "
+            "Using data downloaded from source instead. "
+        )
+        for record in fetch_calm_records():
+            yield record
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


monitoring/scripts/populate_dummy_calm_vhs.py
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env python
+# -*- encoding: utf-8
+
+import os
+import json
+import boto3
+import hashlib
+from download_oai_harvest import fetch_calm_records
```
alexwlchan: nit: PEP 8 import order would split out `boto3` and `download_oai_harvest` as separate import groups, because they're third-party and first-party respectively.

monitoring/scripts/populate_dummy_calm_vhs.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+    file_path = os.path.join(dir_path, "calm_records.json")
+    print(file_path)
+
+    if os.path.exists(file_path):
+        with open(file_path) as f:
+            for record in json.load(f):
+                yield record
+    else:
```
alexwlchan: Ooh, looking for the file first is quite nice!

One suggestion: rather than checking `if exists(file) … else`, it's slightly better to do `try: open(file) … except FileNotFoundError`, which prevents against race conditions if the file gets deleted between your check and open.

This is admittedly unlikely, but it's one of the reasons `try … except` is a more idiomatic pattern. Use that unless you have a good reason not to.

monitoring/scripts/populate_dummy_calm_vhs.py
```diff
@@ -0,0 +1,69 @@
...
+            "location": {
+                "M": {
+                    "key": {"S": s3_key},
+                    "namespace": {"S": CALM_BUCKET}
+                }
+            },
+            "version": {"N": "1"}
+        }
+
+        dynamodb_client.put_item(
```
alexwlchan: This is fine, but potentially slow – you have to make a PutItem call for every record in Calm. It’s probably faster to use a BatchWriter. See https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html?highlight=dynamodb#DynamoDB.Table.batch_writer – you’d write something like:

```python
dynamodb = boto3.resource("dynamodb")
table = dynamodb.Table(CALM_TABLE)

with table.batch_writer() as batch:
     for vhs_record in vhs_records_to_store():
          batch.put_item(Item=vhs_record)
```

This will bunch up the records and reduce the number of API calls you have to make. Faster (and cheaper!).

alexwlchan:
(Why auto-formatting is broken SSH keys got messed up somewhere and haven’t got round to fixing it.)

monitoring/scripts/populate_dummy_calm_vhs.py
harrisonpim: Great suggestion, thank you - going to make this change as it's fairly likely that I'll want to run this again in future.

PR #3464, opened 2019-02-27
create VHS for data inferred from calm source

harrisonpim:
#3250, it's VHSs all the way down


alexwlchan:
The correct plural of “VHS” is “Blockbuster”.

Issue #3503, opened 2019-03-21
Verify how asset origin is determined


tomcrane:
**Given a bag identifier, how does a client of the storage service determine where to obtain the bytes of an asset managed by the storage service?**

_This is not specific to the DDS and DLCS, but is obviously informed by what they currently do._

### General case
_Based only on the storage manifest (the storage service can be used for anything as long as it's bagged)._

As an API consumer, I only know the identifier; I don't have any information about version history for that identifier.

* Fetch the current (i.e., latest) version of the storage manifest: 
`GET /bags/{spaceId}/{externalId}` 
(I'll never request a particular version here).
* In the storage manifest, find the Location object in `locations` where `provider.id` is `aws-s3-ia`. This special knowledge is the client's preference for S3 locations. This is client configuration.
* Store the `url` property of this Location object for use as a prefix in the next step.
* **For a given File object in `manifest.files`, the access location is given by `{url}/{file.bagVersion}/{file.path}` where {url} is the url from the previous step.**

### Specific case (digitised object described by METS)
_...where the a METS file is present in the bag, and the METS file refers to other assets also in the bag, and the client is using the METS file to determine which assets it wants to obtain the location of (e.g., providing asset delivery for digitised content)._

The extra complexities here are:

1. Determining which file in the storage manifest is the METS file.
2. Determining where to obtain the correct version of the METS file.
3. Traversing the METS file, which contains only relative paths, and matching files mentioned in the METS file to files mentioned in the storage manifest. There may be files mentioned in the storage manifest that aren't of interest to the client; the file information in the METS file is driving the client business logic, rather than the simple list of assets in the storage manifest.
4. There are further complexities around anchor files and multiple manifestations but won't go into those here.

#### Finding the METS file, then determining access locations for the files it mentions

* From the list in `manifest.files` in the storage manifest, build a map of:
    `file.path` from which the prefix `"data/"` has been stripped => {file object}
   (we'll use this later)
  for example
  ```
   {
       "objects/b22454408_0001.jp2": {..},
       ...
   }
  ```
* To determine the location of the METS file, find the File object in `manifest.files` with the `path` property `data/{identifier}.xml`. Again, this is special client knowledge - the client understands that the things in the storage service that it's interested in contain METS files at that location.
* Determine the location of this file by using the strategy given in **General Case** above
* Load the METS file 
* As you traverse the METS file, you come across relative file locations like this: 
   `xlink:href="objects/b22454408_0001.jp2"`
[...]

PR #3522, opened 2019-03-26
Changes to storage RFC based on previous comments

jtweed:

### DIFF
--- a/docs/rfcs/002-archival_storage/README.md
+++ b/docs/rfcs/002-archival_storage/README.md
@@ -46,7 +46,7 @@ Two copies of every bag will be stored in S3, one using the Glacier storage clas
 
 Bags will be versioned in storage and all previous versions will be kept indefinitely. We will adopt a forward delta versioning model, where files in more recent versions of bags can refer to files in earlier versions.
 
-In conjunction with worfklow systems that provide only changed files, this model will enable us to reduce our storage costs and the amount of unneccesary reprocessing of unchanged files.
+In conjunction with workflow systems that provide only changed files, this model will enable us to reduce our storage costs and the amount of unneccesary reprocessing of unchanged files.
 
 #### Locations
 
@@ -63,7 +63,7 @@ Within each location, assets will be grouped into related spaces of content and 
 
 #### Assets
 
-Assets will be stored in the above spaces inside the BagIt bags that were transferred for ingest. Unlike during transfer, bags will be stored uncompressed. BagIt is a standard archival file format: https://tools.ietf.org/html/draft-kunze-bagit-08
+Assets will be stored in the above spaces inside the BagIt bags that were transferred for ingest. Unlike during transfer, bags will be stored uncompressed. BagIt is a standard archival file format: https://tools.ietf.org/html/rfc8493
 
 > The BagIt specification is organized around the notion of a “bag”. A bag is a named file system directory that minimally contains:
 >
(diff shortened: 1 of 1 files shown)


docs/rfcs/002-archival_storage/README.md
```diff
@@ -290,15 +290,16 @@ Content-Type: application/json
...
 
 - Check that the supplied version matches the current version
-- Validate and unpack the supplied bag
+- Unpack the supplied bag
 - Store the supplied bag as a new version
-- Update the current version of the bag
+- Register the new version of the bag as the current version
 
-Partial updates, where files that are not changed are not resupplied, are supported through the use of `fetch.txt` in the supplied bag. File references must include the bag version in which the file was previously supplied.
+Partial updates, where files that are not changed are not resupplied, are supported through the use of `fetch.txt` in the supplied bag. File references must specify the full storage location of the previously supplied file, including the version number of the bag in which it was last supplied in the path.
```
mklander: Perhaps an example `fetch.txt`
> https://s3-eu-west-1.amazonaws.com/wellcomecollection-storage-staging-access/alex-testing/b12345/v3/data/b24923333.xml 98765 data/b24923333.xml

docs/rfcs/002-archival_storage/README.md
```diff
@@ -290,15 +290,16 @@ Content-Type: application/json
...
 
 Updates with fetch files should be processed as follows:
 
-- Check that files in `fetch.txt` exist
+- Check that files in `fetch.txt` reference files in the correct bag
+- Check that files in `fetch.txt` exist most recently at the specified version
```
mklander: We should make this explicit too, does the fetch as described here only support references to files already in the storage service (which seems a sensible first step)?  If so how should those entries be recognised, just by matching bucket/space/bag. If entries do not match a previous version or attempt to skip a version by pointing to an older version what should be done?  If we are match only against the previous bag version that could be explicit here too -- either match a fetch entry in the previous version or an existing file.

tomcrane:
I have updated the client view of the API described in https://github.com/wellcometrust/platform/issues/3503. The new version is cleaner, and works from my perspective.

tomcrane:
The DDS, and any DDS successor, are version-unaware (or rather, version-don't-care) consumers of the API. They only want the latest, access, view of the assets for a given identifier. 

Clients that do care can use the `versions` listed in the bags response and then make requests for versioned bags by appending `[?version={version}]` (for example, to generate a report on the asset history for an identifier, showing diffs or whatever). So that's OK.

Although the current usage for testing and development of the storage service, and migration, is not a typical production workflow usage of the API, they are real use cases for, well, testing and development. Should the API have support for pre-production scenarios where you want to delete or overwrite bags without version increments?

tomcrane:
A thought

```

      {
        "type": "File",
        "name": "data/objects/b24923333_001.jp2",
        "path": "v1/data/objects/b24923333_001.jp2",
        "checksum": "e68c93a5170837420f63420bd626650b2e665434e520c4a619bf8f630bf56a7e"
      },
```

While the dropping of `bagVersion`, and the inclusion of both `name` and `path` (with the versioning incorporated into the path) means that the _version-don't-care_ client doesn't have to compute paths, would it still be useful for the `File` to have an explicit version property as well, for clients that do need to be aware of versions? e.g., Goobi?

docs/rfcs/002-archival_storage/README.md
jtweed: There's one in the partial bag example later on in the RFC.

jtweed:
> Should the API have support for pre-production scenarios where you want to delete or overwrite bags without version increments?

I'd say no, unless it proves to be really necessary. We intend to fully clear things down as part of the migration, so post-migration everything should start as v1.

jtweed:
> A thought
> 
> ```
> 
>       {
>         "type": "File",
>         "name": "data/objects/b24923333_001.jp2",
>         "path": "v1/data/objects/b24923333_001.jp2",
>         "checksum": "e68c93a5170837420f63420bd626650b2e665434e520c4a619bf8f630bf56a7e"
>       },
> ```
> would it still be useful for the `File` to have an explicit version property as well, for clients that do need to be aware of versions? e.g., Goobi?

I don't want to make it look like we have file level versioning. I am interested in Intranda's feedback on this before we confirm it though. I'm going to merge now so I can send it to them.

docs/rfcs/002-archival_storage/README.md
jtweed: Is that not covered by "exist most recently at the specified version"? If these checks fail, the ingest should fail.

tomcrane:
The scenario would be a client that sweeps through the file list, picking out all those that match a particular bag version (perhaps the current version) and doing something with them. Would Goobi, or something else, ever need to do that kind of thing?

PR #3621, opened 2019-05-14
add reporting permissions to the reindex SNS topic

jamesgorrie:
As per the [comment here](https://github.com/wellcometrust/catalogue/pull/45#issuecomment-491224116) - this should probably live in an adapter account to avoid direct couplings.

However the reindexer infra (and other catalogue stuff) should probably not live in this account anyway, and it helps me get the [reporting lambdas](https://github.com/wellcometrust/reporting) closer to be usable for working on the date stuff.

This is the last time (rule of three) before abstracting anyway.


jamesgorrie:
Closing in favour of #3639

PR #3677, opened 2019-06-11
Add RFC for new born digital workflow

jtweed:

Add an RFC that describes the initial version of a new born digital workflow that uses Archivematica.

### Who is this change for?

C&R and Digital Production


docs/rfcs/014-born_digital_workflow/README.md
```diff
@@ -0,0 +1,176 @@
...
+|------------------------|-------------|
+| Find/create collection | Find, or create if necessary, the collection in Calm that the items will be part of. |
+| Find/create series     | Find, or create if necessary, the series and sub-series within the collection that the items will be part of. |
+| Create items           | For each set of related files, create an item in Calm that represents and describes them collectively in the catalogue. This should describe the files as a set, not list details of individual files. |
+
+### Transfer
+
+Once each set of related files as been identified, they need to be assembled into a package that can be transferred into Archivematica for processing and storage. This package is a simple zip file, with the addition of a minimal set of descriptive metadata.
+
+This metadata is used for identification only. The canonical version of the metadata, which is also the data that will be displayed to end users, remains the data in Calm. Similarly, whilst Archivematica does have support for rights information, we will not be supplying this through the workflow. Calm will be the sole location for rights information.
```
aray-wellcome: To clarify, does this mean that the metadata in Calm will display in the Universal Viewer rather than the basic metadata packaged with the images in much the same way that books/archival material digitzed through Goobi has some metadata with it but ultimately the UV is actually displaying most of the metadata from Sierra? So, as changes about an item are made in Sierra, the UV will display the most correct metadata? And in the instance of born digital, will the UV just be displaying the information from CALM itself or from a record in Sierra harvested from CALM?

docs/rfcs/014-born_digital_workflow/README.md
```diff
@@ -0,0 +1,176 @@
...
+	|-- folder2
+	    \-- [...] 
+	|-- [...]
+	\-- metadata
+	    \-- metadata.csv
+
+ Note the addition of a `metadata` directory, with a single file named `metadata.csv`, which contains item level metadata from Calm. `metadata.csv` should have the following structure:
+
+	filename,dc.title,dc.creator,dc.date,dc.identifier
+	objects,<Title>,<CreatorName>,<Date>,<AltRefNo>
```
aray-wellcome: I see the dc.identifier will be the <AltRefNo> out of Calm, which I understand is usually an archival number like WT/D/6/3/8 for example. All items that run through Goobi are created through Sierra marc records so their identifier actually becomes the system number, or a bnumber, that is created when an item is added in Sierra. 

How would using the AltRefNo as an identifier work/look like? Most systems don't like having the slash there so would the identifier for the item become WT_D_6_3_8? Also all of our items are ingested by their system number or bnumber but born digital items would need to be found using the AltRefNo instead? How would items from the born digital work stream go into the UV as I believe that it only handles bnumbers from what I can tell?

docs/rfcs/014-born_digital_workflow/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+Once the zip files have been created, they need to be uploaded to AWS S3. As soon as the upload of a zip completes, it should be picked up by Archivematica and a processing workflow started automatically.
+
+Access to S3 requires the installation of a client application. Until this can be made available more widely, zip files should be passed to the Digital Production team for upload to S3.
+
+| Step                 | Description |
+|----------------------|-------------|
+| Create packages      | Move the set of files for each item into its own directory, preserving the original order. |
+| Add package metadata | Add a `metadata.csv` file to each item, inside a directory called `metadata`. This file should include the appropriate item level metadata from Calm. |
+| Upload packages      | Create a separate zip file from the parent directory for each item and upload to AWS S3, via the Digital Production team. |
```
aray-wellcome: Totally understand the need to run these zip files through me for the time being until the appraisal bits are in. Just interested about learning more as to what adding to AWS S3 would look like initially for me and what it might look like when the archivists can ingest into Archivematica themselves.

docs/rfcs/014-born_digital_workflow/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+The final stage of the workflow involves the creation of Archive Information Packages (AIPs) and safely moving them to long term storage. The parts of the process are handled automatically by Archivematica and the new storage service.
+
+Once an AIP has been successfully stored in the storage service,  its unique identifier should be added to the equivalent item record in Calm, so that it can be easily located using the catalogue. For now this addition needs to be a manual process, but we hope to be able to automate it in the future.
+
+| Step                | Description |
+|---------------------|-------------|
+| Create SIP          | Archivematica automatically creates a Submission Information Package (SIP), that describes the files that will be stored as an Archive Information Package (AIP). |
+| Store AIP           | Archivematica automatically creates an AIP from the SIP and ensures it is correctly stored in the storage service. |
+| Link from catalogue | Add the unique identifier of the AIP to the associated item record in Calm, so that the catalogue can be used to find files in the storage service. |
```
aray-wellcome: How will this link work? Click on it and it takes a user into the storage service and depending on their access they can view/download the images as necessary?

aray-wellcome:
Just as a general comment in terms of who would need to watch the items' status in Archivematica to make sure it doesn't get stuck/error. What kind of dashboard overview of items does Archivematica have in order to watch what's being ingested? 

I know Alexandra has expressed interest in giving her team a little more knowledge in ingest systems in order to problem solve basic errors themselves and escalate to me as necessary, which I'm happy to do.

docs/rfcs/014-born_digital_workflow/README.md
amme2: Please for the love of god no b numbers... 

I think it might be advisable to add `<RefNo>` to this list, as this is system allocated (sort of, at least you can't have a Calm record without one) whereas `<AltRefNo>` is too easily left blank in error, duplicated, or can be just designed to be human readable in an illogical sort of way. `<AltRefNo>` is the only identifier we would want to make public, but for robustness the actual link should be `<RefNo>` in my view.

docs/rfcs/014-born_digital_workflow/README.md
amme2: I definitely want to avoid dragging anything through Sierra. The sync to Calm is one way only and is a massive headache for my team. 

This may - almost certainly will - result new workflows / code being required to enable us to render born digital items online. But we need to think that through anyway vis-a-vis sensitivity.

docs/rfcs/014-born_digital_workflow/README.md
amme2: There's a whole missing section on DIPs here really. I would like to get this written up sooner rather than later, even if for practical implementation the priority is getting ingest into storage configured. Not considering access requirements has been a big barrier to progressing born digital in the past and I'd like to avoid that happening again.

amme2:
And a general comment from me to note a lack of general comments from me, as I haven't got time to read this thoroughly and comment properly before I go on leave tomorrow. I've now added a few responses to Ashley's questions above, where she's highlighted some stuff to think about, but we've arranged to meet up on 2nd July to go through together with a fine toothed comb, followed by a workshop with the whole team later in July. Will also need to involve Jenny S. as soon as she's back from sick leave.

docs/rfcs/014-born_digital_workflow/README.md
jtweed: Indeed, this workflow won't result in anything being made available online at this point in time.

To do that will require us to have the Calm data in the Catalogue API, support for born-digital on the new site and the ability to securely authenticate and delivery assets that we can make available.

So this won't result in things going into the UV or require any metadata in Sierra.

docs/rfcs/014-born_digital_workflow/README.md
jtweed: Indeed, only a Calm ID should be used here.

Happy to change to RefNo, will do that before I merge the PR.

docs/rfcs/014-born_digital_workflow/README.md
jtweed: Sure, we can pick this up separately. We need to get you set up for this.

There's a placeholder ticket #3553

docs/rfcs/014-born_digital_workflow/README.md
jtweed: This is one area where we may diverge from Archivematica slightly. It can create a separate DIP, but it's designed for uploading to Atom. I think we can do everything we need to do better with a single package, the AIP, which contains the originals and any copies created for preservation or access.

There is a real limit to the degree that we can consider access up front. Making this content available is not currently a priority for the next year at least.

We should definitely cover off what gets put in to Calm and where during the more detailed conversations around cataloguing implications.

jtweed:
> Just as a general comment in terms of who would need to watch the items' status in Archivematica to make sure it doesn't get stuck/error. What kind of dashboard overview of items does Archivematica have in order to watch what's being ingested?

It does provide this overview, but it is one of the weaker areas of the UI currently. All the information is there, but it doesn't scale well to large numbers of processes and expects routine cleanup to keep the numbers of process logs small.

We intend to improve this, but in the first instance it is what it is. I will talk you through it when we cover off uploading to S3.

jtweed:
And thanks @amme2 that's much appreciated. Merging this absolutely does not mean it is done, it's a living document and remains so. Just that we're happy with it as a stating point for the next round of work. Happy to come along in July and listen in to the whole conversation.

docs/rfcs/014-born_digital_workflow/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+We will be actively contributing to the development of Archivematica, including ensuring that it better addresses our needs around appraisal and cataloguing. This will mean we are able to streamline the workflow over time, as the amount of born-digital material increases.
+
+## Proposed solution
+
+### Principles
+
+There are three key principles underlying this initial workflow:
+
+- Appraisal and selection will remain manual until the functionality in Archivematica improves
```
Jenny-Shaw: This is OK for the initial workflow but we would like to be able to experiment with appraisal tools for de-duping etc. We'd like to do this sooner rather than later for Archive Accreditation.

docs/rfcs/014-born_digital_workflow/README.md
```diff
@@ -0,0 +1,176 @@
...
+- Cataloguing will be done in Calm and file-level description will not be part of the catalogue record
+- All processing steps within Archivematica will be fully automated and require no intervention
```
Jenny-Shaw: This concerns me slightly if it means that we won't have the skills or understanding to know what has happened/gone wrong.

docs/rfcs/014-born_digital_workflow/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Overview
+
+![Workflow diagram](workflow.png)
+
+The workflow consists of the following high-level stages:
+
+1. Appraisal
+2. Cataloguing
```
Jenny-Shaw: I think that it would be more accurate to describe this as inventory rather than cataloguing (in line with the Collections Information Policy and Plan). For all acquisitions, in whatever format, we create an accession record and a minimal record in the catalogue. If you change the name of this stage then this workflow would be suitable for acquisitions. There would then need to be further stages of appraisal and cataloguing which will happen later, sometimes much later.

docs/rfcs/014-born_digital_workflow/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Appraisal
+
+During the appraisal stage, an archivist will determine which files are in scope for the collections and perform a policy-led sensitivity check. We intend to move towards system-assisted appraisal, using future version of Archivematica, but for now this stage will build on existing manual processes.
+
+| Step                 | Description |
+|----------------------|-------------|
+| Disk level appraisal | High level appraisal to determine what storage media is in scope. |
+| File level appraisal | Selection of in scope files from selected storage media. |
+| Sensitivity check    | Policy-led review of selected files to determine whether they contain sensitive material. |
```
Jenny-Shaw: We don't really do a proper sensitivity check at this stage. We will flag some likely issues such as 'contains patient records' or 'some of the correspondence is highly personal'. Sensitivity checking would happen as part of the cataloguing process.

docs/rfcs/014-born_digital_workflow/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Cataloguing
+
+Once an initial appraisal has been completed, the files in scope for archiving need to be arranged into archive items and catalogued in Calm. Items should be created in Calm to represent each related set of files, which will be submitted and stored as a single package in the storage service.
```
Jenny-Shaw: I don't think that I fully understand what you mean by this. This is partly because of terminology - item and file mean specific things to archivists. It would be useful to know what level of ingest will be happening. How big is a bag? One hard-drive? One section of a hard-drive? One floppy disk? 

It might be more appropriate to use the accession record as part of the ingest. If necessary we could create part numbers to an accession if more than one bag is needed per accession. We have tried to do this in the past for ingesting into Preservica - it is a little repetitive for the accessioning archivist but definitely possible.

docs/rfcs/014-born_digital_workflow/README.md
```diff
@@ -0,0 +1,176 @@
...
+	    \-- metadata.csv
+
+ Note the addition of a `metadata` directory, with a single file named `metadata.csv`, which contains item level metadata from Calm. `metadata.csv` should have the following structure:
+
+	filename,dc.title,dc.creator,dc.date,dc.identifier
+	objects,<Title>,<CreatorName>,<Date>,<RefNo>
+
+Once the zip files have been created, they need to be uploaded to AWS S3. As soon as the upload of a zip completes, it should be picked up by Archivematica and a processing workflow started automatically.
+
+Access to S3 requires the installation of a client application. Until this can be made available more widely, zip files should be passed to the Digital Production team for upload to S3.
```
Jenny-Shaw: What would this involve? The movement/copying of files concerns me. Through doing the digital audit it has become clear that this is how we get ourselves into a pickle. Staff changes and inconsistent practise certainly haven't helped, but there are files all over the place and it's difficult to know to which stage in the workflow they have reached.

docs/rfcs/014-born_digital_workflow/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+Archivematica will be configured to perform all processing steps automatically. No manual intervention should be required once zip files created with the current manual appraisal process have been uploaded to S3.
+
+We will use the default format policies and tools that come pre-configured in Archivematica, ensuring ongoing best practice for format identification, metadata extraction and file normalisation. All files are virus checked by Archivematica before processing.
+
+| Step                  | Description |
+|-----------------------|-------------|
+| Format identification | Determine the most reliable way to identify the format of each file, as described in PRONOM. |
+| Metadata extraction   | Run format-specific tools to extract as much metadata as possible from within each file. |
+| File normalisation    | Convert files into the formats best suited for preservation and access, in addition to keeping the original file. |
```
Jenny-Shaw: What sort of reporting is possible on this? As part of the digital audit we know that we need to keep better records about our digital material. It would be very useful to be able to add a record of this normalisation process to the file for audit purposes.

docs/rfcs/014-born_digital_workflow/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+To achieve a fully automated workflow, we will use a pipeline with the following processing configuration:
+
+| Setting                                           | Value    |
+|---------------------------------------------------|----------|
+| Assign UUIDs to directories                       | No       |
+| Send transfer to quarantine                       | No       |
```
Jenny-Shaw: I have some practical questions about virus checking/use of the V drive etc. We currently have to virus check everything before it is transferred onto the V drive ready for ingest. Due to the recent data loss we are also under pressure to encrypt all digital acquisitions - this fills me with fear based on the IT competence of most of our donors! What will the mechanical steps of acquisition look like (this workflow only starts once material is physically in the building)? What scope is there for cutting out some of the copying/transfer media? How long will we have to wait for this to be possible?

docs/rfcs/014-born_digital_workflow/README.md
```diff
@@ -0,0 +1,176 @@
...
+- Cataloguing will be done in Calm and file-level description will not be part of the catalogue record
+- All processing steps within Archivematica will be fully automated and require no intervention
+
+### Overview
+
+![Workflow diagram](workflow.png)
+
+The workflow consists of the following high-level stages:
+
+1. Appraisal
```
Jenny-Shaw: Appraisal is likely to happen twice (this is referred to in our Archive Accreditation plans and application). We will do an initial appraisal on acquisition to remove duplicates (ideally automated, but we don't currently have anything) and things like system files or printer drivers. A more detailed appraisal will then happen as part of the cataloguing process.

docs/rfcs/014-born_digital_workflow/README.md
```diff
@@ -0,0 +1,176 @@
...
+
+- Appraisal and selection will remain manual until the functionality in Archivematica improves
+- Cataloguing will be done in Calm and file-level description will not be part of the catalogue record
+- All processing steps within Archivematica will be fully automated and require no intervention
+
+### Overview
+
+![Workflow diagram](workflow.png)
+
+The workflow consists of the following high-level stages:
```
Jenny-Shaw: Our workflow is not as linear as this makes out. It tends to consist of loops often with significant time lags.

docs/rfcs/014-born_digital_workflow/README.md
amme2: Agree. Automating steps is good for efficiency but only if the whole team have a full understanding of exactly what is happening at each step, so that they can troubleshoot when things go wrong and have confidence in the integrity and authenticity of the results when things work well. If this means more manual interventions in these early stages of implementation, then so be it.

docs/rfcs/014-born_digital_workflow/README.md
amme2: I think this stage should be renamed 'Acquisition Appraisal' or something similar, to distinguish it from the more detailed appraisal and sensitivity checks which happen during arrangement and description (cataloguing).

docs/rfcs/014-born_digital_workflow/README.md
amme2: Like Jenny, I think this would be better described as 'inventory' and linked to the accessioning step in the archive processing workflow. That will entail some changes to the metadata fields required (see below).

docs/rfcs/014-born_digital_workflow/README.md
```diff
@@ -0,0 +1,176 @@
...
+	|-- folder2
+	    \-- [...] 
+	|-- [...]
+	\-- metadata
+	    \-- metadata.csv
+
+ Note the addition of a `metadata` directory, with a single file named `metadata.csv`, which contains item level metadata from Calm. `metadata.csv` should have the following structure:
+
+	filename,dc.title,dc.creator,dc.date,dc.identifier
+	objects,<Title>,<CreatorName>,<Date>,<RefNo>
```
amme2: If this is viewed as inventory rather than cataloguing, it makes more sense to align the metadata to the Calm accession record and the SPECTRUM inventory standard (the minimal viable record). This would require:

- Object (item number) = <RefNo> plus <AccNo> and possibly <AccNo> sub-number for digital components
- Name = <Title>
- Number of objects = <Extent> Do we currently record this in accession records, Jenny or does this require some amendments to the Calm accessions template?
- Brief description = <Title>
- Current location - can presumably be added after processing through in S3?
- Current owner = <Repository>
- Recorder and recording date for this information - again, added via Archivematica workflow?

amme2: This of course assumes that 'metadata.csv' can be customised? 

We can't supply <CreatorName> from either Accession or Catalogue item records.
Nor is <RefNo> on its own sufficient to identify a single accession.

docs/rfcs/014-born_digital_workflow/README.md
amme2: I think there's a missing transfer to Wellcome step anyway. With the corporate move to Sharepoint, I am not sure how much longer we will have access to a network share as a staging area, plus there are security and tracking issues associated with multiple movements of files, as Jenny points out. 

My understanding was that Archivematica enables iterative processing, ie that we could submit accession bundles directly to S3 and then recall them later for detailed appraisal and cataloguing? There could well be a time lag of several years between the acquisition of digital material and the point of cataloguing, and I'd prefer to have bags stored in S3 during this time than on an internal network drive.

docs/rfcs/014-born_digital_workflow/README.md
amme2: Agreed. And what is described as 'cataloguing' here is just basic inventory. Cataloguing would be another step after 'storage'. 

Access / dissemination is missing too. I realise this is an 'initial' workflow but since the decisions we make about access will have an impact on options available earlier in the processing lifecycle, it is important we don't lose sight of the fact that the whole purpose of preservation is to facilitate access.

docs/rfcs/014-born_digital_workflow/README.md
aray-wellcome: There's a backlog feature on Archivematica that I think we could use for this. If we transfer the items into Archivematica, it could sit in the backlog (I believe we can point the backlog storage to be anywhere we want so it could be on S3) anyone can pick them up out of the backlog when they're ready and take them into appraisal/create a SIP for it.

PR #3839, opened 2019-08-16
Split locations into primary and replica

jtweed:

### DIFF
--- a/docs/rfcs/002-archival_storage/README.md
+++ b/docs/rfcs/002-archival_storage/README.md
@@ -555,7 +555,7 @@ Response:
       }
     ]
   },
-  "primaryLocation": {
+  "location": {
     "type": "Location",
     "provider": {
       "type": "Provider",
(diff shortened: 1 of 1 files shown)


docs/rfcs/002-archival_storage/README.md
```diff
@@ -555,28 +555,28 @@ Response:
...
-      "type": "Location",
-      "provider": {
-        "type": "Provider",
-        "id": "aws-s3-ia",
-        "label": "AWS S3 - Infrequent Access"
-      },
-      "bucket": "bucketname-access",
-      "path": "digitised/b24923333",
-      "url": "http://bucketname-access.s3-eu-west-1.amazonaws.com/digitised/b24923333"
+  "primaryLocation": {
```
jtweed: Any opinions on whether this should be `location` or `primaryLocation`?

alexwlchan: I would say `location`. The name of the bucket is `wc-storage`, not `wc-storage-primary`.

kenoir: I think I also prefer `location` for the same reason.

jtweed: Ok will change 👍

PR #3874, opened 2019-09-04
Add RFC for 'how we work'

kenoir:

Update docs to remove old or confusing information and add some documentation about shared libraries and code reviews.

### Who is this change for?

Developers in Digital Engagement


docs/rfcs/015-how_we_work/README.md
```diff
@@ -0,0 +1,21 @@
...
+## Background
+
+In order to work together effectively it is useful to have some guiding principals that underpin work.
+
+This is intended to be a living document and will not represent _everything all the time_.
+
+## Proposed solution
+
+### Areas
+
```
jamesgorrie: Principles?
Things like transparency, pragmatism, collaboration etc?

docs/rfcs/015-how_we_work/code_reviews.md
```diff
@@ -0,0 +1,23 @@
...
+
+Some guidelines:
+
+- Nit picking is okay but should be labelled as such.
+- Developers from any team can comment on any PR that they feel qualified to (it is expected that this happen).
+- The developer who opened a PR should be the one to merge it.
+- Examples of how to structure code or links to resources that clarify are helpful.
+
+
+
```
jamesgorrie: Is there something here around how to take a PR - the thing that comes to mind is "assume good intent". `Polite` is quite subjective.

Maybe something about adding a linting rule is easier than nitpicking code style?

docs/rfcs/015-how_we_work/code_reviews.md
```diff
@@ -0,0 +1,23 @@
+# Code Reviews
+
+Code reviews are intended to ensure code changes meet a high standard and to share best practise between team members.
```
jamesgorrie: what about sharing with the rest of the team (non-devs) what is going out by being transparent. Commenting shouldn't only be on code in the case of function / business logic, and thus PRs in this field should be clear in their description as to what they are doing if someone who doesn't read code will be looking at it.

docs/rfcs/015-how_we_work/README.md
kenoir: Yes. 

This isn't a complete list! Please PR things you want to see.

docs/rfcs/015-how_we_work/code_reviews.md
kenoir: Can you either suggest an inline change or clarify in a future PR?

docs/rfcs/015-how_we_work/README.md
jamesgorrie: Isn't this a PR?

docs/hosted_zones/README.md
```diff
@@ -1,49 +0,0 @@
-# Modifying Route 53 hosted zones
```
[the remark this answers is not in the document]
kenoir: Yeah, good point. It's the same but not exactly - this guide is really nice and clear but IMHO too detailed (especially as we have aspirations to change the mechanism).

docs/rfcs/015-how_we_work/README.md
kenoir: Either works - I'd prefer to merge this and then add things.

docs/rfcs/015-how_we_work/code_reviews.md
jamesgorrie: Code reviews are intended to ensure code changes meet a high standard and to share best practise between team members.

```suggestion
Code reviews are intended to keep the quality of code high and understandable, while ensuring we are working in an iterative, pragmatic way.

They also help share best practise between team members.
```

PR #3883, opened 2019-09-05
remove datascience

jamesgorrie:
👋 

Moved to it's new home


kenoir:
`make: *** No rule to make target `data_science/Makefile'.  Stop.`

Makefile needs updating to not reference missing file.

PR #3900, opened 2019-09-18
Experimental: return the Content-Disposition header in Loris

alexwlchan:
This patch deploys https://github.com/loris-imageserver/loris/pull/476, which I wrote in the PyCon UK sprints yesterday.

It means that images downloaded from Loris have marginally more useful filenames which actually contain an ID, rather than just `default.jpg`. An example ID is `s3_L0026000_L0026168.jpg`. They're not perfect, but they're a step in the right direction.

For https://github.com/wellcometrust/wellcomecollection.org/issues/4364

Actually doing what’s described in the ticket – customising them more fully for our use – would need a way to configure Loris to define an (ident) => (filename) mapping, and that was more than I had time for yesterday.

### DIFF
--- a/assets/Makefile
+++ b/assets/Makefile
@@ -3,12 +3,6 @@ include $(ROOT)/makefiles/functions.Makefile
 
 STACK_ROOT 	=
 
-SBT_APPS 	 =
-SBT_SSM_APPS =
-
-SBT_DOCKER_LIBRARIES    =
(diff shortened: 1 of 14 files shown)


jamesgorrie:
Is this getting very stale?

kenoir:
Done gone 💀

PR #3956, opened 2019-10-23
Remove the ECS/EC2 instance tagger

alexwlchan:
We don't use EC2 instances for ECS any more (most of our ECS tasks run in Fargate), and besides, they're running in entirely separate AWS accounts, so invisible to this process.

This change is already deployed.

Closes #3950.


kenoir:
💥

PR #3962, opened 2019-10-24
Create 017-URL Design

derivadow:
Moving the first draft of the URL design principles to GitHub to allow for open conversation and improvements.

### Who is this change for?
People who care about URLs, people minting new URLs/ schemes

### DIFF
--- a/docs/rfcs/017-URL Design
+++ b/docs/rfcs/017-URL Design
@@ -1,0 +1,36 @@
+URLs are part of the user's online experience--they affect a user's ability to reliably share and reference online resources and discover and interact with the content and services we produce. 
+
+They can also have an impact on the amount of long-term technical debt we carry since getting the URL scheme wrong means we need to maintain redirects for the duration of the site.
+
+We therefore follow the following principles:
+
+1. URLs are designed to be persistent—this is the most important thing to consider when minting new URLs or devising a URL scheme. In practice, this means that URLs MUST NOT include:
+a. reference to specific technology;
+b. dates unless the URL is about a date;
+c. status (old/new/draft etc.)
+d. subject unless the URL is about that subject
+
+2. If a URL changes we SHOULD redirect to a new URL and maintain that redirect forever where a semantically equivalent resource exists;
+
+3. When content is removed and there is no equivalent resource we SHOULD return 410 (http gone) with a link to an archived copy of the resource.
+
+4. URLs SHOULD be as short as possible but no shorter. Deeply nested URLs have an impact on SEO and long URLs can cause problems when used in emails etc. short URLs are therefore preferable while maintaining sufficient entropy to support future growth.
+
+5. There MUST be one URL per thing and all things have a URL. URLs are there to identify things on the Web and people use URLs to point to those things, this means:
+a. all resources MUST have a unique URL;
+b. URLs can't be used to identify two or more resources;
+c. all fragments SHOULD dereference i.e. .../foo#bar should be addressable at .../foo/bar
+d. fragments (anything after a #) don't count as unique URLs;
+e. hash-bang URLs (#!) and other techniques that rely on client side JS MUST NOT be used.
+
+6. URLs are globally unique. A user MUST be able to share a URL and anyone, anywhere in the world MUST be able to de reference the same resource.
+
+7. URLs can identify: things, lists of things and forms. Query parameters SHOULD be avoided for anything that’s not a list. 
+
+8. URLs SHOULD be hackable. A user should be able to hack back a URL and get a broader set of resources e.g. it should be possible to hack back the URL for a story: .../stories/$story to .../stories/ and be returned a list of all stories.
+
+9. URLs MUST NOT include any personally identifiable information, tracking parameters nor state.
+
+10. All content MUST be served over https
+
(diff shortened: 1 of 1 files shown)


docs/rfcs/017-URL Design
```diff
@@ -0,0 +1,36 @@
...
+
+1. URLs are designed to be persistent—this is the most important thing to consider when minting new URLs or devising a URL scheme. In practice, this means that URLs MUST NOT include:
+a. reference to specific technology;
+b. dates unless the URL is about a date;
+c. status (old/new/draft etc.)
+d. subject unless the URL is about that subject
+
+2. If a URL changes we SHOULD redirect to a new URL and maintain that redirect forever where a semantically equivalent resource exists;
+
+3. When content is removed and there is no equivalent resource we SHOULD return 410 (http gone) with a link to an archived copy of the resource.
```
alexwlchan: nit: capitalise HTTP, title case for HTTP statuses

```suggestion
3. When content is removed and there is no equivalent resource we SHOULD return 410 (HTTP Gone) with a link to an archived copy of the resource.
```

docs/rfcs/017-URL Design
```diff
@@ -0,0 +1,36 @@
...
+
+4. URLs SHOULD be as short as possible but no shorter. Deeply nested URLs have an impact on SEO and long URLs can cause problems when used in emails etc. short URLs are therefore preferable while maintaining sufficient entropy to support future growth.
```
alexwlchan:
```suggestion
4. URLs SHOULD be as short as possible but no shorter. Deeply nested URLs have an impact on SEO and long URLs can cause problems when used in emails etc. Short URLs are therefore preferable while maintaining sufficient entropy to support future growth.
```

docs/rfcs/017-URL Design
```diff
@@ -0,0 +1,36 @@
...
+5. There MUST be one URL per thing and all things have a URL. URLs are there to identify things on the Web and people use URLs to point to those things, this means:
+a. all resources MUST have a unique URL;
+b. URLs can't be used to identify two or more resources;
+c. all fragments SHOULD dereference i.e. .../foo#bar should be addressable at .../foo/bar
+d. fragments (anything after a #) don't count as unique URLs;
+e. hash-bang URLs (#!) and other techniques that rely on client side JS MUST NOT be used.
+
+6. URLs are globally unique. A user MUST be able to share a URL and anyone, anywhere in the world MUST be able to de reference the same resource.
+
+7. URLs can identify: things, lists of things and forms. Query parameters SHOULD be avoided for anything that’s not a list.
```
alexwlchan: I’m not sure I understand this one. What’s a good/bad query parameter?

e.g. we use query parameters in the catalogue API, `/works?q=pictures+of+fish`. Is that okay? Or do those URLs not count?

docs/rfcs/017-URL Design
```diff
@@ -0,0 +1,36 @@
...
+
+6. URLs are globally unique. A user MUST be able to share a URL and anyone, anywhere in the world MUST be able to de reference the same resource.
+
+7. URLs can identify: things, lists of things and forms. Query parameters SHOULD be avoided for anything that’s not a list. 
+
+8. URLs SHOULD be hackable. A user should be able to hack back a URL and get a broader set of resources e.g. it should be possible to hack back the URL for a story: .../stories/$story to .../stories/ and be returned a list of all stories.
+
+9. URLs MUST NOT include any personally identifiable information, tracking parameters nor state.
+
+10. All content MUST be served over https
```
alexwlchan: And HTTP URLs redirected to HTTPS?

docs/rfcs/017-URL Design
```diff
@@ -0,0 +1,36 @@
...
+
+11. URLs MUST be designed alongside the user interface and given the same level of care as any other UI component (possibly more because they are harder to change). We SHOULD try to have beautiful URLs
```
alexwlchan:
```suggestion
11. URLs MUST be designed alongside the user interface and given the same level of care as any other UI component (possibly more because they are harder to change). We SHOULD try to have beautiful URLs.
```

docs/rfcs/017-URL Design
```diff
@@ -0,0 +1,36 @@
...
+3. When content is removed and there is no equivalent resource we SHOULD return 410 (http gone) with a link to an archived copy of the resource.
+
+4. URLs SHOULD be as short as possible but no shorter. Deeply nested URLs have an impact on SEO and long URLs can cause problems when used in emails etc. short URLs are therefore preferable while maintaining sufficient entropy to support future growth.
+
+5. There MUST be one URL per thing and all things have a URL. URLs are there to identify things on the Web and people use URLs to point to those things, this means:
+a. all resources MUST have a unique URL;
+b. URLs can't be used to identify two or more resources;
+c. all fragments SHOULD dereference i.e. .../foo#bar should be addressable at .../foo/bar
+d. fragments (anything after a #) don't count as unique URLs;
+e. hash-bang URLs (#!) and other techniques that rely on client side JS MUST NOT be used.
```
kenoir: 🎆

docs/rfcs/017-URL Design
derivadow: Less to do with the API, more to do with the public website. The main issue is that Google won't reindex anything it 'thinks' is a search page nor will it crawl those pages. Which is why our catalogue mostly isn't indexed.

kenoir:
Merging as a starting point.

Issue #3989, opened 2019-10-28
Move Wellcome Collection projects to wellcomecollection Github org


kenoir:
Wellcome Collection has a Github Org: https://github.com/wellcomecollection

## Moved
- [x] https://github.com/wellcometrust/wellcomeplayer-pdf-extension
- [x] https://github.com/wellcometrust/wellcomeplayer-dialogues-module
- [x] https://github.com/wellcometrust/wellcomeplayer-seadragon-extension
- [x] https://github.com/wellcometrust/wellcomeplayer-seadragoncenterpanel-module
- [x] https://github.com/wellcometrust/wellcomeplayer-treeviewleftpanel-module
- [x] https://github.com/wellcometrust/wellcomeplayer-searchfooterpanel-module
- [x] https://github.com/wellcometrust/wellcomeplayer-extendedfooterpanel-module
- [x] https://github.com/wellcometrust/wellcomeplayer-moreinforightpanel-module
- [x] https://github.com/wellcometrust/wellcomeplayer-shared-module
- [x] https://github.com/wellcometrust/wellcomeplayer-mediaelement-extension
- [x] https://github.com/wellcometrust/alpha (should be public wellcometrust/platform#3989)
- [x] https://github.com/wellcometrust/catalogue
- [x] https://github.com/wellcometrust/platform
- [x] https://github.com/wellcometrust/wellcomecollection.org
- [x] https://github.com/wellcometrust/dockerfiles
- [x] https://github.com/wellcometrust/docker_run
- [x] https://github.com/wellcometrust/prismic-remapper
- [x] https://github.com/wellcometrust/storage-service
- [x] https://github.com/wellcometrust/tzitzit
- [x] https://github.com/wellcometrust/archivematica
- [x] https://github.com/wellcometrust/archivematica-docs
- [x] https://github.com/wellcometrust/archivematica-storage-service
- [x] https://github.com/wellcometrust/loris
- [x] https://github.com/wellcometrust/python-gearman
- [x] https://github.com/wellcometrust/scala-fixtures
- [x] https://github.com/wellcometrust/scala-json
- [x] https://github.com/wellcometrust/scala-messaging
- [x] https://github.com/wellcometrust/scala-monitoring
- [x] https://github.com/wellcometrust/scala-storage
- [x] https://github.com/wellcometrust/wellcome-collection-tei
- [x] https://github.com/wellcometrust/wellcome-typesafe-app
- [x] https://github.com/wellcometrust/workflow
[...]

PR #3990, opened 2019-10-28
Remove docs folder

kenoir:
docs has moved to https://github.com/wellcomecollection/docs

Part of https://github.com/wellcometrust/platform/issues/3989


jamesgorrie:
> We've linked to these docs externally

Do we know where? In stacks? If so - could we change it there too?

alexwlchan:
> Do we know where?

The storage service RFC in particular has been handed around a bunch of times (sometimes in Stacks, sometimes private comms). There may be others.

We don't need to change every link right now, but at least a signpost rather than a 404 on the whole docs folder would be useful.

PR #4018, opened 2019-11-01
Catalogue VPC

kenoir:

Create a dedicated catalogue VPC to run services in (currently catalogue infra is running legacy style in the platform account.

### Who is this change for?

Folk who want properly partitioned & secure network / AWS resources.

For https://github.com/wellcomecollection/item-requests-service/issues/16

### DIFF
--- a/infrastructure/terraform/shared/vpcs.tf
+++ b/infrastructure/terraform/shared/vpcs.tf
@@ -32,6 +32,10 @@ locals {
   datascience_cidr_block_vpc     = "172.17.0.0/16"
   datascience_cidr_block_public  = "172.17.0.0/17"
   datascience_cidr_block_private = "172.17.128.0/17"
+
+  catalogue_cidr_block_vpc     = "172.18.0.0/16"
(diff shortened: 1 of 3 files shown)


infrastructure/terraform/shared/vpcs.tf
```diff
@@ -116,4 +120,28 @@ module "datascience_vpc" {
...
+}
+
+# Used by:
+# - Item requesting service
+# TODO: Move all catalogue services into this VPC
+
+module "catalogue_vpc" {
+  source = "github.com/wellcometrust/terraform//network/prebuilt/vpc/public-private-igw?ref=v19.5.3"
+
+  name = "catalogue-172-18-0-0-16"
```
alexwlchan: tiny nit: not in love with this name; if the CIDR block changes suddenly the name no longer matches. Why not call it catalogue?

kenoir: The CIDR block is fixed. Our VPCs subnets are all intended to be non-intersecting and need active planning to be so, hence the rather definite name.

PR #4020, opened 2019-11-01
Register a 'stacks' API with cognito

kenoir:

In order to build a service that allows users to request physical items we need to authenticate them.

### Who is this change for?

This change is to enable the item-requests/holdings project.


infrastructure/terraform/critical/cognito.tf
```diff
@@ -23,6 +23,24 @@ resource "aws_cognito_user_pool_domain" "domain" {
...
+  name       = "Stacks API V1"
+
+  scope = [
+    {
+      scope_name        = "requests"
+      scope_description = "Read and write requests"
+    },
+    {
+      scope_name        = "items"
+      scope_description = "Read the status of items"
```
alexwlchan: FWIW, I'd bake the permissions into the scope names, so if I just see the scope name, it's clear at a glance what I get. i.e.

```hcl
{
  scope_name = "requests_readwrite",
},
{
  scope_name = "items_readonly",
}
```

PR #4057, opened 2019-11-15
Grafana tweaks

alicefuzier:


monitoring/terraform/stack/grafana/main.tf
```diff
@@ -32,6 +32,8 @@ module "task" {
   container_image = "grafana/grafana:${var.grafana_version}"
   container_port  = "3000"
 
+  // If creating a new volume, note that the grafana folder on the volume will be owned by the root user.
+  // Grafana runs as user 472 so be sure to ssh into the EC2 instance and change the owner of the directory to 472.
```
alicefuzier: Turns out the reason why grafana wasn't starting is that the grafana directory in the efs volume was owned by the user root. Grafana runs as user 472 so I needed to ssh into the EC2 instance and change the owner of the directory. I'm not sure how to have terraform/amazon create the directory owned by the right user, but since the directory is in the EFS volume, it won't be an issue until we destroy and recreate the EFS volume, which shouldn't happen

alicefuzier: If anyone has any idea as to how to make this thing better it would be very appreciated though

PR #4132, opened 2020-01-06
Snapshot bucket

kenoir:

We currently rely on Elastic Cloud to hold our snapshots for the catalogue API ES cluster.

In order to make working with Catalogue API ES data easy to do locally and to do outside of Elastic Cloud for reduced time and expense, it is useful and arguably more reliable to hold these in an S3 bucket we have direct access to.

**This change will allow us to restore snapshots to ES clusters outside of Elastic Cloud.**

To complete this change we will need to make some changes to the Elastic Cloud configuration.

See: https://www.elastic.co/guide/en/cloud/current/ec-aws-custom-repository.html

### Who is this change for?

Developers and others wishing to make use of the ES data and to provide a way to extract data for the upcoming ES consultancy.


jamieparkinson:
Should this be in the `catalogue` repo?

kenoir:
@jamieparkinson it's intentionally here alongside other "don't lose this or you're in trouble" resources. As per https://github.com/wellcometrust/platform/tree/master/infrastructure

alexwlchan:
Although it’s all catalogue-critical infra, I think?

All the storage/archivematica-critical infra lives in those repos, not here.

jamieparkinson:
Only thought if it's to remain here is whether it would be more appropriate to allow the `catalogue-developer` principal to assume the readonly role, rather than  `platform-developer` - given the dynamic caller identity stuff I guess this could be done just by changing the `provider` to `catalogue` (?)

jamesgorrie:
What's the standard for infra now?

```
monorepo
- thingimade/src/etc
- terraform/thingimade/critical
- terraform/thingimade/infra
```

Or something like that? We have a few patterns about the place.

kenoir:
> What's the standard for infra now?
> 
> ```
> monorepo
> - thingimade/src/etc
> - terraform/thingimade/critical
> - terraform/thingimade/infra
> ```
> 
> Or something like that? We have a few patterns about the place.

Agree - there's no `src` here though and platform is weird as it's intended to hold shared things.

kenoir:
> Although it’s all catalogue-critical infra, I think?
> 
> All the storage/archivematica-critical infra lives in those repos, not here.

Agree - this should be in `critical` not `shared`, will move.

jamesgorrie:
Can we have definitions of `critical` and `shared`?

kenoir:
> Only thought if it's to remain here is whether it would be more appropriate to allow the `catalogue-developer` principal to assume the readonly role, rather than `platform-developer` - given the dynamic caller identity stuff I guess this could be done just by changing the `provider` to `catalogue` (?)

Agree (but will fix later) - This gives the platform account as the principal rather than the platform-dev role specifically as I needed something to fill the variable with, it won't be used immediately - yes it **should** be the catalogue-dev/readonly role.

kenoir:
> Can we have definitions of `critical` and `shared`?

There's not a good distinction here ... backups are more critical than shared?

Maybe:

- Critical: Anything that would require significant effort to recover from the loss of (which counts for backups in the event of ES cluster loss).
- Shared: Anything that doesn't fit the definition for "critical" but is shared across accounts.

PR #4239, opened 2020-02-06
add and store cloudfront logs in s3 bucket

jamesgorrie:
Adds logging to the Cloudfront distribution with a data retention period of objects being 30 days.

If we need larger time frames we should stream off this data and store it elsewhere as Cloudfront logs store things such as IPs etc.

### DIFF
--- a/infrastructure/terraform/critical/api.wellcomecollection.org/variables.tf
+++ b/infrastructure/terraform/critical/api.wellcomecollection.org/variables.tf
@@ -9,3 +9,5 @@ variable "cert_domain" {
 variable "public_api_bucket_domain_name" {}
 
 variable "description" {}
+
+variable "cf_logging_bucket" {}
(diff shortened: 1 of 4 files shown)


infrastructure/terraform/critical/api.wellcomecollection.org/variables.tf
```diff
@@ -9,3 +9,5 @@ variable "cert_domain" {
 variable "public_api_bucket_domain_name" {}
 
 variable "description" {}
+
+variable "cf_logging_bucket" {}
```
alexwlchan: Call it `cloudfront_logging_bucket` so it makes more sense for the next person who reads this code?

infrastructure/terraform/critical/s3.tf
```diff
@@ -0,0 +1,14 @@
+resource "aws_s3_bucket" "api_root_cf_logs" {
+  bucket = "weco-cloudfront-logs"
```
alexwlchan: nit: the convention elsewhere is to use `wellcomecollection-`, not `weco-`.

PR #4345, opened 2020-03-05
Add Experience nginx config

kenoir:
Build a new nginx container for wc.org using variable substitution like all other projects.

### DIFF
--- a/docker_run.py
+++ b/docker_run.py
@@ -1,0 +1,156 @@
+#!/usr/bin/env python3
+# -*- encoding: utf-8 -*-
+"""
+Run a Docker image used by a Make task.
+
+This script does _not_ run inside a container, so it has to work on any
+system.  This means:
+ 1. No third-party Python libraries!
+ 2. Python 2 and Python 3 compatible.
+
+Arguments after the two flags are passed directly to ``docker run``.
+
+"""
+
+import argparse
+import os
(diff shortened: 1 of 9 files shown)


kenoir:
Part of https://github.com/wellcomecollection/wellcomecollection.org/pull/5154/files

kenoir:
I'm not at all sure this is the best way to do nginx config - but it's less confusing for me if they all work the same way and I can see them all in the same place.

alexwlchan:
> I'm not at all sure this is the best way to do nginx config - but it's less confusing for me if they all work the same way and I can see them all in the same place.

At some point it would be good to remind myself: what do we get by putting nginx in front of our apps?

jamesgorrie:
@alexwlchan in this instance you get to put the webapp specific headers and control HTTP here rather than in the app, which is nice.

kenoir:
> > I'm not at all sure this is the best way to do nginx config - but it's less confusing for me if they all work the same way and I can see them all in the same place.
> 
> At some point it would be good to remind myself: what do we get by putting nginx in front of our apps?

Only services with APIs have nginx and if you look in the nginx configs, there are a few things (CORS, compression, redirection with API Gateway things) - extracting those from the applications makes the services simpler and having a standard pattern for types of app makes them more understandable.

